PR #19, opened 2019-09-20
string: strtok fix for alternating delimiters

ziemleszcz:
strtok delimiters can change between calls for the same str so you can't call strspn in advance

### DIFF
--- a/string/string.c
+++ b/string/string.c
@@ -367,19 +367,24 @@ char *strtok(char *s1, const char *s2)
 {
 	char *tokend;
 
-	if (s1 != NULL)
-		s1 += strspn(s1, s2);
-	else
+	if (s1 == NULL)
 		s1 = string_common.next_token;
+	s1 += strspn(s1, s2);
 
-	if (!*s1)
+	if (!*s1) {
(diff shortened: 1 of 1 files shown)


string/string.c
```diff
@@ -367,19 +367,23 @@ char *strtok(char *s1, const char *s2)
...
 
 	tokend = s1 + strcspn(s1, s2);
 
-	string_common.next_token = tokend + strspn(tokend, s2);
-
-	*tokend = 0;
+	if (*tokend) {
+		*tokend = 0;
+		string_common.next_token = tokend + 1;
+	} else {
```
agkaminski: else { should be in new line

string/string.c
```diff
@@ -367,19 +367,23 @@ char *strtok(char *s1, const char *s2)
...
 		return NULL;
+	}
 
 	tokend = s1 + strcspn(s1, s2);
 
-	string_common.next_token = tokend + strspn(tokend, s2);
-
-	*tokend = 0;
+	if (*tokend) {
+		*tokend = 0;
```
agkaminski: Please use '\0' for strings

PR #21, opened 2019-09-20
strncpy fix

ziemleszcz:
strncpy should pad dest string with zeroes when src len is less that n

### DIFF
--- a/string/string.c
+++ b/string/string.c
@@ -210,12 +210,17 @@ char *strcpy(char *dest, const char *src)
 #define __STRNCPY
 char *strncpy(char *dest, const char *src, size_t n)
 {
-	int i = 0;
+	size_t i;
 
-	do {
-		dest[i] = src[i];
-		i++;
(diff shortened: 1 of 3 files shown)


string/string.c
```diff
@@ -210,12 +210,16 @@ char *strcpy(char *dest, const char *src)
 #define __STRNCPY
 char *strncpy(char *dest, const char *src, size_t n)
 {
-	int i = 0;
+	size_t i = 0;
 
-	do {
+	while (i < n && src[i] != '\0') {
```
agkaminski: Wouldn't for loops fit better?

string/string.c
```diff
@@ -210,12 +210,16 @@ char *strcpy(char *dest, const char *src)
...
 		dest[i] = src[i];
 		i++;
-	} while (i < n && src[i - 1] != '\0');
+	}
+	while (i < n) {
```
agkaminski: memset perhaps?

PR #22, opened 2019-09-23
unistd: posix compatible dirname and basename

ziemleszcz:
dirname and basename didn't handle properly NULL input and "", "/a/", "a/", "a///b" cases


agkaminski:
This code won't handle, for example, a////b case.

PR #24, opened 2019-10-01
Fix includes

anglov:
This changes makes header files more compatible with C and POSIX and causing less pollution on global namespace. 
Also it adds some function from stubs to proper header, required by POSIX or common practice for compatibility.

### DIFF
--- a/sys/select.h
+++ b/sys/select.h
@@ -16,9 +16,14 @@
 #ifndef _SYS_SELECT_H_
 #define _SYS_SELECT_H_
 
+#include <time.h>
 #include <stdint.h>
-#include <string.h>
-#include <sys/time.h>
+
+
+struct timeval {
+	time_t tv_sec;
+	suseconds_t tv_usec;
+};
 
 
 #ifndef FD_SETSIZE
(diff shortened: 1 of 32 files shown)


sys/select.h
```diff
@@ -36,7 +41,12 @@ typedef struct fd_set_t_
 
 #define __FD_OP(i, s, op) ((s)->fds_bits[(i) / __NFDBITS] op (1u << ((i) % __NFDBITS)))
 
-#define FD_ZERO(s) memset((s), 0, sizeof(fd_set))
+#define FD_ZERO(s) \
+do { \
+	uint32_t i; \
```
agkaminski: why not memset((s)->fds_bits, 0, sizeof((s)->fds_bits))?

anglov: I defined this in that way because of needing memset prototype if it will be defined as macro in that header, what is obviously not permitted by a POSIX (global name space pollution). Previously it causes that symbols from string.h was available after including almost any other libphoenix headers, after this PR it's probably little better, but still can cause problems. 
So there are some solutions:
1. Provide string.h symbols and document it as a POSIX extension
2. Provide prototype only for memset and document it as a POSIX extension
3. Define FD_ZERO as a function and provide prototype only
4. Define FD_ZERO as a architecture-dependent macro and use inline assembly
5. Do not use memset at all
AFAIK a lot of open source C standard library providers (glibc, newlib, musl) decide not to use memset in sys/select.h header (glibc uses inline assembly but only on 64-bit targets). 

So I think it's reasonable to define macro in the way that I did.

PR #26, opened 2019-10-07
Split arch headers

anglov:
Please check it out.

Do not merge until all other repositories will not be ready for it.


include/arch/arm-imx/limits.h
```diff
@@ -16,11 +16,45 @@
...
+
+#define INT_MIN -2147483648
+#define INT_MAX 0x7fffffff
+#define UINT_MAX 0xffffffff
+
+#define LONG_MIN INT_MIN
+#define LONG_MAX INT_MAX
+#define ULONG_MAX UINT_MAX
+
+#define LONG_LONG_MIN 0x8000000000000000ll
```
agkaminski: Use the LL specifier to be consistent with rest of the code.

include/arch/arm-imx/arch.h
```diff
@@ -33,79 +33,7 @@
...
-#define PRIi32      "i"
-#define PRIo32      "o"
-
-#define PRId64      "lld"
-#define PRIu64      "llu"
-#define PRIx64      "llx"
-#define PRIX64      "llX"
-#define PRIi64      "lli"
-#define PRIo64      "llo"
+#define _PAGE_SIZE 0x1000
```
agkaminski: Would be great to have this value from a kernel header, if possible.

include/arch/riscv64/arch.h
```diff
@@ -16,63 +16,16 @@
...
-	id_t id;
-} oid_t;
-
-
-typedef u32 handle_t;
-
-
-#define SIZE_PAGE 0x1000
-
+#define _PAGE_SIZE 0x200
```
agkaminski: Wrong value! Should be 0x1000

include/stdio.h
```diff
@@ -16,14 +16,13 @@
...
 
-
 #include <arch.h>
 #include <stddef.h>
 #include <stdarg.h>
 #include <sys/types.h>
 
 #ifndef NOMMU
-#define BUFSIZ SIZE_PAGE
+#define BUFSIZ _PAGE_SIZE
```
agkaminski: PAGE_SIZE perhaps?

include/arch/arm-imx/limits.h
agkaminski: Rationale: l looks very similar to 1 on some fonts.

anglov: Agree, fixed.

include/arch/arm-imx/arch.h
anglov: Yeah, I think that it can be achieved along with posix revamp in kernel where many headers will be changed.

include/arch/riscv64/arch.h
anglov: Good catch, fixed.

include/stdio.h
anglov: Then we need separate header just for PAGE_SIZE to avoid name space pollution. I think that _PAGE_SIZE is reasonable consensus.

Issue #28, opened 2019-12-03
mktime invalid normalization


anglov:
mktime incorrectly normalizes struct tm. When year is leap year and you normalize month interval, one day is lost. 

eg.:
```
for (int i = 0; i < 60; ++i) {
    tm.tm_mon += 1;
    time = mktime(&tm);
    printf("%s", ctime(&time));
}
```

results with:
Tue Feb 1 06:00:00 2000
Wed Mar 1 06:00:00 2000
Sat Apr 1 06:00:00 2000
Mon May 1 06:00:00 2000
Thu Jun 1 06:00:00 2000
Sat Jul 1 06:00:00 2000
Tue Aug 1 06:00:00 2000
Fri Sep 1 06:00:00 2000
Sun Oct 1 06:00:00 2000
Wed Nov 1 06:00:00 2000
**Fri Dec 1 06:00:00 2000
Sun Dec 31 06:00:00 2000
Tue Jan 30 06:00:00 2001**
Fri Mar 2 06:00:00 2001
Mon Apr 2 06:00:00 2001
Wed May 2 06:00:00 2001
Sat Jun 2 06:00:00 2001
Mon Jul 2 06:00:00 2001
Thu Aug 2 06:00:00 2001
Sun Sep 2 06:00:00 2001
Tue Oct 2 06:00:00 2001
Fri Nov 2 06:00:00 2001
Sun Dec 2 06:00:00 2001
Wed Jan 2 06:00:00 2002
Sat Feb 2 06:00:00 2002
Sat Mar 2 06:00:00 2002
Tue Apr 2 06:00:00 2002
Thu May 2 06:00:00 2002
Sun Jun 2 06:00:00 2002
Tue Jul 2 06:00:00 2002
Fri Aug 2 06:00:00 2002
Mon Sep 2 06:00:00 2002
Wed Oct 2 06:00:00 2002
Sat Nov 2 06:00:00 2002
Mon Dec 2 06:00:00 2002
Thu Jan 2 06:00:00 2003
Sun Feb 2 06:00:00 2003
Sun Mar 2 06:00:00 2003
Wed Apr 2 06:00:00 2003
Fri May 2 06:00:00 2003
Mon Jun 2 06:00:00 2003
Wed Jul 2 06:00:00 2003
Sat Aug 2 06:00:00 2003
Tue Sep 2 06:00:00 2003
Thu Oct 2 06:00:00 2003
Sun Nov 2 06:00:00 2003
[...]

PR #29, opened 2019-12-03
Fixed mktime being off by one when not normalized fields are causing …

agkaminski:
…next year to occur


anglov:
Seems like it solves #28

PR #35, opened 2020-01-26
Makefile: unnecessary echo in adding removed

kemonats:


agkaminski:
This pull request is conflicting with the other one because of merge commit. Rebase it, please

PR #38, opened 2020-03-16
Makefile: fixed bugs

kemonats:


kemonats:
The library cannot be compiled without these patches.

If it depends on kernel includes, then I have two suggestions:
1.
```bash
path_inc := $(shell $(CC) -v 2>&1 | awk -F= 'BEGIN{RS=" "} /--with-sysroot/ {print $$2; exit}')
```
it won't always work if the toolchain was created in a different directory than it is currently in

2.
```bash
path_cc := $(shell bash -c 'command -v $(CC)')
target_cc :=$(shell $(CC) -v 2>&1 | awk '/Target:/ {print $$2}')
path := $(realpath $(dir $(path_cc)))
path := $(dir $(path))
path_inc := $(addprefix $(path), $(target_cc))
```
and 
```bash
%/errno.list: $(path_inc)/usr/include/phoenix/errno.h include/errno.h
```
The variable names are examples.

agkaminski:
Fixed in d1eb669a04dd5420a03f0f82ac80ce6594dc0678

Thank you for your help in fixing those issues.

PR #50, opened 2020-04-23
arm/stdlib: fix unaligned access to chunk_t struct

Cinek28:


agkaminski:
This will cause chunk_t to take aprox. 2 times the space it should take

agkaminski:
malloc structs are internal, they aren't send between processes/computers, so I would rather see no packing at all instead. This is packed structure to save memory. Proper solution, I think, is to use temperary variables instead of getting pointers to struct member.

@anglov is there a way to fix this problem without using temporary variables (to avoid getting pointer to packed struct meber) or setting aligment or not packing stuct at all?

nalajcie:
The `aligned(8)` makes only the required alignment of the struct equal to 8 (not all of it's fields) - so it just adds required padding at the end of the struct.

I think this just silences the compiler - makes it believe every chunk would be 8-aligned, but we do out own chunk address math (are we ensuring every `chunk_t` is aligned to 8?)

Here the `packed` attr is used in a hackish way - we don't actually want to pack the struct members, but ensure compiler uses safe approach in generating ASM to access it's members (byte-access instead of word-access - so it makes the the code slower).

This is also a bad pattern - the `packed` attr does nothing regarding the structure packing so my first instinct was to remove it altogether (if `chunk_t` is allowed to be unaligned the code would mysteriously stop working :) - the same applies for `heap_t`).

I don't know the exact error so I don't know is exactly what we're fighting with.

agkaminski:
> I don't know the exact error so I don't know is exactly what we're fighting with.

There you go:
`
CC  stdlib/malloc_dl.c      
stdlib/atexit.c:33:2: warning: #warning FIXME: atexit() uses GCC built-in function defined in Intel Pentium ABI! Works only with -mcpu=i586 [-Wcpp]
   33 | #warning FIXME: atexit() uses GCC built-in function defined in Intel Pentium ABI! Works only with -mcpu=i586
      |  ^~~~~~~
CC  stdlib/pty.c            
CC  stdlib/rand.c           
In file included from stdlib/malloc_dl.c:18:
stdlib/malloc_dl.c: In function '_malloc_chunkAdd':
stdlib/malloc_dl.c:209:76: warning: taking address of packed member of 'struct _chunk_t' may result in an unaligned pointer value [-Waddress-of-packed-member]
  209 |  exist = lib_treeof(chunk_t, node, lib_rbInsert(&malloc_common.lbins[idx], &chunk->node));
      |                                                                            ^~~~~~~~~~~~
/home/balek/git/_build/armv7m4-stm32l4x6/include/sys/rb.h:24:19: note: in definition of macro 'lib_treeof'
   24 |  void *tmpnode = (node);     \
      |                   ^~~~
stdlib/malloc_dl.c:212:24: warning: taking address of packed member of 'struct _chunk_t' may result in an unaligned pointer value [-Waddress-of-packed-member]
  212 |   chunk->node.parent = &chunk->node;
      |                        ^~~~~~~~~~~~
CC  stdlib/strtod.c         
stdlib/malloc_dl.c: In function '_malloc_chunkRemove':
stdlib/malloc_dl.c:238:43: warning: taking address of packed member of 'struct _chunk_t' may result in an unaligned pointer value [-Waddress-of-packed-member]
  238 |   lib_rbRemove(&malloc_common.lbins[idx], &chunk->node);
      |                                           ^~~~~~~~~~~~
stdlib/malloc_dl.c:245:44: warning: taking address of packed member of 'struct _chunk_t' may result in an unaligned pointer value [-Waddress-of-packed-member]
  245 |   rb_transplant(&malloc_common.lbins[idx], &chunk->node, &next->node);
      |                                            ^~~~~~~~~~~~
stdlib/malloc_dl.c:245:58: warning: taking address of packed member of 'struct _chunk_t' may result in an unaligned pointer value [-Waddress-of-packed-member]
  245 |   rb_transplant(&malloc_common.lbins[idx], &chunk->node, &next->node);
      |                                                          ^~~~~~~~~~~
stdlib/malloc_dl.c:247:30: warning: taking address of packed member of 'struct _chunk_t' may result in an unaligned pointer value [-Waddress-of-packed-member]
  247 |    next->node.left->parent = &next->node;
      |                              ^~~~~~~~~~~
stdlib/malloc_dl.c:249:31: warning: taking address of packed member of 'struct _chunk_t' may result in an unaligned pointer value [-Waddress-of-packed-member]
  249 |    next->node.right->parent = &next->node;
      |                               ^~~~~~~~~~~
In file included from stdlib/malloc_dl.c:18:
stdlib/malloc_dl.c: In function '_malloc_allocLarge':
stdlib/malloc_dl.c:372:81: warning: taking address of packed member of 'struct _chunk_t' may result in an unaligned pointer value [-Waddress-of-packed-member]
  372 |   chunk = lib_treeof(chunk_t, node, lib_rbFindEx(malloc_common.lbins[idx].root, &t.node, malloc_find));
      |                                                                                 ^~~~~~~
/home/balek/git/_build/armv7m4-stm32l4x6/include/sys/rb.h:24:19: note: in definition of macro 'lib_treeof'
   24 |  void *tmpnode = (node);     \
      |                   ^~~~
1

agkaminski:
Here we can't simply use temporary variables to solve this. Perhaps dropping packed attribute altogether is the best choice here.

Cinek28:
I will check that and correct the commit.

nalajcie:
> Here we can't simply use temporary variables to solve this. Perhaps dropping packed attribute altogether is the best choice here.

Yes, as it introduces run-time penalty which should be avoided at all costs in places like `malloc` (for more detailed explanation see: https://github.com/phoenix-rtos/phoenix-rtos-devices/pull/39#issuecomment-618880068).

What needs to be checked - if our `chunk_t` allocation policy and pointer arithmetic ensures `chunk_t` is always 4-byte aligned (or 8-byte if we would like to support arm64 in the future). This might already be handled as `malloc` needs to return 8-byte aligned pointer on this arch (alingof(uint64_t)).

agkaminski:
This might be the reason for packing this struct - to ensure that memory after it is aligned to 8 in all cases.

This might be not that simple to fix.

anglov:
Well @agkaminski not sure why even packed, also don't know how exactly our malloc works on big chunks but maybe it's badly formed (in meaning of packing or this node member)?

Anyways - about the question – don't know any way to avoid issue without using local variables (probably not suitable here as you pointed) nor at least partially unpacking struct.

agkaminski:
Maybe } __attribute__ ((packed, aligned(8))) chunk_t; is the best solution for now. It causes some memory to be wasted, but fixes problem without mayor code overhaul

nalajcie:
> This might be the reason for packing this struct - to ensure that memory after it is aligned to 8 in all cases.

I think not :). I don't want to dive into malloc algorithm but I think the `chunk_t` is at least 4-byte aligned right now (otherwise it would fail right away (in the lines with errors, or when casting chunk to `size_t*`)). It's not evident from the first glance at the code so that may just be a coincidence :).

Ok, the curiosity hat the better of me - I think that return address is 8-byte aligned because we always `CEIL` size in malloc to multiple of 8 in case of size and `CHUNK_OVERHEAD` is dividable by 8. But this would still need a through review.

> This might be not that simple to fix.

:+1: I completely agree.

I would vote for keeping the compiler warnings until this is properly addressed.

agkaminski:
@nalajcie Agreed

agkaminski:
Fixed in fad1ce94cf875ceae650a840a2769c384a652853

PR #65, opened 2020-08-26
add inb, outb functions to access memory-mapped registers

youki0:


pawelpisarczyk:
Julia I suggest to do develop UART registers access in another way.

PR #69, opened 2020-09-17
file.c: fixed buffer overflow and added NULL pointer checking

kemonats:


unistd/file.c
```diff
@@ -142,6 +161,11 @@ int symlink(const char *path1, const char *path2)
 
 	canonical1 = canonicalize_file_name(path2);
 	canonical2 = strdup(canonical1);
```
agkaminski, line 163: Shouldn't we check canonical1 for NULL before calling strdup()?

kemonats:
You're right, I should check canonical1 first before calling `strdup()`.
There is no `NULL` pointer check in `strdup()`.
I changed it.

PR #77, opened 2021-02-04
Improve IPv6 support

MaciejPurski:


net/if_indextoname.c
```diff
@@ -0,0 +1,42 @@
...
+#include <net/if.h>
+
+char *if_indextoname(unsigned int ifindex, char *ifname)
+{
+	struct ifreq ifr = {0};
+	int fd, res;
+
+	fd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (fd <= 0)
+		return 0;
```
kemonats, line 29:
```suggestion
		return NULL;;
```
as it is bellow

net/inet_pton.c
```diff
@@ -1,6 +1,146 @@
...
+	in_addr_t addr4 = 0;
+	unsigned long val;
+	int i;
+	char *endptr;
+
+	for (i = 0; i < 4; i++) {
+		if (!isdigit(*src))
+			return 0;
+
+		val = strtoul(src, &endptr, 0);
```
kemonats, line 34:
```suggestion
		val = strtoul(src, &endptr, 10);
		
```
address ip can contain digits

sys/socket.c
```diff
@@ -435,3 +436,122 @@ int socketpair(int domain, int type, int protocol, int socket_vector[2])
...
+	size_t bufsz, namesz, addrsz;
+	char *p;
+
+	msg.type = sockmGetIfAddrs;
+	msg.o.size = 256;
+
+	for (;;) {
+		void *buf = realloc(msg.o.data, msg.o.size);
+		if (!buf) {
+			realloc(msg.o.data, 0);
```
kemonats, line 455:
```suggestion
			free(msg.o.data);
```
In `libphoenix` this functionality is probably not implement.

sys/socket.c
```diff
@@ -435,3 +436,122 @@ int socketpair(int domain, int type, int protocol, int socket_vector[2])
...
+			return EAI_MEMORY;
+		}
+		msg.o.data = buf;
+
+		if (socksrvcall(&msg) < 0) {
+			realloc(msg.o.data, 0);
```
kemonats, line 461:
```suggestion
			free(msg.o.data);
```
as above

sys/socket.c
```diff
@@ -435,3 +436,122 @@ int socketpair(int domain, int type, int protocol, int socket_vector[2])
...
+
+	bufsz = smo->sys.buflen;
+	if (smo->ret || bufsz > msg.o.size) {
+		free(msg.o.data);
+		if (smo->ret == EAI_SYSTEM)
+			(void)SET_ERRNO(-smo->sys.err);
+		return smo->ret;
+	}
+
+	*ifap = msg.o.data = realloc(msg.o.data, bufsz);
```
kemonats, line 482: - no check for NULL pointer
- memory leak if `raelloc()` return NULL

net/if_indextoname.c
```diff
@@ -0,0 +1,42 @@
...
+
+	ifr.ifr_ifindex = ifindex;
+	res = ioctl(fd, SIOCGIFNAME, &ifr);
+
+	close(fd);
+	if (res < 0) {
+		return NULL;
+	}
+	else {
+		strncpy(ifname, ifr.ifr_name, IFNAMSIZ);
```
kemonats, line 39:
```suggestion
		strncpy(ifname, ifr.ifr_name, IFNAMSIZ);
		ifbame[IFNAMSIZ - 1] = '\0';
```
eliminating buffer overflows

net/inet_pton.c
```diff
@@ -1,6 +1,146 @@
...
+			dst->s6_addr16[i] = addr6.s6_addr16[i - (n_hextets - idx)];
+	}
+
+	return 1;
+}
+
+
+int inet_pton(int af, const char *src, void *dst)
+{
+	if (src == NULL || dst == NULL)
```
kemonats, line 135: You can possibly add a check for empty `src`.
```
	if (src == NULL || dst == NULL || *src == '\0')
```

sys/socket.c
MaciejPurski: This idea of calling realloc() with 0 size looked a bit suspicious for me, as I based this implementation on `getaddrinfo()`. So if it does not work properly I guess we should make the same fix in `getaddrinfo()` too.

include/arpa/inet.h
```diff
@@ -23,9 +23,9 @@
 int inet_aton(const char *cp, struct in_addr *inp);
 in_addr_t inet_addr(const char *cp);
 char *inet_ntoa(struct in_addr in);
-const char *inet_ntop(int af, const void *restrict src,
-		       char *restrict dst, socklen_t size);
-int inet_pton(int af, const char *restrict src, void *restrict dst);
+const char *inet_ntop(int af, const void *src,
```
nalajcie, line 26: FYI, `restrict` keyword is in the Open Group POSIX specification for this function: https://pubs.opengroup.org/onlinepubs/9699919799/functions/inet_ntop.html

Were the keywords removed due to the implementation?

net/inet_ntop.c
```diff
@@ -1,27 +1,115 @@
...
+		if (tmp_idx != -1 && (addr->s6_addr16[i] != 0 || i == n_hextets - 1)) {
+			if (tmp_len > 1 && tmp_len > zero_len) {
+				zero_len = tmp_len;
+				zero_idx = tmp_idx;
+			}
+			tmp_idx = -1;
+		}
+	}
+
+	for (i = 0; i < n_hextets;) {
```
nalajcie, line 84: seems like a `while` :)

net/inet_ntop.c
```diff
@@ -1,27 +1,115 @@
...
+			if (i == 0)
+				*(dst_ptr++) = ':';
+			*(dst_ptr++) = ':';
+			i += zero_len;
+		} else {
+			dst_ptr += snprintf(dst_ptr, 5, "%x", ntohs(addr->s6_addr16[i]));
+			if (i != n_hextets - 1)
+				*(dst_ptr++) = ':';
+			i++;
+		}
```
nalajcie, line 95: I don't know how the IPv6 addresses are usually printed but it seems that common parts could be moved to outside of the if-else block?
```suggestion
		if (i == zero_idx) {
			if (i == 0)
				*(dst_ptr++) = ':';
			i += zero_len;
		} else {
			dst_ptr += snprintf(dst_ptr, 5, "%x", ntohs(addr->s6_addr16[i]));
			i++;
		}

		if (i != n_hextets)
				*(dst_ptr++) = ':';
```

net/inet_ntop.c
```diff
@@ -1,27 +1,115 @@
...
+	}
+
+	return dst;
+}
+
+const char *inet_ntop(int af, const void *src,
+					  char *dst, socklen_t size)
```
nalajcie, line 102: I think the coding convention says that multiline alignment should be done by spaces - otherwise it looks like that :).

net/inet_pton.c
nalajcie: the function would still return valid data for `src == '\0'` so it's not a strictly coding bug (as opposed to passing NULL to this function.

Regarding NULL checking - I might opt for making asserts here against NULL or allowing to crash the app by accessing the `NULL`, otherwise it might be hard to find a bug regarding invalid usage of `inet_pton`/`inet_ntop`, but that's probably broader discussion.

net/if_indextoname.c
nalajcie: or use `snprintf`, but that's wasteful in this particular case

net/inet_ntop.c
MaciejPurski: Unfortunately, your suggestion does not work properly, when in a case, where the address ends with an all-zero string. For example, after applying your change, the function would produce: `fd00:0:8000:80:` instead of `fd00:0:8000:80::`.
Maybe, I'll think it over and find a better way to simplify this ugly loop.

net/inet_pton.c
```diff
@@ -1,6 +1,146 @@
...
+
+		val = strtoul(src, &endptr, 0);
+		if (endptr == src || val > 255)
+			return 0;
+		if ((*endptr != '.' && i < 3) ||
+			(*endptr != '\0' && i == 3))
+			return 0;
+
+		src = endptr + 1;
+		addr4 = addr4 << 8 | (val & 0xFFFF);
```
kemonats, line 42:
```suggestion
		addr4 = addr4 << 8 | val ;
```
`val` is `unsigned int` and `<= 255`

net/inet_pton.c
```diff
@@ -1,6 +1,147 @@
...
+	in_addr_t addr4 = 0;
+	unsigned long val;
+	int i;
+	char *endptr;
+
+	for (i = 0; i < 4; i++) {
+		if (!isdigit(*src))
+			return 0;
+
+		val = strtoul(src, &endptr, 10);
```
kemonats, line 34: As for me checking if the first character is a digit is redundant, function `strtoul()`  checks it anyway.

net/inet_pton.c
```diff
@@ -1,6 +1,147 @@
...
+		if (endptr == src || (endptr - src) > 3 || val > 255)
```
kemonats, line 35:
```suggestion
		if (endptr == src || val > 255)
```
If `endptr - src > 3` (min 4 digits)  condition `val > 255` is true.

sys/socket.c
```diff
@@ -435,3 +436,120 @@ int socketpair(int domain, int type, int protocol, int socket_vector[2])
...
+
+	msg.type = sockmGetIfAddrs;
+	msg.o.size = 256;
+
+	for (;;) {
+		msg.o.data = realloc(msg.o.data, msg.o.size);
+		if (!msg.o.data) {
+			free(msg.o.data);
+			return EAI_MEMORY;
+		}
```
kemonats, line 457:
```suggestion
		void *tmp;
		tmp = realloc(msg.o.data, msg.o.size);
		if (tmp == NULL) {
			free(msg.o.data);
			return EAI_MEMORY;
		}
		msg.o.data = tmp;
		
```
Memory leak.
If `realloc()` returns `NULL` you will lose previous value of `msg.o.data`.

net/inet_pton.c
MaciejPurski: According to [Open Group](https://pubs.opengroup.org/onlinepubs/009604499/functions/strtoul.html) `strtoul()` accepts leading white spaces as well as a sign character. I was referring to that documentation, when writing my function. I have just checked that our implementation of `strtoul` does not conform to that documentation and won't accept such strings. However, I think that this one redundant check is not that painful, given that someone might fix the `strtoul` implementation in the future.

net/inet_pton.c
MaciejPurski: According to [OpenGroup](https://pubs.opengroup.org/onlinepubs/009695399/functions/inet_ntop.html) `inet_pton` function in IPv4 address case (IPv6 is analogous) shall accept addresses consisting of 4 numbers, which are 1 to 3 digits. This time our implementation of `strtoul` conforms to the docs and accepts leading-zeros strings. So, without this check addresses like this: `0000000000192.00000000168.0001.000000000000001` would be accepted. I guess, you'd agree, we don't want such addresses to be accepted. I have added a similar check in the IPv6 case.

sys/socket.c
MaciejPurski: You're right here. I'll fix that.

nalajcie (approved):

PR #81, opened 2021-04-26
libphoenix: add contructors and destructors

HBuczynski:
This PR depends on changes in kernel: https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/191

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,7 @@ include err/Makefile
 include errno/Makefile
 include locale/Makefile
 include math/Makefile
+include misc/Makefile
 include net/Makefile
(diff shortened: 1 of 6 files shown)


agkaminski (approved):
LGTM

It might be a right impulse to finaly investige removal of nostdlib flag. 
Decision of making new dir `misc` might be controversial, but I don't have any better idea, so I'm good with this.

nalajcie (approved):
I think we should make `libphoenix` a proper standard library sooner or later (compile it to `libc.a`, separate initialization files, compile with toolchain, etc.) - but not as a part of this PR :).

PR #86, opened 2021-05-25
stdio: snprintf(), vsnprintf() return value fix

ziemleszcz:
According to POSIX specification:

"Upon successful completion, the snprintf() function shall return
the number of bytes that would be written to s had n been
sufficiently large excluding the terminating null byte."

"If the value of n is zero on a call to snprintf(), nothing shall
be written, the number of bytes that would have been written had n
been sufficiently large excluding the terminating null shall be
returned, and s may be a null pointer."

JIRA: RTOS-24

### DIFF
--- a/stdio/sprintf.c
+++ b/stdio/sprintf.c
@@ -18,6 +18,12 @@
 #include "format.h"
 
 
+typedef struct _vsprintf_ctx_t {
+	char *buff;
+	size_t n;
+} vsprintf_ctx_t;
+
+
 typedef struct _vsnprintf_ctx_t {
 	char *buff;
 	size_t n;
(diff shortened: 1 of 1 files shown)


agkaminski (approved):
LGTM

stdio/sprintf.c
```diff
@@ -18,28 +18,38 @@
...
+{
+	vsnprintf_ctx_t* ctx = (vsnprintf_ctx_t *)context;
+
+	if (ctx->n < ctx->max_len) {
+		ctx->buff[ctx->n] = c;
 	}
+
+	/* Count anyway to return the number of characters that would have been
+	 * written if buffer had been sufficiently large */
+	ctx->n++;
```
nalajcie, line 52: according to C99 the `snprintf` output should always be null-terminated if `max_len > 0` (https://en.cppreference.com/w/c/io/fprintf).

I think You removed this functionality in case the result string overflows the buffer.

nalajcie (approved):

PR #87, opened 2021-05-25
include/getopt.h: add getopt_long()

ziemleszcz:
JIRA: DTR-10


nalajcie:
I'm not sure, but I think there is a convention to explicitly prepend the function declaration with `extern`, other than that LGTM :).

PR #88, opened 2021-05-27
Libphoenix fixes needed for running some busybox applets

nalajcie:


sys/socket.c
```diff
@@ -555,3 +564,16 @@ void freeifaddrs(struct ifaddrs *ifa)
 {
 	free(ifa);
 }
+
+const char *hstrerror(int err)
+{
+	switch (err) {
+		case RESOLVED_OK: return "SUCCESS";
```
HBuczynski, line 571: Please, add return in a new line.

nalajcie:
added CI by separate PR + rebased on current master for automatic checks

sys/socket.c
nalajcie: ok, `clang-format` allowed this construction, resolving

PR #95, opened 2021-06-10
uio: add readv, writev, IOV_MAX, _SC_IOV_MAX

ziemleszcz:
JIRA: RTOS-46

### DIFF
--- a/include/sys/uio.h
+++ b/include/sys/uio.h
@@ -16,16 +16,24 @@
 #ifndef _SYS_UIO_H_
 #define _SYS_UIO_H_
 
-#include <stdint.h>
+#include <stddef.h>
+#include <sys/types.h>
+#include <sys/syslimits.h>
 
 
-#define UIO_MAXIOV	1024
+#define UIO_MAXIOV IOV_MAX
 
 
 struct iovec {
 	void *iov_base;
 	size_t iov_len;
 };
+
+
+ssize_t readv(int fd, const struct iovec *iov, int iovcnt);
+
+
+ssize_t writev(int fd, const struct iovec *iov, int iovcnt);
 
 
 #endif /* _SYS_UIO_H_ */
--- a/sys/uio.c
+++ b/sys/uio.c
@@ -1,0 +1,91 @@
+/*
+ * Phoenix-RTOS
+ *
+ * libphoenix
+ *
+ * sys/uio.c
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Ziemowit Leszczynski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <errno.h>
+#include <unistd.h>
+#include <limits.h>
+#include <sys/uio.h>
+
+
+static int checkv(const struct iovec *iov, int iovcnt)
+{
+	unsigned int i;
+	ssize_t max_len = SSIZE_MAX;
+
+	if (iovcnt <= 0 || iovcnt > IOV_MAX)
+		return -1;
+
+	for (i = 0; i < iovcnt; ++i) {
+		if (iov[i].iov_len > max_len)
+			return -1;
+		max_len -= iov[i].iov_len;
+	}
+
+	return 0;
+}
+
+
+ssize_t readv(int fd, const struct iovec *iov, int iovcnt)
+{
+	unsigned int i;
+	ssize_t n, tot_len = 0;
+
+	if (-1 == checkv(iov, iovcnt))
+		return -EINVAL;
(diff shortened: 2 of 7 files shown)


include/sys/uio.h
```diff
@@ -28,4 +30,8 @@ struct iovec {
 };
 
 
+ssize_t readv(int fd, const struct iovec *iov, int iovcnt);
+ssize_t writev(int fd, const struct iovec *iov, int iovcnt);
```
agkaminski, line 34: There should be 2 empty lines between prototypes (just a style thing).

sys/uio.c
```diff
@@ -0,0 +1,88 @@
...
+
+
+static int checkv(const struct iovec *iov, int iovcnt)
+{
+	ssize_t max_len = SSIZE_MAX;
+
+	if (iovcnt <= 0 || iovcnt > IOV_MAX)
+		return -1;
+
+	for (unsigned int i = 0; i < iovcnt; ++i) {
```
agkaminski, line 29: "Local variables

Local variables should be defined before the function code according to ANSI C 89 standard. (...)"

https://phoenix-rtos.com/documentation/coding.md

sys/uio.c
```diff
@@ -0,0 +1,88 @@
...
+	ssize_t tot_len = 0;
+
+	if (-1 == checkv(iov, iovcnt))
+		return -EINVAL;
+
+	/* TODO: file locking */
+
+	for (unsigned int i = 0; i < iovcnt; ++i) {
+		if (iov[i].iov_len > 0) {
+			ssize_t n = read(fd, iov[i].iov_base, iov[i].iov_len);
```
agkaminski, line 50: As above, variables should be defined before the function code.

sys/uio.c
```diff
@@ -0,0 +1,88 @@
...
+	ssize_t tot_len = 0;
+
+	if (-1 == checkv(iov, iovcnt))
+		return -EINVAL;
+
+	/* TODO: file locking */
+
+	for (unsigned int i = 0; i < iovcnt; ++i) {
+		if (iov[i].iov_len > 0) {
+			ssize_t n = write(fd, iov[i].iov_base, iov[i].iov_len);
```
agkaminski, line 76: As above, variables should be defined before the function code.

sys/uio.c
```diff
@@ -0,0 +1,88 @@
...
+	for (unsigned int i = 0; i < iovcnt; ++i) {
+		if (iov[i].iov_len > 0) {
+			ssize_t n = read(fd, iov[i].iov_base, iov[i].iov_len);
+			if (n == -1)
+				return -1;
+
+			tot_len += n;
+
+			if (n < iov[i].iov_len)
+				break;
```
agkaminski, line 57: Why we give up if we read less than iov_len? I think we should read again and try to fill this buffer and not move on to next buffer

sys/uio.c
```diff
@@ -0,0 +1,88 @@
...
+	for (unsigned int i = 0; i < iovcnt; ++i) {
+		if (iov[i].iov_len > 0) {
+			ssize_t n = write(fd, iov[i].iov_base, iov[i].iov_len);
+			if (n == -1)
+				return -1;
+
+			tot_len += n;
+
+			if (n < iov[i].iov_len)
+				break;
```
agkaminski, line 83: Why we give up if we wrote less than iov_len? I think we should write the missing data from current buffer, not move on to the next buffer

sys/uio.c
ziemleszcz: readv semantics is similar read, it can return smaller number of bytes than requested (no data available, interrupted by signal). Note that we are escaping "for" loop instead of moving to next buffer in this case.

sys/uio.c
ziemleszcz: Check previous reply.

sys/uio.c
nalajcie: please note that it's false that C89 allows variable declaration at the beginning of the function - it actually allows declarations at the beginnig of each compound statement (commonly called a "block" - `{ .. }`) - exact syntax from C89 specification: http://port70.net/~nsz/c/c89/c89-draft.html#3.6.2.

If we're going for C89 compatibility, we should reflect that in our code guidelines and in CFLAGS (`-std=c89 -pedantic`) which will make non-compliant code throw compilation errors :).

IMHO, this code is a valid C89 and also limits the scope of the variable which improves the readability and reduces change for additional errors (by limiting the scope of variable).

Of course, declaring variables inside `for`/`while` (the above comment) is not C89-compliant.

agkaminski (approved):

PR #97, opened 2021-06-15
string: add legacy bcopy

ziemleszcz:
JIRA: RTOS-46


agkaminski (commented):
As linter suggest, `__attribute__` should be in the same line as the function declaration

agkaminski (approved):

PR #98, opened 2021-06-23
Ziemleszcz/ipsec

ziemleszcz:
## Description
IPSEC headers and getprotobyname/getprotobynumber implementation.

## Motivation and Context
Required for IPSEC driver and openiked port.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (PC).

## Checklist:

## Special treatment

### DIFF
--- a/sys/proto.c
+++ b/sys/proto.c
@@ -1,0 +1,157 @@
+/*
+ * Phoenix-RTOS
+ *
+ * libphoenix
+ *
+ * sys/proto.c
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Ziemowit Leszczynski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <netdb.h>
+#include <string.h>
+#include <stdlib.h>
+
+
+#define PROTOCOLS_NUM (sizeof(protocols) / sizeof(struct protoent_internal))
+#define MAX_ALIASES   2
+
+
+struct protoent_internal {
+	const char *name;
+	const char *aliases[MAX_ALIASES];
+	int proto;
+};
+
+
+static const struct protoent_internal protocols[] = {
+	{ "ip", { "IP", NULL }, 0 },
+	{ "hopopt", { "HOPOPT", NULL }, 0 },
+	{ "icmp", { "ICMP", NULL }, 1 },
+	{ "igmp", { "IGMP", NULL }, 2 },
+	{ "ggp", { "GGP", NULL }, 3 },
+	{ "ipencap", { "IP-ENCAP", NULL }, 4 },
+	{ "st", { "ST", NULL }, 5 },
+	{ "tcp", { "TCP", NULL }, 6 },
+	{ "egp", { "EGP", NULL }, 8 },
+	{ "igp", { "IGP", NULL }, 9 },
+	{ "pup", { "PUP", NULL }, 12 },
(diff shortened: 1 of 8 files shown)


sys/proto.c
```diff
@@ -0,0 +1,158 @@
...
+
+
+struct protoent_internal {
+	const char *name;
+	const char *aliases[MAX_ALIASES];
+	int proto;
+};
+
+
+struct protoent_internal protocols[PROTOCOLS_NUM] = {
```
agkaminski, line 32: Could be static const

sys/proto.c
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * %LICENSE%
+ */
+
+#include <netdb.h>
+#include <string.h>
+#include <stdlib.h>
+
+
+#define PROTOCOLS_NUM 55
```
agkaminski, line 21: It's better and safer to get this number by using `sizeof(protocols) / sizeof(*protocols)`

sys/proto.c
```diff
@@ -0,0 +1,158 @@
...
+
+static void protoent_free(struct protoent *proto)
+{
+	unsigned int i = 0;
+
+	if (proto->p_name)
+		free(proto->p_name);
+
+	if (proto->p_aliases) {
+		while (i < MAX_ALIASES && proto->p_aliases[i]) {
```
agkaminski, line 103: This while works exactly once - for `i == 0`. Is this intended? What about `p_aliases[1]` and `p_aliases[2]` (`p_aliases` have 3 indexes)?

agkaminski: Also, `for` loop fits here better

sys/proto.c
```diff
@@ -0,0 +1,158 @@
...
+static void protoent_copy(struct protoent *proto, unsigned int num)
+{
+	unsigned int i = 0;
+
+	proto->p_name = strdup(protocols[num].name);
+
+	while (i < MAX_ALIASES && protocols[num].aliases[i]) {
+		proto->p_aliases[i] = strdup(protocols[num].aliases[i]);
+		++i;
+	}
```
agkaminski, line 120: As above, only one iteration, for loop fits better

sys/proto.c
```diff
@@ -0,0 +1,158 @@
...
+
+	proto->p_proto = protocols[num].proto;
+}
+
+
+struct protoent *getprotobyname(const char *name)
+{
+	unsigned int i = 0;
+
+	while (i < PROTOCOLS_NUM) {
```
agkaminski, line 131: For loop fits better

sys/proto.c
```diff
@@ -0,0 +1,158 @@
...
+
+	return NULL;
+}
+
+
+struct protoent *getprotobynumber(int proto)
+{
+	unsigned int i = 0;
+
+	while (i < PROTOCOLS_NUM) {
```
agkaminski, line 148: For loop fits better

agkaminski (changes requested):

sys/proto.c
ziemleszcz: It loops 0 to MAX_ALIASES times. `p_aliases` is array of string pointers terminated by NULL. This function is used only on `proto_common` variable.

agkaminski (approved):
Searching perhaps could be done better (right now is O(n), could be O(1) by number and O(log n) by name), but if we don't need peak performance LGTM.

PR #99, opened 2021-06-23
Real pathname resolution

nalajcie:
## Description
This change brings real path resolution to Phoenix-RTOS (with '.', '..' and symlinks support (depends on FS used)) with correct errno reporting.

**WARN: This is a breaking change as canonicalize_path_name() works now according to POSIX (which previous implementation was non-conformant with). For internal usage please use either POSIX API or native resolve_path() with extended functionality.**

## Motivation and Context
JIRA: DTR-75
phoenix-rtos/phoenix-rtos-project#113

## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/29
- [x] Tested by hand on: **ia32-generic** (ext2), **DTR** (jffs, dummyfs)

## Checklist:

## Special treatment

  - https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/204
  - https://github.com/phoenix-rtos/phoenix-rtos-filesystems/pull/39

### DIFF
--- a/unistd/dir.c
+++ b/unistd/dir.c
@@ -18,6 +18,8 @@
 #include <dirent.h>
 #include <stdio.h>
 #include <errno.h>
+#include <limits.h>
+#include <assert.h>
 #include <string.h>
 #include <libgen.h>
 #include <sys/msg.h>
@@ -29,6 +31,9 @@
 static struct {
 	char *cwd;
 } dir_common;
+
+static ssize_t _readlink_abs(const char *path, char *buf, size_t bufsiz);
+static int _resolve_abspath(char *path, char *result, int resolve_last_symlink, int allow_missing_leaf);
 
 
 int chdir(const char *path)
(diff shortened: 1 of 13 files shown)


nalajcie:
@phoenix-rtos/phoenix-team **This is a breaking change**, please check for usage of `canonicalize_path_name()` in Your code.

As it brings us closer to POSIX compatibility, it makes some programs from `ports` work better (eg. `mkdir -p` started to work).

@Maxez I know for sure it breaks tab-completion in `psh`, please look into that as soon as possible.

unistd/file.c
```diff
@@ -268,10 +257,15 @@ int create_dev(oid_t *oid, const char *path)
 					tpath = tpathalloc;
 				}
 
+				/* TODO: not needed, we may assume it will always be canonical? */
+#if 0
```
nalajcie, line 261: @agkaminski : I don't know how `create_dev()` is being used (it would be nice to add some unit tests with standard use cases.

Please comment.

unistd/file.c
```diff
@@ -290,8 +284,15 @@ int create_dev(oid_t *oid, const char *path)
 	if (!strncmp("/dev", path, 4))
 		path += 4;
 
+#if 0
```
nalajcie, line 287: @agkaminski : I don't know how `create_dev()` is being used (it would be nice to add some unit tests with standard use cases.

Please comment.

agkaminski:
@nalajcie Here's one example: https://github.com/phoenix-rtos/phoenix-rtos-devices/blob/9042ab1a8433e597ee17fc7746cedb85ccfc432e/tty/imx6ull-uart/imx6ull-uart.c#L429

unistd/dir.c
```diff
@@ -117,50 +122,185 @@ char *getcwd(char *buf, size_t size)
...
+	*r = 0;
+	return r - result;
+}
+
+
+/* Resolving path to absolute paths with '.', '..' and symlinks support. Additional params to satisfy multiple libc use cases.
+ *   resolved_path: if NULL, will be allocated by malloc()
+ *   allow_missing_leaf: don't return ENOENT if last path node is not existing
+ *   resolve_last_symlink: if 0 return final symlink path instead of symlink destination
+ * NOTE: stack usage: ~(2 * PATH_MAX)
```
agkaminski, line 254: 2 KB stack usage (on top on all current usage) is a real breaking change here. It's worth to consider malloc here

nalajcie:
** High priority**: I would like to merge it today.

- decided to leave `create_dev` as is, it's use cases should be reviewed and some unit tests should be added. I believe parts of it's implementation is obsolete as it works with commented-out code :)
- @HBuczynski @anglov @gerard5 can You check if it breaks some use cases on small targets? (If it's due to invalid `canonicalize_path_name` usage - the relevant code should be fixed (so either create an issue or fix if possible).

unistd/file.c
```diff
@@ -147,67 +150,55 @@ int mkfifo(const char *filename, mode_t mode)
...
 		return -EINVAL;
 
-	canonical1 = canonicalize_file_name(path2);
-	if (canonical1 == NULL) {
-		return -ENOMEM;
+	/* allow_missing_leaf = 1 -> we will be creating a file */
+	canonical_linkpath = resolve_path(linkpath, NULL, 1, 1);
+	if (canonical_linkpath == NULL) {
+		return -1; /* errno set by resolve_path */
 	}
```
agkaminski, line 169: No need for brackets

unistd/dir.c
nalajcie: please note that on small tagets `PATH_MAX` was defined as `256` (see relevant `limits.h`)

anglov:
What I see is large increase of create_dev stack usage 728 → 1284
Also major stack usage increase of getpw* functions – 1456 → 2012, what causes psh_login/psh_ls exceeding 2048 B stack usage (currently set to 4096 on all platforms I think. Hope that larger PATH_MAX/NAME_MAX doesn't cause stack overflow on bigger platforms).

dummyfs is now really near to reach stack overflow: 1064 → 1436.

I don't found any direct breakage right away – platform initializing correctly and looks working stable, but don't look carefully on code and don't strict checking everything.

anglov:
lwip will stack overflow with this PR  1104→1660

unistd/file.c
nalajcie: the current code is quite inconsequential, I went for "minimal change" - so if there was single-statement if with brackets I left it as-is.

Changed in next patchset (will resolve when pushed)

unistd/dir.c
nalajcie: To allocate all buffers I would need to use 3 mallocs, I think we can compromise and make `path_copy` malloc'ed as the code flow would be quite easy. This will result in stack usage increase by ~256 bytes on small targets (instead of 512).

The best approach for the future would be to specify proper `struct reent` (per-thread libc state to support MT-safe and reentrant functions) and keep some "temp buffer" there for use in similar cases.

Please note that BSS can't be used as `realpath()` is required to be MT-safe.

nalajcie:
> What I see is large increase of create_dev stack usage 728 → 1284
> Also major stack usage increase of getpw* functions – 1456 → 2012, what causes psh_login/psh_ls exceeding 2048 B stack usage (currently set to 4096 on all platforms I think. Hope that larger PATH_MAX/NAME_MAX doesn't cause stack overflow on bigger platforms).
> 
> dummyfs is now really near to reach stack overflow: 1064 → 1436.

Thanks for checking, the stack increase of ~512 (2* PATH_MAX - 256 is minimum required by POSIX) is to be expected. I think the best approach would be to change `path_copy` allocation to dynamic, so the stack usage increase will be only ~256.

gerard5:
On imxrt1064 found that:
- occasionally stack overflows in `dummyfs`, probably of what @anglov wrote about
- `ps_plcio_driver` does not start (I'm investigating that), possibly the issue is connected with `dummyfs`
- `psh` tab completion is broken (as You know), `out of memory`.

unistd/file.c
nalajcie: leaving as-is, to be done in separate PR

unistd/file.c
nalajcie: leaving as-is, to be done in separate PR

agkaminski (approved):
LGTM, @gerard5 @anglov any remarks?

gerard5 (commented):
- The code looks good to me, but I can't verify the symbolic link part on `imxrt1064`
-  `mkdir()`, `stat()`, `open()`, `create_dev()`, and others seems to work how I use them,
- `ps_plcio_driver` is fixed now and works fine,
- `dummyfs` seams to work too.
- `psh` still needs to be fixed.

Additionally, I built unit tests for the `imxrt1064` and ran `test-libc` without success, there is my comment about that what failed: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/29#issuecomment-868717515

nalajcie:
Will rebase and merge in a few minutes, if anybody is against it - now it's the moment to state Your claim :)

gerard5:
> Will rebase and merge in a few minutes, if anybody is against it - now it's the moment to state Your claim :)

For me, it's ok.

nalajcie:
Merging without approval, as there was a previous approval from @agkaminski dismissed by some editorial changes (rebasing on top of current master)

PR #100, opened 2021-06-28
uio: fix readv & writev return value, set EINVAL errno

ziemleszcz:
According to POSIX:

1. If a read() is interrupted by a signal before it reads any
data, it shall return -1 with errno set to [EINTR].
2. If a read() is interrupted by a signal after it has
successfully read some data, it shall return the number of
bytes read.
3. If write() is interrupted by a signal before it writes any
data, it shall return -1 with errno set to [EINTR].
4. If write() is interrupted by a signal after it successfully
writes some data, it shall return the number of bytes written.

JIRA: RTOS-46

## Motivation and Context

Current implementation of readv/writev breaks dropbear (channel write).

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imx6ull).

## Checklist:

## Special treatment

### DIFF
--- a/sys/uio.c
+++ b/sys/uio.c
@@ -42,16 +42,21 @@ ssize_t readv(int fd, const struct iovec *iov, int iovcnt)
 	unsigned int i;
 	ssize_t n, tot_len = 0;
 
-	if (-1 == checkv(iov, iovcnt))
-		return -EINVAL;
+	if (-1 == checkv(iov, iovcnt)) {
+		errno = EINVAL;
+		return -1;
+	}
 
 	/* TODO: file locking */
 
 	for (i = 0; i < iovcnt; ++i) {
 		if (iov[i].iov_len > 0) {
 			n = read(fd, iov[i].iov_base, iov[i].iov_len);
-			if (n == -1)
+			if (n < 0) {
+				if (tot_len > 0 && (errno == EINTR || errno == EAGAIN))
+					break;
 				return -1;
+			}
 
 			tot_len += n;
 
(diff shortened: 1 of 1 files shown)


sys/uio.c
```diff
@@ -42,16 +42,21 @@ ssize_t readv(int fd, const struct iovec *iov, int iovcnt)
...
+	}
 
 	/* TODO: file locking */
 
 	for (i = 0; i < iovcnt; ++i) {
 		if (iov[i].iov_len > 0) {
 			n = read(fd, iov[i].iov_base, iov[i].iov_len);
-			if (n == -1)
+			if (n < 0) {
+				if ((tot_len > 0 && errno == EINTR) || errno == EAGAIN)
```
agkaminski, line 56: Shouldn't we return `-1` in case of `EAGAIN`?

sys/uio.c
```diff
@@ -69,16 +74,21 @@ ssize_t writev(int fd, const struct iovec *iov, int iovcnt)
...
+	}
 
 	/* TODO: file locking */
 
 	for (i = 0; i < iovcnt; ++i) {
 		if (iov[i].iov_len > 0) {
 			n = write(fd, iov[i].iov_base, iov[i].iov_len);
-			if (n == -1)
+			if (n < 0) {
+				if ((tot_len > 0 && errno == EINTR) || errno == EAGAIN)
```
agkaminski, line 88: As above, shouldn't we return `-1` in case of `EAGAIN`?

agkaminski:
Add reviewers as reviewers, not assignees.

sys/uio.c
ziemleszcz: Yes if tot_len is zero. If tot_len is non-zero we should return number of bytes written and return EAGAIN on next call. For context check out this code from dropbear:

	written = writev(fd, iov, io_count);

	if (written < 0) {
		if (errno != EINTR && errno != EAGAIN) {
			TRACE(("channel IO write error fd %d %s", fd, strerror(errno)))
			close_chan_fd(channel, fd, SHUT_WR);
			return DROPBEAR_FAILURE;
		}
	} else {
		int cbuf_written = MIN(circ_len1+circ_len2, (unsigned int)written);
		cbuf_incrread(cbuf, cbuf_written);
		if (morelen) {
			*morelen = written - cbuf_written;
		}
		channel->recvdonelen += written;
	}

agkaminski: Ok, so your condition should look like that:
`if (tot_len > 0 && (errno == EINTR || errno == EAGAIN))`
i.e. you should return `tot_len` only, if it's not zero.

agkaminski (approved):

PR #101, opened 2021-06-28
string: add strncat

ziemleszcz:
JIRA: RTOS-62

## Description

Implementation of strncat.

## Motivation and Context

strncat is used by lighttpd, sscep and dropbear.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/string/string.c
+++ b/string/string.c
@@ -447,9 +447,16 @@ int strverscmp(const char *s1, const char *s2)
 	return strcmp(s1, s2);
 }
 
-char *strncat(char *s1, const char *s2i, size_t num)
+
+char *strncat(char *dest, const char *src, size_t n)
 {
-	debug(__func__);
-	debug(" : not implemented\n");
-	return 0;
+	size_t len = strlen(dest);
+	size_t i;
+
+	for (i = 0; i < n && src[i] != '\0'; ++i)
+		dest[len + i] = src[i];
+
+	dest[len + i] = '\0';
+
+	return dest;
 }


kemonats:
1. This version is only good for -Os. (better after taking into account point 3).
2. Instead of arrays it will be better to use pointers (it will be shorter and faster with `-O2`).
3. Instead of calling `strlen()` function you can move pointer ( it will be faster).

agkaminski:
@kemonats Current version have an advantage of being MISRA compliant. Compiler optimalization should make both version identical.

agkaminski (approved):

kemonats:
PR has been accepted but a few comments.
I see that this implementation is taken from man `strncat(3)`.

Example code using pointers taken from the newlib library.
```
char *strncat(char *__restrict s1,
	const char *__restrict s2,
	size_t n)
{
  char *s = s1;

  while (*s1)
    s1++;
  while (n-- != 0 && (*s1++ = *s2++))
    {
      if (n == 0)
	*s1 = '\0';
    }

  return s;
}
```
Code size comparison:
```
armv7a7 -Os
this   42
newlib 38

armv7m7 -O2
this   90
newlib 52

i386 -O2
this   117
newlib 77

riscv64 -O2
this   146
newlib 68
```

Considering code size, are the two versions identical under -O2 optimization?
In terms of speed, the version with pointer offset instead of using strlen() function will be faster.
(as is done in newlib)
In summary, this version when using the -O2 flag is longer. it is also slower by using `strlen()`.

PR #102, opened 2021-06-29
unistd: implement gethostname and sethostname

ziemleszcz:
JIRA: DTR-12

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


include/arch/arm-imx/limits.h
```diff
@@ -49,10 +49,11 @@
...
 #define SSIZE_MAX INT_MAX
 
-#define PTHREAD_STACK_MIN 256
-#define PATH_MAX 1024
-#define ARG_MAX 128
+#define PTHREAD_STACK_MIN  256
+#define PATH_MAX           1024
+#define ARG_MAX            128
 #define _POSIX2_RE_DUP_MAX 255
+#define HOST_NAME_MAX      255
```
agkaminski, line 56: Kernel uses this value - it should be defined in kernel. Libphoenix uses kernel header, not vice versa

ziemleszcz: OK, where do do think it should go in the kernel? I have to include kernel header because HOST_NAME_MAX must be available in "limit.h" (POSIX).

agkaminski: This file looks perfect for this: https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/master/include/posix.h

agkaminski: All constants and structures shared with kernel should be defined in this directory

anglov: @agkaminski remember that C and POSIX requirees that you avoid unneeded defines in headers (underscore and big letter or another underscore is allowed). You cannot easy include kernel posix.h in limits.h because you pollute limits.h header and viiolate rule about reserved identifiers.

agkaminski: Ok, then maybe we need to create limits.h in kernel include? This would solve this problem.

anglov: I agree with that. It seems as a better approach

include/sys/utsname.h
```diff
@@ -16,20 +16,23 @@
...
+	char version[16];
+	char machine[16];
 };
 
 
 static inline int uname(struct utsname *name)
 {
 	memset(name, 0, sizeof(*name));
+	gethostname(name->nodename, sizeof(name->nodename));
 	return 0;
```
agkaminski, line 36: Maybe `return gethostname(name->nodename, sizeof(name->nodename));` so we use retval of `gethostname()`? (errno should be set by `WRAP_ERRNO_DEF`, so that's ok)

agkaminski (approved):

agkaminski:
CI should pass now, lets see if it will

agkaminski (changes requested):
It seems that phoenix/limits.h is not installed. It need fixing first (in phoenix-rtos-kernel repo)

ziemleszcz:
phoenix-rtos-project has old version of phoenix-rtos-kernel, check https://github.com/phoenix-rtos/phoenix-rtos-project/pull/125.

agkaminski (approved):

nalajcie:
I would gladly see some unit tests regarding `sethostname` and `gethostname` (normal function and checking what happens near the `HOST_NAME_MAX` limit).

PR #103, opened 2021-07-01
setjmp: fix alignments

jsarzy:
## Description

1. Use ldr instead of addw
2. Add storing mask
3. Fix stack alignment before syscall
## Motivation and Context

1. In the Thumb mode PC points to the address of an executing instruction + 2 next instructions (4 bytes). But, read of PC value in 
`add` instruction is always word-aligned (when instructions are half-word aligned). So, if line `addw r1, pc, #((1f - . - 4) | 1)` is half-word aligned then `r1` has value moved by 2 bytes. (A5.1.2 Reference Manual ARMv7m or [here](https://www.keil.com/support/man/docs/armasm/armasm_dom1361289861747.htm))
2. Just add missing mask store.
3. When pushing context to stack, the hardware saves eight 32-bit words, comprising xPSR, ReturnAddress, LR (R14), R12, R3, R2, R1, and R0. If stack is not aligned to 8 bytes and `CCR.STKALIGN` is set 1 (by default is 1) then additionally stack is aligned to 8 bytes. Kernel does not take into account this alignment so make sure stack is aligned before syscall. (B1.5.6-7 Reference Manual ARMv7m)

I'm not an expert of ARM architecture so I'd appreciate any feedback if I handled it right.
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (armv7m7-imxrt106x using [this test](https://github.com/phoenix-rtos/phoenix-rtos-tests/blob/master/setjmp/setjmp.c)).

## Checklist:

## Special treatment

### DIFF
--- a/arch/armv7m/jmp.S
+++ b/arch/armv7m/jmp.S
@@ -23,7 +23,7 @@
 .type _setjmp, %function
 _setjmp:
 	/* set jump address and stack pointer */
-	addw r1, pc, #((1f - . - 4) | 1) /* 4 is the size of addw instruction itself */
+	ldr r1, =(1f + 1)
 	mov r2, sp
 	/* store everythig in jmpbuf */
 	stm r0, {r1-r2, r4-r11, lr}
@@ -62,17 +62,21 @@ setjmp:
 	mov r3, r0
 	pop {r0, lr}
 	/* set jump address and stack pointer */
-	addw r1, pc, #((1f - . - 4) | 1) /* 4 is the size of addw instruction itself */
+	ldr r1, =(1f + 1)
 	mov r2, sp
 	/* store everythig in jmpbuf */
-	stm r0, {r1-r2, r4-r11, lr}
+	stmia r0!, {r1-r2, r4-r11, lr}
+	/* save signal mask */
+	str r3, [r0]
 	mov r0, #0
 	bx lr
 1:
 	/* this is where we land after the jump
 	 * r0 is now pointing to signal mask
 	 * which we want to restore */
-	push {r1, r4, lr}
+
+	 /* push r2 to align stack to 8 bytes before call */
+	push {r1-r2, r4, lr}
 	ldr r4, [r0]
 	mov r0, r4
 	mov r4, r3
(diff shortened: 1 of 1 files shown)


arch/armv7m/jmp.S
```diff
@@ -62,24 +62,29 @@ setjmp:
...
+	str r3, [r0]
 	mov r0, #0
 	bx lr
 1:
 	/* this is where we land after the jump
 	 * r0 is now pointing to signal mask
 	 * which we want to restore */
 	push {r1, r4, lr}
+	/* make sure SP is aligned to 8 bytes */
+	sub sp, #4
```
agkaminski, line 79: You can do this in one operation quite easly - you push (and later pop) one register extra, e.g. r2:
`push {r1-r2, r4, lr}`

arch/armv7m/jmp.S
```diff
@@ -62,24 +62,29 @@ setjmp:
...
 	ldr r4, [r0]
 	mov r0, r4
 	mov r4, r3
 	ldr r1, =0xffffffff
 	bl signalMask
 	/* set return value and pop the stack */
 	mov r0, r4
+	add sp, #4
 	pop {r1, r4, lr}
```
agkaminski, line 88: `pop {r1-r2, r4, lr}`

arch/armv7m/jmp.S
```diff
@@ -23,7 +23,7 @@
 .type _setjmp, %function
 _setjmp:
 	/* set jump address and stack pointer */
-	addw r1, pc, #((1f - . - 4) | 1) /* 4 is the size of addw instruction itself */
+	ldr r1, =(1f + 1)
```
agkaminski, line 26: Can we do `ldr r1, =(1f | 1)` instead? I'm not sure if we can, but it would be more reliable (if for some reason should 1f have thumb bit set already)

jsarzy: We can't, syntax error. Docs says that `ldr r1, =1f` should be sufficient in Thumb mode but when I disassemble it doesn't have last bit set.

agkaminski (approved):

agkaminski:
It seems that armv7a might require the same fix

PR #117, opened 2021-07-27
resolve_path: fix one letter leaf resolve

niewim19:
JIRA: PD-160

## Description

Change in `unistd/dir.c` in `resolve_path()` that fixes some issues with path resolving mainly:
 - one character missing leaf
 - path with missing branch (e.g `existing/missing/existing/file.txt`)

## Motivation and Context

This PR is a byproduct of `psh` autocompletion PR: 
 - https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/78

`psh` tries to autocomplete path and filenames when `TAB` is used and path validation is done with `resolve_path()`. After changes to `resolve_path` problems occured with autocompletion in `psh`. During adapting `psh` to new `resolve_path()` two major flawes were found:
 - resolving path where missing leaf is one character long (e.g resolving path like: `b` or `bin/b`) resulted in wrong resolve
 - resolving paths where not a leaf, but part of branch is missing resulted in the whole path being resolved, while it would be logical to return `NULL` because path with missing branch cannot be resolved. **This needs to be discussed in this PR as `resolve_path` is Phoenix-RTOS specific**

Together with this change there is need for change in tests for `resolve_path()`:
 - https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/40

## Types of changes

## How Has This Been Tested?

- [x] tests modified: (https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/40).

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work (https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/40)

### DIFF
--- a/unistd/dir.c
+++ b/unistd/dir.c
@@ -156,8 +156,8 @@ static int _resolve_abspath(char *path, char *result, int resolve_last_symlink, 
 		}
 		else {
 			char *rem = slash;
-			while (*rem++ == '/')
-				;
+			while (*rem == '/')
+				rem++;
 			if (*rem == 0)
 				is_leaf = 1; /* trailing '/' */
 


damianloew:
Without this change "if [ -d "/bin/nonExiststentDirectory" ]" in bash script is true. It helps busybox test suite to be run properly.

nalajcie:
Ohh, I see why this would make `/etc/x` being resolved to `/etc` (off-by-one bug), thanks for finding and fixing this.
I would welcome test for this particular use case (to prevent regressions in the future), eg:
```c
         check_and_free_str("/x", resolve_path("/x", NULL, 1, 1));
         check_and_free_str("/etc/x", resolve_path("/etc/x", NULL, 1, 1));
```
(this bug is also present in `canonicalize_file_name` but for test it we would need to create single-char file (which might also be a good idea).

I don't see how this would affect more than single-char leaf.

Actually, I've checked it on `ia32-generic` on current master:
image
So @damianloew `test -d "/bin/nonExiststentDirectory"` should be false even without this patch

nalajcie (dismissed):
Approving, but would welcome more tests for fixed paths (in respective PR)

damianloew:
> Ohh, I see why this would make `/etc/x` being resolved to `/etc` (off-by-one bug), thanks for finding and fixing this.
> I would welcome test for this particular use case (to prevent regressions in the future), eg:
> 
> ```c
>          check_and_free_str("/x", resolve_path("/x", NULL, 1, 1));
>          check_and_free_str("/etc/x", resolve_path("/etc/x", NULL, 1, 1));
> ```
> 
> (this bug is also present in `canonicalize_file_name` but for test it we would need to create single-char file (which might also be a good idea).
> 
> I don't see how this would affect more than single-char leaf.
> 
> Actually, I've checked it on `ia32-generic` on current master:
> image
> So @damianloew `test -d "/bin/nonExiststentDirectory"` should be false even without this patch

That's right, sorry. I tested it with single char, but i wrote like that for better readability.

unistd/dir.c
```diff
@@ -156,7 +156,7 @@ static int _resolve_abspath(char *path, char *result, int resolve_last_symlink,
 		}
 		else {
 			char *rem = slash;
-			while (*rem++ == '/')
+			while (*++rem == '/')
```
nalajcie, line 159: after sleeping on it, maybe we should write it in "longer" form to be more readable:
```c
while (*rem == '/')
    rem++;
```

This seems less mind-bending for me :). But I'm leaving the decision up to You.

niewim19: I will go with your proposition as it is more clear and more DO-178 compliant. Thanks!

nalajcie (approved):
Approving and merging. Please update test cases as requested :+1:.

nalajcie:
... aaand do it ASAP as the tests will start failing as soon as somebody would update submodules in phoenix-rtos-project :)

niewim19:
Updating tests in a minute

PR #124, opened 2021-08-19
unistd: raise SIGPIPE signal on write() EPIPE error

### LINKED ISSUE #166: busybox: 'ash: write error: EPIPE' when piping some large text variable into head command
Screenshot from 2021-08-12 13-00-55
To reproduce:

1. edit /etc/rc.psh in rootskel like that (for posixsrv run in background):
Screenshot from 2021-07-14 14-50-28
2. run system image
3. run /bin/psh /etc/rc.psh (for posixsrv run in background)
4. run /bin/sh
5. run text='yes abc | head -c 9999'
6. run echo "$text" | head -c 5

Maxez:
## DO NOT MERGE THIS YET, IT'S JUST A PROPOSED SOLUTION
`EPIPE` and `SIGPIPE` in `write()` as described by [OpenGroup](https://pubs.opengroup.org/onlinepubs/007904975/functions/write.html):
> 
> [EPIPE]
>     An attempt is made to write to a pipe or FIFO that is not open for reading by any process, or that only has one end open. A SIGPIPE signal shall also be sent to the thread.
> [EPIPE]
>     A write was attempted on a socket that is shut down for writing, or is no longer connected. In the latter case, if the socket is of type SOCK_STREAM, a SIGPIPE signal shall also be sent to the thread.

Things to consider/change:
- Should every `write()` `EPIPE` error generate `SIGPIPE` signal?
No. As mentioned above writing to shut down `SOCK_DGRAM`(UDP) socket should return `EPIPE` error without `SIGPIPE` signal.
- Should `SIGPIPE` be sent to the writing thread or process (all process threads)?
As mentioned above `SIGPIPE` should be sent to the writing thread only.

## Description

Changed `write()` to raise `SIGPIPE` signal on `EPIPE` error.

## Motivation and Context

Fixes phoenix-rtos/phoenix-rtos-project#166
[JIRA: RTOS-84](https://jira.phoenix-rtos.com/browse/RTOS-84)

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/unistd/file.c
+++ b/unistd/file.c
@@ -21,6 +21,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <libgen.h>
+#include <signal.h>
 #include <sys/msg.h>
 #include <sys/file.h>
 #include <sys/types.h>
@@ -35,11 +36,12 @@ extern int sys_open(const char *filename, int oflag, ...);
 extern int sys_mkfifo(const char *filename, mode_t mode);
 extern int sys_link(const char *path1, const char *path2);
 extern int sys_unlink(const char *path);
+extern int sys_write(int fildes, const void *buf, size_t nbyte);
 extern int sys_pipe(int fildes[2]);
 extern int sys_fstat(int fd, struct stat *buf);
 
 WRAP_ERRNO_DEF(int, read, (int fildes, void *buf, size_t nbyte), (fildes, buf, nbyte))
-WRAP_ERRNO_DEF(int, write, (int fildes, const void *buf, size_t nbyte), (fildes, buf, nbyte))
+// WRAP_ERRNO_DEF(int, write, (int fildes, const void *buf, size_t nbyte), (fildes, buf, nbyte))
 WRAP_ERRNO_DEF(int, close, (int fildes), (fildes))
 WRAP_ERRNO_DEF(int, ftruncate, (int fildes, off_t length), (fildes, length))
 WRAP_ERRNO_DEF(off_t, lseek, (int fildes, off_t offset, int whence), (fildes, offset, whence))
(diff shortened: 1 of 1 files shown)


agkaminski:
@nalajcie if we add such wrapper for send, sendto and sendmsg will this be ok? It's a simple solution and fits nicely in our architecture, question is - is it enough?

nalajcie:
I've found some more insights: 
https://riptutorial.com/posix/example/17424/handle-sigpipe-generated-by-write---in-a-thread-safe-manner

Seems that on MacOS / BSD there is a socket option (`SO_NOSIGPIPE`) to disable signals for particular socket which would impair the current approach (You would need to call additional syscall for each `write()`). On Linux tere is `MSG_DONTWAIT` flag for `send` / `sendmsg` (and socket option is not supported) so this approach might be good enough - we probably need to decide which API we would like to support. The current approach doesn't sit well with me because of additional syscall which in kernel would be a no-op (most complex programs will have `SIGPIPE` ignored) - but I agree that it fits with our current approach.

On a side note - it's written in the link provided that POSIX-2004 requires us to send signal to the correct thread (not process) which is something we should take into the consideration in the future.

agkaminski:
@nalajcie About sending signal to the right thread - our raise already does this - it send a signal specifically to the thread that called raise, so we're covered here

nalajcie:
> @nalajcie About sending signal to the right thread - our `raise` already does this - it send a signal specifically to the thread that called raise, so we're covered here. It is required by POSIX for `raise` to behave this way anyway.

Oh, that's great! I don't know how advanced our signal implementation is, do we support per-thread signal masking also (`phtread_sigmask` as in https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_sigmask.html)?

PR #128, opened 2021-09-06
armv7m: rewrite setjmp

jsarzy:
## Description
Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/176

1. `longjmp` now uses `setjmp` context, no need for pc-relative instructions
2. fpu registers are stored in the `env` buffer
3. additionally savesig flag is stored in `env` buffer, thanks to that we can use one `sigsetjmp` function for both `setjmp` and `_setjmp`

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/arch/armv7m/jmp.S
+++ b/arch/armv7m/jmp.S
@@ -7,91 +7,108 @@
  * _setjmp, _longjmp, setjmp, sigsetjmp
  *
  * Copyright 2018, 2019 Phoenix Systems
- * Author: Kamil Amanowicz, Aleksander Kaminski
+ * Author: Kamil Amanowicz, Aleksander Kaminski, Jakub Sarzyński
  *
  * This file is part of Phoenix-RTOS.
  *
  * %LICENSE%
  */
 
+ /* setjmp env buffer description
+  * Offset	Description
+  * 0		Signal mask flag - if != 0 then next value stores a signal mask
+  * 4		Signal mask
+  * 8		SP register
+  * 12-44	r4-r11 registers
+  * 44		LR register
+  * 48-80	d8-d15 registers
+  * 80		fpscr register
+  */
+
 #define __ASSEMBLY__
 
 .thumb
 .syntax unified
 
-.globl _setjmp
-.type _setjmp, %function
-_setjmp:
-	/* set jump address and stack pointer */
-	ldr r1, =(1f + 1)
-	mov r2, sp
-	/* store everythig in jmpbuf */
-	stm r0, {r1-r2, r4-r11, lr}
-	mov r0, #0
-	bx lr
-1:
-	/* this is where we land after the jump
-	 * just pass the return value which is in r3 */
-	mov r0, r3
-	bx lr
-.size _setjmp, .-_setjmp
-
-.globl _longjmp
-.type _longjmp, %function
-_longjmp:
-	/* move return value to r3.
-	 * longjmp should not make setjmp return 0 though */
-	cmp r1, #0
-	ite eq
-	moveq r3, #1
-	movne r3, r1
-	/* restore registers from jmpbuf */
-	ldmia r0!, {r1-r2, r4-r11, lr}
-	mov sp, r2
-	bx r1
-.size _longjmp, .-_longjmp
-
-.globl setjmp
-.type setjmp, %function
-setjmp:
-	push {r0, lr}
-	/* get signal mask */
-	mov r0, #0
-	mov r1, #0
-	bl signalMask
-	mov r3, r0
-	pop {r0, lr}
-	/* set jump address and stack pointer */
-	ldr r1, =(1f + 1)
-	mov r2, sp
-	/* store everythig in jmpbuf */
-	stmia r0!, {r1-r2, r4-r11, lr}
-	/* save signal mask */
-	str r3, [r0]
-	mov r0, #0
-	bx lr
-1:
-	/* this is where we land after the jump
-	 * r0 is now pointing to signal mask
-	 * which we want to restore */
-
-	 /* push r2 to align stack to 8 bytes before call */
-	push {r1-r2, r4, lr}
-	ldr r4, [r0]
-	mov r0, r4
-	mov r4, r3
-	ldr r1, =0xffffffff
-	bl signalMask
-	/* set return value and pop the stack */
-	mov r0, r4
-	pop {r1-r2, r4, lr}
-	bx lr
-.size setjmp, .-setjmp
 
 .globl sigsetjmp
 .type sigsetjmp, %function
 sigsetjmp:
-	cmp r1, #0
-	bne setjmp
-	b _setjmp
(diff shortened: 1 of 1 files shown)


arch/armv7m/jmp.S
```diff
@@ -7,91 +7,118 @@
...
-	/* store everythig in jmpbuf */
-	stm r0, {r1-r2, r4-r11, lr}
+
+.globl sigsetjmp
+.type sigsetjmp, %function
+sigsetjmp:
+	/* Store signal mask if savesigs != 0 */
+	mov r3, #0
+	sub r2, r1, #0
+	beq 1f
```
kemonats, line 40: Which statement sets the `Z` flag?
`subs r2, r1, #0`?

arch/armv7m/jmp.S
```diff
@@ -7,91 +7,118 @@
...
+
+	/* Get signal mask */
+	/* We don't need to keep r1 */
+	push {r0, lr}
```
kemonats, line 44:
```suggestion
	push {r0, r2, lr}
```
You need to keep the `r2` register.

arch/armv7m/jmp.S
```diff
@@ -7,91 +7,118 @@
...
 	mov r0, #0
-	bx lr
+	mov r1, #0
+	bl signalMask
+	mov r3, r0
+	pop {r0, lr}
```
kemonats, line 49:
```suggestion
	pop {r0, r2, lr}
```

arch/armv7m/jmp.S
```diff
@@ -7,91 +7,118 @@
...
-	bx r1
-.size _longjmp, .-_longjmp
+	beq 1f
 
-.globl setjmp
-.type setjmp, %function
-setjmp:
+	/* Restore mask */
+	/* We don't need to keep r1 */
 	push {r0, lr}
```
kemonats, line 99:
```suggestion
	push {r0, r3, lr}
```
You need to keep the r3 register.

arch/armv7m/jmp.S
```diff
@@ -7,91 +7,118 @@
...
-	/* get signal mask */
-	mov r0, #0
-	mov r1, #0
+	mov r0, r2
+	ldr r1, =0xffffffff
```
kemonats, line 101: This is the same:
```
mov r1, -1
```

arch/armv7m/jmp.S
```diff
@@ -7,91 +7,118 @@
...
 	bl signalMask
-	mov r3, r0
 	pop {r0, lr}
```
kemonats, line 103:
```suggestion
	pop {r0, r3, lr}
```

arch/armv7m/jmp.S
jsarzy: Yes, missing s. I changed a implementation a little bit in favour of `cmp`.

arch/armv7m/jmp.S
jsarzy: thx, now all registers are preserved.

arch/armv7m/jmp.S
jsarzy: Changed, thx

arch/armv7m/jmp.S
```diff
@@ -7,91 +7,116 @@
...
+.globl sigsetjmp
+.type sigsetjmp, %function
+sigsetjmp:
+	/* Store signal mask if savesigs != 0 */
+	mov r2, #0
+	cmp r1, #0
+	beq 1f
+
+	/* Get signal mask */
+	sub sp, #4
```
kemonats, line 43: Why `sub sp, #4`?

arch/armv7m/jmp.S
```diff
@@ -7,91 +7,116 @@
...
 1:
-	/* this is where we land after the jump
-	 * just pass the return value which is in r3 */
-	mov r0, r3
+	/* Store mask (fake or real) */
+	stmia r0!, {r1, r2}
+
+	/* Store general registers */
+	mov r2, sp
+	stmia r0!, {r2, r4-r11, lr}
```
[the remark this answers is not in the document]
jsarzy: Procedure Call Standard for the Arm Architecture ([here](https://developer.arm.com/documentation/ihi0042/j/?lang=en#aapcs32-section5-1)) states that:
> A subroutine must preserve the contents of the registers r4-r8, r10, r11 and SP (and r9 in PCS variants that designate r9 as v6).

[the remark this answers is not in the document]
jsarzy: Procedure Call Standard for the Arm Architecture ([here](https://developer.arm.com/documentation/ihi0042/j/?lang=en#aapcs32-section5-1)) states that:
> A subroutine must preserve the contents of the registers r4-r8, r10, r11 and SP (and r9 in PCS variants that designate r9 as v6).

arch/armv7m/jmp.S
jsarzy: Stack must be aligned to 8 bytes before syscall

kemonats: Ok. I wanted to make sure.
Instead of:
```
sub sp, #4
...
add sp, #4 
```
can be done like this:
```
push {r0, r1, r3, lr}
...
pop {r0, r1, r3, lr}
```

arch/armv7m/jmp.S
kemonats, line 58: Instead of two  `stmia` one
```suggestion
	/* Store mask (fake or real) and general registers */
	mov r3, sp
	stmia r0!, {r1-r11, lr}
```

arch/armv7m/jmp.S
```diff
@@ -7,91 +7,116 @@
...
-.globl setjmp
-.type setjmp, %function
-setjmp:
-	push {r0, lr}
-	/* get signal mask */
-	mov r0, #0
-	mov r1, #0
+	/* Restore mask */
+	sub sp, #4
+	push {r0, r1, lr}
```
kemonats, line 96: Is it necessary to preserve the `lr` register? In line 105 the current value is overwriten.

arch/armv7m/jmp.S
jsarzy: I am in favour of leaving here `sub/add` - these instructions say directly: "hey we need to change stack pointer, it must be aligned". Pushing/poping an unnecessary register may lead to confusion.

Wait for other reviews.

agkaminski: push {r0, r1, r3, lr} version is better. We can't have stack unaligned at any time - interrupt can occur between sub and push - we'll enter irq with unaligned stack (some context is stored on user stack on interrupts).

agkaminski:
I agree with @kemonats remarks. It is essential to keep stack aligned to 8 bytes _at all times__, because we can have an interrupt between any two instructions.

agkaminski (approved):
LGTM

PR #142, opened 2021-11-30
lookup: add temporary fix for errors introduced in 86482f2b828a755be215447a8c0a0442a7a4f42b

Maxez:
## Description

Fixed errors introduced in previous lookup usage fix.

## Motivation and Context

[JIRA: DTR-180](https://jira.phoenix-rtos.com/browse/DTR-180)

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7a7-imx6ull.

## Checklist:

## Special treatment


unistd/dir.c
```diff
@@ -368,20 +368,25 @@ DIR *opendir(const char *dirname)
...
 
-	if (!dirname[0] || (safe_lookup(canonical_name, &s->oid, NULL) < 0)) {
+	if (!dirname[0] || (safe_lookup(canonical_name, &s->oid, &mnt) < 0)) {
 		free(canonical_name);
 		free(s);
 		errno = ENOENT;
 		return NULL; /* ENOENT */
 	}
 
+	/* TODO: add proper mount point detection */
```
agkaminski, line 386: This is actually pretty proper mount point detection.

unistd/dir.c
```diff
@@ -368,20 +368,25 @@ DIR *opendir(const char *dirname)
...
+	if ((s->oid.port != mnt.port) || (s->oid.id != mnt.id))
```
nalajcie, line 387: seems like something we should do with every access to the file? (are current changes enough to eg. remove file from mounted partition?

If it's always needed, it should be taken care of inside lookup IMHO

agkaminski: @nalajcie Lookup is a universal function - it returns both file/dir and device/fs mounted on this file/dir. It is responsibility of a lookup user to one oid or another (do we intend to access the file/dir or a dev/fs mounted there?). If nothing is mounted then both oids are equal. What would you like to change about lookup behaviour?

nalajcie: I assume the problem is that if a file is on a mounted filesystem in `/`,  the `dir` points to the directory oid in the outer system instead of the `/` in the mounted partiton (so is actually invalid, as this file does not exist in the `dir` returned by the lookup?)

agkaminski (approved):
LGTM

PR #143, opened 2021-11-30
unistd/file.c: open() on a directory with write access - set errno to EISDIR

Maxez:
## Description

Added POSIX compliant behaviour for open(), fopen() and freopen() on a directory.

## Motivation and Context

[JIRA: RTOS-132](https://jira.phoenix-rtos.com/browse/RTOS-132)

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-generic, armv7a7-imx6ull.

## Checklist:

## Special treatment

### DIFF
--- a/unistd/file.c
+++ b/unistd/file.c
@@ -110,6 +110,7 @@ int unlink(const char *path)
 int open(const char *filename, int oflag, ...)
 {
 	va_list ap;
+	struct stat st;
 	mode_t mode = 0;
 	int err;
 	char *canonical;
@@ -118,6 +119,16 @@ int open(const char *filename, int oflag, ...)
 	va_start(ap, oflag);
 	mode = va_arg(ap, mode_t);
 	va_end(ap);
+
+	if (oflag & (O_WRONLY | O_RDWR)) {
+		if ((err = stat(filename, &st)) < 0) {
+			if (errno != ENOENT)
+				return err;
+		}
+		else if (S_ISDIR(st.st_mode)) {
+			return SET_ERRNO(-EISDIR);
+		}
+	}
 
 	/* allow_missing_leaf = 1 -> open() may be creating a file */
 	canonical = resolve_path(filename, NULL, 1, 1);


stdio/file.c
```diff
@@ -146,9 +146,22 @@ static ssize_t safe_write(int fd, const void *buf, size_t size)
 
 static int safe_open(const char *path, int oflag, mode_t mode)
 {
+	struct stat s;
 	int err;
-	while ((err = open(path, oflag, mode)) < 0 && errno == EINTR)
+
+	if (oflag & (O_WRONLY | O_RDWR | O_APPEND)) {
```
nalajcie, line 152: `O_APPEND` should not be here, POSIX requires to always use one of the basic open flags (`O_APPEND` is an additional one)

nalajcie (commented):
What about `open()`? POSIX requires for it to also return `EISDIR` error on trying to open dirs for writing.

image

So maybe this functionality should in fact be part of `open()` implementation?

agkaminski:
@nalajcie You're right, it should be implemented in open(), it will take care of fopen() too

Maxez:
I moved proposed checks to `open()` implementation - now `open()`, `fopen()` and `freopen()` should all correctly handle directories with write access.

agkaminski (approved):

stdio/file.c
Maxez: Changed as requested. Please note that in that case we won't properly handle fopen() on an existing directory with "a" mode. The "a" mode gets translated to O_APPEND | O_CREAT (so doesn't pass our write access check) and we will successfully open directory in append mode.
I did check this case on Linux - it does set errno to EISDIR and return -1 in that case.

Maxez: Actually, shouldn't append mode ("a") [here](https://github.com/phoenix-rtos/libphoenix/blob/bd09e6dc360d76ef1b9facda9ecefa335aa5e58a/stdio/file.c#L83) be translated to O_APPEND | O_WRONLY | O_CREAT?
That would solve the issue.

Maxez: Added append mode flags in another [PR](https://github.com/phoenix-rtos/libphoenix/pull/146)

nalajcie (approved):
Seems valid and IMHO should support symlinks correctly.

I would welcome some unit test cases for these particularities (the path is already paved with `resolve_path` unit tests so it's easy to add).

PR #147, opened 2021-12-02
exec: correctly resolve shebang scripts in PATH

nalajcie:
Fixes: phoenix-rtos/phoenix-rtos-project#186

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `imx6ull`.

## Checklist:

## Special treatment

### DIFF
--- a/unistd/sys.c
+++ b/unistd/sys.c
@@ -23,6 +23,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <errno.h>
+#include <limits.h>
 #include <stdarg.h>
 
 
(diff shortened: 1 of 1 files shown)


kemonats:
Out of this PR, but this function does not check the number of bytes actually read by the `read()` function.

nalajcie:
> Out of this PR, but this function does not check the number of bytes actually read by the `read()` function.

Yeah, You're right, safe `read()` should be used here (and probably inside `shebang` also), but I'll merge the current PR anyway as it fixes other problems ;).

PR #149, opened 2021-12-13
execv: check if file is executed

niewim19:
JIRA: PD-213

## Description

Added check in `execv()` function that checks if path (after resolving) leads to a regular file as it is possible to _execute a directory_ which is nonsense

## Motivation and Context

Issue from which this PR arised:
 - https://github.com/phoenix-rtos/phoenix-rtos-project/issues/248

As stated in documentation:
 > The exec family of functions shall replace the current process image with a new process image. The new image shall be constructed from a regular, executable file called the new process image file.

So i added check via `stat()` if resolved path is indeed regular file.

## Types of changes

## How Has This Been Tested?

- [ ] Tested by hand on: (ia32-generic).

## Checklist:

## Special treatment

### DIFF
--- a/unistd/sys.c
+++ b/unistd/sys.c
@@ -16,6 +16,7 @@
 #include <sys/types.h>
 #include <sys/threads.h>
 #include <sys/msg.h>
+#include <sys/stat.h>
 #include <unistd.h>
 #include <fcntl.h>
 #include <ctype.h>
@@ -47,7 +48,7 @@ char exec_buffer[256];
 static int shebang(const char *path)
 {
 	int fd;
-	char sb[2] = {0, 0};
+	char sb[2] = { 0, 0 };
 
 	if ((fd = open(path, O_RDONLY)) < 0)
 		return -1;
@@ -76,6 +77,7 @@ int execve(const char *file, char *const argv[], char *const envp[])
 	char *path = getenv("PATH");
 	char filepath_buffer[PATH_MAX];
 	int filename_len = strlen(file);
+	struct stat buf;
 
 	fflush(NULL);
 
@@ -138,7 +140,13 @@ int execve(const char *file, char *const argv[], char *const envp[])
 	if ((canonical_path = resolve_path(file, NULL, 1, 0)) == NULL)
 		return -1; /* errno set by resolve_path */
 
-	err = exec(canonical_path, argv, envp);
+	/* execute only if it is regular file */
+	err = stat(canonical_path, &buf);
+	if (!err) {
+		/* TODO: check execution bit presence when native chmod is available */
+		err = (S_ISREG(buf.st_mode)) ? exec(canonical_path, argv, envp) : -EACCES;
+	}
+
 	free(canonical_path);
 	free(sb_args);
 


unistd/sys.c
```diff
@@ -138,7 +140,13 @@ int execve(const char *file, char *const argv[], char *const envp[])
 	if ((canonical_path = resolve_path(file, NULL, 1, 0)) == NULL)
 		return -1; /* errno set by resolve_path */
 
-	err = exec(canonical_path, argv, envp);
+	/* check if it is file */
+	err = stat(canonical_path, &buf);
+	if (!err && !S_ISREG(buf.st_mode))
+		err = -ENOENT;
```
agkaminski, line 146:
```suggestion
		err = -ENOEXEC;
```

agkaminski: Behaviour confirmed on linux/glibc

agkaminski (approved):

unistd/sys.c
```diff
@@ -60,6 +61,22 @@ static int shebang(const char *path)
 	return fd;
 }
 
+static int iself(const char *path)
```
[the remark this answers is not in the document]
niewim19: obviously not an error, but I will change it so it does not produce an spelling error :)

niewim19:
in `unistd: execve()` added check if we are executing an elf file. Further down in kernel it is also checked, however there is some kind of wrong propagation of error which leads to system hangup after trying to execute a non-ELF file. SO check here, however redundant, is necessary for quality.

Additional issue will be created

unistd/sys.c
```diff
@@ -60,6 +61,22 @@ static int shebang(const char *path)
...
+	unsigned char elfhead[4] = { 0x7F, 'E', 'L', 'F' };
+
+	if ((fd = open(path, O_RDONLY)) < 0)
+		return -1;
+
+	if ((read(fd, head, 4) < 0) || memcmp(head, elfhead, 4)) {
+		close(fd);
+		return -1;
+	}
+	return 0;
```
kemonats, line 77:
```suggestion
	if ((read(fd, head, 4) < 4) || memcmp(head, elfhead, 4)) {
		close(fd);
		return -1;
	}
	close(fd);
	return 0;
```
Missing `close(fd);`
You should check how many bytes were actually read.

niewim19: Thank you! Added missing `close()`

unistd/sys.c
```diff
@@ -60,6 +61,23 @@ static int shebang(const char *path)
 	return fd;
 }
 
+static int checkelf(const char *path)
```
agkaminski, line 64: Wouldn't this screw up scripts executions (#!/bin/bash)?

niewim19: If we execute a script this comes into play:
```c
if ((fd = shebang(file)) >= 0) {
(...)
	while (*interp == ' ') ++interp;
	end = interp;
	while (*end && !isspace(*end)) ++end;
	*end = 0;
(...)
	file = interp;
	argv = sb_args;
}
```

which substitutes the `file` path with interpreter path (for `#!/bin/bash` -> `/bin/bash`) which is later resolved to interpreter binary file which is good.

Also I checked against `psh` scripts adding a file called `abc` which contains only:
> !/bin/psh

and it produces this output:
> psh: /abc is not a psh script

which means that passing to script works well. After all, syscall `exec()` is **only** for binary files. Only arguments change so user sees it as _executing a script_

nalajcie:
IMHO ELF verification should not be done in userspace as nothing prevents developer from using syscall directly (and this adds another time penalty for `exec()`). if we know there is a bug in error propagation in kernel, we should fix it there.

I'm not sure about whether we should verify if a file is a normal file in kernel also or it should/could be done in userspace only.

Side note: we're not checking anywhere if a file has executable bit set but I assume it's not related to this PR (mentioning it as it could easily be added now that we're calling `stat()` just before the exec).

damianloew:
Running scripts seems to work properly with these changes. Tested using the busybox test suite.

unistd/sys.c
```diff
@@ -60,6 +61,23 @@ static int shebang(const char *path)
...
+{
+	int fd;
+	unsigned char head[4];
+	unsigned char elfhead[4] = { 0x7F, 'E', 'L', 'F' };
+
+	if ((fd = open(path, O_RDONLY)) < 0)
+		return -1;
+
+	if ((read(fd, head, 4) < 0) || memcmp(head, elfhead, 4)) {
```
kemonats, line 73:
```suggestion
	if ((read(fd, head, 4) < 4) || memcmp(head, elfhead, 4)) {
```
Why give up checking the actual number of bytes read?
Why compare against the ELF header if less than 4 bytes were read.

unistd/sys.c
```diff
@@ -60,6 +61,23 @@ static int shebang(const char *path)
 	return fd;
 }
 
+static int checkelf(const char *path)
+{
```
agkaminski, line 65: We can drop this after/if https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/269 is merged

unistd/sys.c
niewim19: Yes. Good notice. Unfortunately after recent commit to kernel this is handled in kernel so while `checkelf()` addition is being dropped. However thank you for an advice!

agkaminski (approved):

unistd/sys.c
niewim19: Mentioned commit merged. Whole `checkelf()` addition was removed from commit

niewim19:
Removed ELF check as it is now correctly handled after https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/269

Merging without executable bit check as our native `psh` does not have any control over it (no `chmod` equivalent yet) and maybe this should be done also in kernel (so that we won`t execute restricted file via direct call to `exec()` from process with lower restrictions)

PR #161, opened 2022-01-26
fopen(): fix closing and errno

niewim19:
JIRA: PD-235

## Description

- Adding necessary file closing if `fopen()` fails to allocate memory after `safe_open()`
- move `mode == NULL` check to `string2mode()`
- accompany `string2mode()` change with correct errno returned when used

## Motivation and Context

- https://github.com/phoenix-rtos/phoenix-rtos-project/issues/298

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/stdio/file.c
+++ b/stdio/file.c
@@ -50,7 +50,10 @@ static int string2mode(const char *mode)
 {
 	int next_char = 1;
 
-	if (mode[0] == 'r') {
+	if (mode == NULL) {
+		return -1;
+	}
+	else if (mode[0] == 'r') {
 		if (mode[next_char] == 'b')
 			next_char += 1;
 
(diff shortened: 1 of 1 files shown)


stdio/file.c
```diff
@@ -259,19 +256,21 @@ FILE *freopen(const char *pathname, const char *mode, FILE *stream)
 {
 	int m;
 
-	if (mode == NULL || stream == NULL)
+	if (stream == NULL) {
+		errno = EBADF;
```
agkaminski, line 260: _EBADF: The mode with which the file descriptor underlying the stream was opened does not support the requested mode when pathname is a null pointer._

This isn't really good return value in this case. Case of `stream` being `NULL` is not handled by the standard and most implementations don't check for `NULL` - they crash instead. This approach have it's benefits - user should check `stream` for `NULL` (after `fopen`), so `NULL` shouldn't be passed here ever. So this check is redundant and causes (very slight) performance hit.

IMHO - `EINVAL` or drop this check altogether.

niewim19: But how about:
>The freopen() function shall fail if:
[EBADF] The file descriptor underlying the stream is not a valid file descriptor when pathname is a null pointer. 

However I will change it to `EINVAL` as suggested. We need this check as then there is `fflush()` which will not throw an error when passed `NULL` an will flush `stdout`. So we might need this check to not flush stdout for no reason if function fails.

niewim19:
Changes:
 - `freopen()` called with `stream = NULL` returns `EINVAL`

agkaminski (approved):
LGTM

niewim19:
Dropped `freopen()` checking `stream == NULL` completely

PR #163, opened 2022-01-28
getline(): correct size handling + little refactor

niewim19:
Fixes: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/274
JIRA: PD-240

## Description

 - remodel code of `getline()` where it handles not initialized buffer and too small size so that:
   - it does not `free()` buffer for no reason if it is too small to hold read line and `realloc()` fails
   - it does not change `*n` if buffer is not resized (memory masking from issue)
   - after successful getline `*n` was too small by one 

## Motivation and Context

https://github.com/phoenix-rtos/phoenix-rtos-project/issues/274

## Types of changes

## How Has This Been Tested?

- [x] New test added:
  - https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/73
- [x] Tested by hand on: ia32-generic

## Checklist:

## Special treatment

### DIFF
--- a/stdio/file.c
+++ b/stdio/file.c
@@ -29,6 +29,7 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <unistd.h>
+#include <limits.h>
 
 
 #define F_EOF     (1 << 0)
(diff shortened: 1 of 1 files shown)


stdio/file.c
```diff
@@ -845,19 +846,14 @@ ssize_t getline(char **lineptr, size_t *n, FILE *stream)
...
 		linesz++;
 
-	if (*lineptr == NULL)
-		*lineptr = malloc(linesz);
-	else if (*n < linesz) {
-		if ((tmp = realloc(*lineptr, linesz)) == NULL) {
-			free(*lineptr);
-		}
+	if (*lineptr == NULL || *n < linesz) {
+		tmp = (*lineptr ? realloc(*lineptr, linesz) : malloc(linesz));
```
gerard5, line 850: `realloc(NULL, linesz)` is the equivalent of `malloc(linesz)`, so the line may be simplified:
```suggestion
		tmp = realloc(*lineptr, linesz);
```

niewim19: Changed to sole `realloc()`. Thanks!

niewim19:
Changed setting `*n` because it was set to be one to small. `*n` was set fo `linesz-1` and should be `linesz` as it is the size of the buffer, not the string length that is returned.

agkaminski (approved):

PR #164, opened 2022-01-31
string: add strlcat() and strlcpy() implementation

niewim19:
JIRA: PD-189

## Description

Two functions that are BSD-based: `strlcpy()` and `strlcat()` implemented in libphoenix for better micropython port compatibility. 

## Motivation and Context

Phoenix-RTOS supports micropython port which internally uses both `strlcpy()` and `strlcat()` functions. Their source code is artificially added do codebase via patches. This PR aims to reduce patches by implementing these functions in libphoenix.

`strlcpy()` and `strlcat()` short description:

> The strlcpy() and strlcat() functions copy and concatenate strings respectively. They are designed to be safer, more consistent, and less error prone replacements for strncpy(3) and strncat(3). Unlike those functions, strlcpy() and strlcat() take the full size of the buffer (not just the length) and guarantee to NUL-terminate the result (as long as size is larger than 0 or, in the case of strlcat(), as long as there is at least one byte free in dst). Note that a byte for the NUL should be included in size.

Full description: https://linux.die.net/man/3/strlcpy

## Types of changes

## How Has This Been Tested?

- [x] New test added: (https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/54).

## Checklist:

## Special treatment

### DIFF
--- a/string/string.c
+++ b/string/string.c
@@ -5,8 +5,8 @@
  *
  * string/string (generic implementation of string functions)
  *
- * Copyright 2017 Phoenix Systems
- * Author: Pawel Pisarczyk
+ * Copyright 2017, 2022 Phoenix Systems
+ * Author: Pawel Pisarczyk, Mateusz Niewiadomski
  *
  * This file is part of Phoenix-RTOS.
  *
@@ -459,4 +459,31 @@ char *strncat(char *dest, const char *src, size_t n)
 	dest[len + i] = '\0';
 
 	return dest;
+}
+
+
+size_t strlcpy(char *dst, const char *src, size_t size)
+{
+	size_t len = 0;
+
+	if (size) {
+		while (--size && *src) {
+			*(dst++) = *(src++);
+			len++;
+		}
+		*dst = '\0';
+	}
+
+	while (*(src++))
+		len++;
+	return len;
+}
+
+
+size_t strlcat(char *dst, const char *src, size_t size)
+{
+	size_t dstlen = strnlen(dst, size);
+	if (dstlen == size)
+		return size + strlen(src);
+	return dstlen + strlcpy(dst + dstlen, src, size - dstlen);
 }
--- a/include/string.h
+++ b/include/string.h
@@ -93,6 +93,14 @@ extern char *stpcpy(char *dest, const char *src);
 extern char *strncpy(char *dest, const char *src, size_t n);
 
 
+/*  copies up to size - 1 characters from the NUL-terminated string src to dst, NUL-terminating the result */
+extern size_t strlcpy(char *dst, const char *src, size_t size);
+
+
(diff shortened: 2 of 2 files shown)


string/string.c
```diff
@@ -5,7 +5,7 @@
  *
  * string/string (generic implementation of string functions)
  *
- * Copyright 2017 Phoenix Systems
+ * Copyright 2017, 2022 Phoenix Systems
```
agkaminski, line 8: Add yourself to authors

string/string.c
```diff
@@ -460,3 +460,38 @@ char *strncat(char *dest, const char *src, size_t n)
...
+
+	if (dstlen == size)
+		return size + srclen;
+
+	dst += dstlen;
+	size = (srclen < size - dstlen ? srclen + 1 : size - dstlen);
+	memcpy(dst, src, size);
+	dst[size - 1] = '\0';
+
+	return dstlen + srclen;
```
gerard5, line 496: lines 491-496, are what `strlcpy` already do, I think it might be simplified to:
```C
size_t strlcat(char *dst, const char *src, size_t size)
{
    size_t dstlen = strnlen(dst, size);
    if (dstlen == size)
        return size + strlen(src);
    return dstlen + strlcpy(dst + dstlen, src, size - dstlen);
}
```
@agkaminski  what do You think?

string/string.c
```diff
@@ -460,3 +460,38 @@ char *strncat(char *dest, const char *src, size_t n)
...
+	return len;
+}
+
+
+size_t strlcat(char *dst, const char *src, size_t size)
+{
+	size_t srclen = strlen(src);
+	size_t dstlen = strnlen(dst, size);
+
+	if (dstlen == size)
```
agkaminski, line 488: We should perform operation anyway, but resulting string should be not NULL terminated

string/string.c
niewim19: :heavy_check_mark: Proposed change passes tests proposed in:
 - https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/54

agkaminski: Looks good, but we shouldn't return early:
https://linux.die.net/man/3/strlcat
"The strlcat() function appends the NUL-terminated string src to the end of dst. It will append at most size - strlen(dst) - 1 bytes, NUL-terminating the result. "

"Note, however, that if strlcat() traverses size characters without finding a NUL, the length of the string is considered to be size and the destination string will not be NUL-terminated (since there was no space for the NUL)."

niewim19: And thats exactly what will happen:
- strlcat traverses _dst_ not finding NUL at at most _size-1_ index
- in scope of _size_ bytes of _dst_ buffer it is not NULL terminated. It means that wrong _size_ was given (buffer is bigger in reality), or _dst_ is not proper c string (buffer is fully used by previous data).

string/string.c
niewim19: Done, thanks!

agkaminski (approved):

PR #168, opened 2022-02-21
time: add utimes(), lutimes() and futimes() implementation

ziemleszcz:
JIRA: DTR-276

## Types of changes

## How Has This Been Tested?

- [x] New test added: (https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/94).
- [x] Tested by hand on: (imx6ull).

## Checklist:

## Special treatment

https://github.com/phoenix-rtos/phoenix-rtos-filesystems/pull/62
https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/304

### DIFF
--- a/sys/time.c
+++ b/sys/time.c
@@ -14,12 +14,49 @@
  */
 
 #include <sys/time.h>
+#include <sys/msg.h>
+#include <sys/file.h>
 #include <errno.h>
 #include <stdlib.h>
 
 #include "posix/utils.h"
 
-extern int sys_utimes(const char *filename, const struct timeval times[2]);
+
+extern int sys_futimens(int fd, const struct timespec times[2]);
+
+
+/* path needs to be canonical */
+static int utimes_internal(const char *path, const struct timeval times[2])
+{
+	oid_t oid, dev;
+	msg_t msg = { 0 };
+	struct timeval now;
+	int err;
+
+	if (lookup(path, &oid, &dev) < 0)
+		return -ENOENT;
+
+	if (times == NULL)
+		gettimeofday(&now, NULL);
+
+	msg.type = mtSetAttr;
+	msg.i.attr.oid = oid;
+
+	msg.i.attr.type = atMTime;
+	msg.i.attr.val = times == NULL ? now.tv_sec : times[1].tv_sec;
+	err = msgSend(oid.port, &msg);
+	if ((err >= 0) && (msg.o.attr.err >= 0)) {
+		msg.i.attr.type = atATime;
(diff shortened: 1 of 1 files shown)


sys/time.c
```diff
@@ -14,24 +14,58 @@
...
+
+	msg.type = mtSetAttr;
+	msg.i.attr.oid = oid;
+
+	if (times == NULL)
+		gettimeofday(&now, NULL);
+
+	msg.i.attr.type = atMTime;
+	msg.i.attr.val = times == NULL ? now.tv_sec : times[1].tv_sec;
+	if (((err = msgSend(oid.port, &msg)) < 0) || ((err = msg.o.attr.err) < 0))
```
agkaminski, line 47: Assignment inside of if condition is against MISRA rule, we are abandoning this code idiom.

agkaminski (approved):

sys/time.c
nalajcie: hey, do You know if this rule also applies to assignments inside `while` conditions? (should probably apply also, so we need to rewrite our `-EINTR` loops in libphoenix?)

agkaminski: Yeah, I do, missed that one during second reading - to be fixed too.

We need to fix all occurrences, for now during natural code evolution, later as part of making static code analysis pass without errors ;)

niewim19:
As we plan to conform to POSIX as much as possible I have one note:
`futimes()` is not specified in POSIX. `futimens()` is. Difference is basically a name as both functions share the same declaration and `futimes()` errno set is subset of `futimens()`\`s.

[futimens documentation](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/)
[futimes documentation](https://man7.org/linux/man-pages/man3/futimes.3.html)

We can think about aliasing one with another or vice versa. We better have POSIX function that is not 100% compliant rather than no function at all.

nalajcie:
> As we plan to conform to POSIX as much as possible I have one note: `futimes()` is not specified in POSIX. `futimens()` is. Difference is basically a name as both functions share the same declaration and `futimes()` errno set is subset of `futimens()``s.

Thanks for mentioning this - the issue was raised in related kernel PR (https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/304) and @ziemleszcz was kind enough to make `sys_futimens` syscall (which means we will have POSIX-compatible kernel API enabling us to implement both `futimes` and `futimens` in libphoenix (not necessarily as a part of this PR)).

agkaminski (approved):

PR #173, opened 2022-04-08
include: replace standard enums with defines in netdb.h and time.h

damianloew:
## Description

- include: replace standard enums with defines in netdb.h and time.h

## Motivation and Context

There are some ifdefs in azure sdk (incoming port) and openssl checking if the specified values are defined.

JIRA: PD-251

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


include/netdb.h
```diff
@@ -21,6 +21,39 @@
 #include <stdint.h>
 
 
+#define AI_PASSIVE     1 << 0
```
nalajcie, line 24: if there is an expression in macro define, it needs to be inside parentheses `(1 << 0)` to avoid possible problems in the future. See https://godbolt.org/z/fbM539zEW for quick sample why this is bad :).

nalajcie (changes requested):

include/netdb.h
damianloew: Oh, that's right, I've missed it, sorry. It's fixed now.

nalajcie (approved):

PR #179, opened 2022-05-13
include: inttypes: add SCN defines

damianloew:
## Description

 - include: inttypes: add SCN defines
 
## Motivation and Context

JIRA: PD-272

`SCNU64` macro needed in mbedtls incoming port.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (using SCNu, when building mbedtls).

## Checklist:

## Special treatment

### DIFF
--- a/include/arch/armv7/stdint.h
+++ b/include/arch/armv7/stdint.h
@@ -142,9 +142,23 @@ typedef uint64_t uintmax_t;
 
 /* Additional definitions */
 
-#define _PRI_8  ""
-#define _PRI_16 ""
-#define _PRI_32 ""
-#define _PRI_64 "ll"
+#define _PRI_8   ""
+#define _PRI_16  ""
+#define _PRI_32  ""
+#define _PRI_64  "ll"
+#define _SCN_8   "hh"
+#define _SCN_16  "h"
+#define _SCN_32  ""
+#define _SCN_64  "ll"
+#define _PRI_PTR ""
+
+#define _SCN_LEAST8  _SCN_8
+#define _SCN_LEAST16 _SCN_16
+#define _SCN_LEAST32 _SCN_32
+#define _SCN_LEAST64 _SCN_64
+#define _SCN_FAST8   _SCN_32
+#define _SCN_FAST16  _SCN_32
+#define _SCN_FAST32  _SCN_32
+#define _SCN_FAST64  _SCN_64
 
 #endif
--- a/include/arch/riscv64/stdint.h
+++ b/include/arch/riscv64/stdint.h
@@ -146,5 +146,19 @@ typedef uint64_t uintmax_t;
 #define _PRI_16 ""
 #define _PRI_32 ""
 #define _PRI_64 "l"
+#define _SCN_8   "hh"
+#define _SCN_16  "h"
+#define _SCN_32  ""
+#define _SCN_64  "l"
+#define _PRI_PTR "l"
+
+#define _SCN_LEAST8  _SCN_8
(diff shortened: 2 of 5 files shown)


include/arch/armv7/stdint.h
```diff
@@ -146,5 +146,6 @@ typedef uint64_t uintmax_t;
 #define _PRI_16 ""
 #define _PRI_32 ""
 #define _PRI_64 "ll"
+#define _PRI_PTR "ll"
```
nalajcie, line 149: IMHO this target has 32-bit pointers (see `UINTPTR_MAX` for example) so the correct prefix should be `""`. Applies for other 32-bit arch also.

nalajcie (changes requested):

include/arch/armv7/stdint.h
damianloew: I wasn't sure and wanted someone to verify if it's correct. Thanks, changed.

nalajcie (approved):

anglov (changes requested):
This comment is for whole code.

1. Please sort and group defines in sane order. – If you prefer yours, then please reorder current one.
2. To do it correctly you need to use platform-specific prefixes. 
eg. you need additional prefix like _PRI_LEAST(8/16/32/64) – as size of those types is architecture-defined  
3. As you use _PRI in SCN macros, then maybe it should be renamed eg. to _FMT and use valid prefixes for both printf and scanf – currently those print prefixes causes implicit cast to higher type. Hope that nothing will break with that change.

damianloew:
> This comment is for whole code.
> 
> 1. Please sort and group defines in sane order. – If you prefer yours, then please reorder current one.
> 2. To do it correctly you need to use platform-specific prefixes.
>    eg. you need additional prefix like _PRI_LEAST(8/16/32/64) – as size of those types is architecture-defined
> 3. As you use _PRI in SCN macros, then maybe it should be renamed eg. to _FMT and use valid prefixes for both printf and scanf – currently those print prefixes causes implicit cast to higher type. Hope that nothing will break with that change.

Thanks, I've adopted your suggestions. Please check last force push.

include/arch/riscv64/stdint.h
```diff
@@ -146,5 +146,6 @@ typedef uint64_t uintmax_t;
 #define _PRI_16 ""
 #define _PRI_32 ""
 #define _PRI_64 "l"
+#define _PRI_PTR "l"
```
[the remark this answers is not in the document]
damianloew: Thanks, corrected.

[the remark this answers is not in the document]
anglov: Still missing defines – now _SCN_X

anglov (changes requested):

include/arch/riscv64/stdint.h
```diff
@@ -146,5 +146,19 @@ typedef uint64_t uintmax_t;
 #define _PRI_16 ""
 #define _PRI_32 ""
 #define _PRI_64 "l"
+#define _SCN_8   "hh"
+#define _SCN_16  "h"
+#define _SCN_32  ""
+#define _SCN_64  "ll"
```
anglov, line 152: We use (unsigned) long for (u)int64_t type for this platform so you should use l prefix. Also compatible with _PRI_64

anglov (changes requested):

include/arch/riscv64/stdint.h
damianloew: That's right, sorry.

PR #180, opened 2022-05-26
- stdlib: strtol: correct return type in definitions

damianloew:
## Description

- stdlib: strtol: change return type to long int from int

## Motivation and Context

- `strtol()`, `strtoul()` definitions incompatible with declarations.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic).

## Checklist:

## Special treatment

### DIFF
--- a/stdlib/strtoul.c
+++ b/stdlib/strtoul.c
@@ -17,9 +17,9 @@
 #include "ctype.h"
 
 
-unsigned int strtoul(char *nptr, char **endptr, int base)
+unsigned long int strtoul(char *nptr, char **endptr, int base)
 {
-	unsigned int t, v = 0;
+	unsigned long int t, v = 0;
 
 	if ((base == 16 || base == 0) && nptr[0] == '0' && (nptr[1] | 0x20) == 'x') {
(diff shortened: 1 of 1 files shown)


stdlib/strtoul.c
```diff
@@ -62,11 +62,11 @@ int strtol(char *nptr, char **endptr, int base)
 
 int atoi(char *str)
 {
-	return strtol(str, NULL, 10);
+	return (int)strtol(str, NULL, 10);
```
agkaminski, line 65: Potential integer overflow - we migh want to check if retval fits in `int`. 64-bit systems frequently have 64-bit `long` type and 32-bit `int`

damianloew: I am not sure how to resolve that. POSIX specifies `atoi()` like `(int) strtol(str, (char **)NULL, 10)` and there is a following sentence: `If the value cannot be represented, the behavior is undefined`. https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/

agkaminski (approved):

PR #182, opened 2022-06-02
string: Fix strncmp() and strncasecmp()

MaciejPurski:
Currently both functions always return 0, if the first argument is
an empty string. We need to handle the case, when the for loop has
been finished before comparing all 'n' characters.

JIRA: RTOS-193

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/string/string.c
+++ b/string/string.c
@@ -86,6 +86,10 @@ int strncasecmp(const char *s1, const char *s2, size_t n)
 			return 1;
 	}
 
+	if (k < n && (tolower(*p) != tolower(*(s2 + k)))) {
+		return -1;
+	}
+
 	return 0;
 }
 #endif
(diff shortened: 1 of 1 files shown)


string/string.c
```diff
@@ -86,6 +86,9 @@ int strncasecmp(const char *s1, const char *s2, size_t n)
 			return 1;
 	}
 
+	if (k < n && (tolower(*p) != tolower(*(s2 + k))))
+		return -1;
```
agkaminski, line 90: New style is advised for new code
```suggestion
	if (k < n && (tolower(*p) != tolower(*(s2 + k)))) {
		return -1;
	}
```

string/string.c
```diff
@@ -105,6 +108,9 @@ int strncmp(const char *s1, const char *s2, size_t n)
 			return 1;
 	}
 
+	if (k < n && (*p != *(s2 + k)))
+		return -1;
```
agkaminski, line 112:
```suggestion
	if (k < n && (*p != *(s2 + k))) {
		return -1;
	}
```

nalajcie:
I'm glad You added the test case showcasing the problem :P

agkaminski (approved):
Might not be a full fix - some archs have strncmp implementation in asm

PR #195, opened 2022-08-03
Libphoenix: add TLS support.

lukileczo:
## Description

- Creates `errno` in thread-local storage
- Implements ```__aeabi_read_tp()``` for armv7m
## Motivation and Context

JIRA: RTOS-209

## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/125.
- [x] Tested by hand on: `ia32-generic-qemu`, `armv7m7-imxrt106x-evk`, `riscv64-generic-qemu`, `armv7a9-zynq7000-qemu`.

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work https://github.com/phoenix-rtos/phoenix-rtos-build/pull/127, https://github.com/phoenix-rtos/libphoenix/pull/195..


arch/armv7m/tls.c
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <sys/types.h>
+
+
+volatile addr_t arm_tls_ptr __attribute__((section("armtls"), used)) = 0;
```
agkaminski, line 19: `addr_t` is reserved for physical addresses, `void *` or `uintptr_t` instead perhaps?

arch/armv7m/tls.c
```diff
@@ -0,0 +1,34 @@
...
+	 * `push {r1-r3}` must be done first in separate __asm__ call
+	 * so that no instruction before changes these registers.
+	 */
+	__asm__ volatile("push {r1-r3}");
+	__asm__ volatile("mov r0,%[tls];"
+					 "pop {r1-r3};"
+					 :
+					 : [ tls ] "r"(arm_tls_ptr)
+					 :);
+}
```
kemonats, line 34: I think that using inline assembly to define this function is not a good solution.
If this function can only use `r0` register, I suggest you write this function in pure assembler.

lukileczo: Ok, rewritten in assembler.

include/errno.h
```diff
@@ -17,6 +17,7 @@
 #define _LIBPHOENIX_ERRNO_H_
 
 #include <phoenix/errno.h>
+#include <arch.h>
```
nalajcie, line 20: `errno.h` should not include `arch.h`, please note that `arch.h` is needed only for errno implementation, so it should be included in `errno.c`.

IMHO `SUPPORT_TLS` name is also too generic, might be set by accident by other packages (eg. mbedTLS ;)). Not sure we have the convention for internal defines in `libphoenix`, maybe prepend it with `__LIBPHOENIX_` (eg. `__LIBPHOENIX_ARCH_TLS_ENABLED` (or SUPPORTED) ?)

lukileczo: Fixed

arch/armv7m/tls_internal.c
```diff
@@ -0,0 +1,28 @@
...
+volatile uintptr_t arm_tls_ptr __attribute__((section("armtls"), used)) = 0;
+
+
+void *aeabi_read_tp_internal(void)
+{
+	__asm__ volatile("mov r0,%[tls]"
+					 :
+					 : [tls] "r"(arm_tls_ptr)
+					 :);
+}
```
agkaminski, line 28: No new line at the end of the file

agkaminski (commented):
Lookin' good! one small remark regarding endl on the eof

agkaminski (approved):
LGTM, let's wait for tests passing in CI

arch/armv7m/tls.S
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+#define __ASSEMBLY__
+
+.thumb
+.syntax unified
+
+
+.globl __aeabi_read_tp
+.type __aeabi_read_tp,%function
```
kemonats, line 23: I suggest adding
```suggestion
.align 2
.thumb_func
.globl __aeabi_read_tp
.type __aeabi_read_tp,%function
```

arch/armv7m/tls.S
```diff
@@ -0,0 +1,30 @@
...
+/* __aeabi_read_tp must not clobber any registers except r0 */
+__aeabi_read_tp:
+	push {r1-r3, lr}
+	blx aeabi_read_tp_internal
+	pop {r1-r3, lr}
+	bx lr
+.size __aeabi_read_tp, .-__aeabi_read_tp
```
kemonats, line 30: Why not whole function in assemler?

agkaminski: Because we need access to a variable in .data - on armv7m (and NOMMU in general) we use PIC/PIE, so all .bss and .data accesses are doubly indirect - via .got table. We wouldn't be able to do this in asm - we don't know .got table index for this variable.

Other feaseable option is to pass ready pointer to asm, but we need wrapper outside to save r1-r3 from being overwritten (compiler chose r3 for .got access operation, as we determined).

kemonats: **A bit long**

A fragment of the file `tls_internal.S` generated by the compiler:
```
aeabi_read_tp_internal:
.LFB0:
	.file 1 "tls_internal-1.c"
	.loc 1 31 1 view -0
	.cfi_startproc
	@ args = 0, pretend = 0, frame = 0
	@ frame_needed = 0, uses_anonymous_args = 0
	@ link register save eliminated.
	.loc 1 32 2 view .LVU1
	ldr	r3, .L3
	ldr	r3, [r9, r3]
	ldr	r3, [r3]
	.syntax unified
@ 32 "tls_internal-1.c" 1
	mov r0,r3
@ 0 "" 2
	.loc 1 36 1 is_stmt 0 view .LVU2
	.thumb
	.syntax unified
	bx	lr
.L4:
	.align	2
.L3:
	.word	.LANCHOR0(GOT)
	.cfi_endproc
.LFE0:
	.size	aeabi_read_tp_internal, .-aeabi_read_tp_internal
	.global	arm_tls_ptr
	.section	armtls,"aw"
	.align	2
	.set	.LANCHOR0,. + 0
	.type	arm_tls_ptr, %object
	.size	arm_tls_ptr, 4
arm_tls_ptr:
	.space	4
	.text
```
Besides, the function is defined as 
```
void *aeabi_read_tp_internal(void)
```
and there is no return statement, compiler warning:
```
warning: control reaches end of non-void function [-Wreturn-type]
```

Then maybe such solution in the assembler:
```
.thumb
.syntax unified

.align 2
.thumb_func
.globl __aeabi_read_tp
.type __aeabi_read_tp,%function
__aeabi_read_tp:
	ldr r0, 1f
	ldr r0, [r9, r0]
	ldr r0, [r0]
	bx lr
	.align 2
1:
	.word arm_tls_ptr(GOT)
.size __aeabi_read_tp, .-__aeabi_read_tp

.section armtls, "aw"
.align 2
.globl arm_tls_ptr
.type	arm_tls_ptr, %object
arm_tls_ptr:
	.word 0x0
/*
instead
	.word 0
can be
	.space 4
*/
.size	arm_tls_ptr, 4
```
and disassembly:
```
00000000 <__aeabi_read_tp>:
   0:   4802            ldr     r0, [pc, #8]    ; (c <__aeabi_read_tp+0xc>)
   2:   f859 0000       ldr.w   r0, [r9, r0]
   6:   6800            ldr     r0, [r0, #0]
   8:   4770            bx      lr
   a:   bf00            nop
   c:   00000000        .word   0x00000000      c: R_ARM_GOT32  arm_tls_ptr
```

arch/armv7m/tls_internal.c
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <sys/types.h>
+
+
+volatile uintptr_t arm_tls_ptr __attribute__((section("armtls"), used)) = 0;
```
kemonats, line 19: `attribute used` is not needed.
From gcc documentation:
```
used
    This attribute, attached to a variable with static storage, means that the 
    variable must be emitted even if it appears that the variable is not referenced.
```
That is, this attribute is needed if the variable is defined as static.

lukileczo: Ok, thanks, removed it.

arch/armv7m/tls.S
agkaminski: I like that compiler does this PIC/PIE stuff "behind the scenes". Right now this code will work regardless if it's PIC/PIE or if GOT pointer is in R9 or R8, compiler handles this stuff without us hardcoding anything. Because of that I opt to leave this as is.

arch/armv7m/tls_internal.c
```diff
@@ -0,0 +1,28 @@
...
+ * %LICENSE%
+ */
+
+#include <sys/types.h>
+
+
+volatile uintptr_t arm_tls_ptr __attribute__((section("armtls"))) = 0;
+
+
+void aeabi_read_tp_internal(void)
```
agkaminski, line 22: I think this function should be void *aeabi_read_tp_internal(void), as we're returning stuff from it. Not super critical, but it's somewhat erroneous that we're returning stuff from void function. C should save all registers apart from r1-r3 and r12, so adding r12 to push list should allow us to safely do mov %[temp], %[tls] and then return temp.

agkaminski (commented):
Apart from one thing LGTM

arch/armv7m/tls.S
kemonats: In that case, why not simpler:
```
void *aeabi_read_tp_internal(void)
{
	return (void *)arm_tls_ptr;
}
```

agkaminski: Huh, yeah, haven't thought about it, good idea

kemonats: What do you think about this solution:
```
void *read_tls_ptr(void)
{
	return (void*)arm_tls_ptr;
}

void __attribute__((naked)) *__aeabi_read_tp_t5(void)
{
	__asm__ volatile(
		"push {r1-r4,r12,lr};"
		"bl read_tls_ptr;"
		"pop {r1-r4,r12,pc}");
}
```

agkaminski (approved):

PR #196, opened 2022-08-16
make pthread thread safe

lukileczo:
## Description

Add locking to global structs that can be written and read.
## Motivation and Context

- JIRA: RTOS-225
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/pthread/pthread.c
+++ b/pthread/pthread.c
@@ -36,10 +36,14 @@ typedef struct pthread_ctx {
 	struct pthread_ctx *prev;
 	int detached;
 	struct __errno_t e;
+	int refcount;
 } pthread_ctx;
 
 
-static pthread_ctx *pthread_list = NULL;
+static struct {
+	handle_t pthread_list_lock;
+	pthread_ctx *pthread_list;
+} pthread_common;
 
 
 static const pthread_attr_t pthread_attr_default = {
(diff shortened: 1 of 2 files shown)


agkaminski (commented):
Please move pthread_common init to pthread_init (called in _start.c). Also, I think pthread_ctx needs refcounter - we find and get context and then release a mutex. What if other thread do either join or detatch and free(ctx)?

pthread/pthread.c
```diff
@@ -51,6 +55,25 @@ static const pthread_attr_t pthread_attr_default = {
...
+
+
+static void pthread_ctx_put(pthread_ctx *ctx)
+{
+	mutexLock(pthread_common.pthread_list_lock);
+	ctx->refcount--;
+	if (ctx->refcount == 0) {
+		free(ctx);
+	}
+	mutexUnlock(pthread_common.pthread_list_lock);
```
agkaminski, line 73: So we don't hold mutex during free - if refcount <= 0 only we hold this pointer and it has been removed from the list, so it's safe to do so.
```suggestion
	int refcnt;
	
	mutexLock(pthread_common.pthread_list_lock);
	refcnt = --ctx->refcount;
	mutexUnlock(pthread_common.pthread_list_lock);
	
	if (refcnt <= 0) {
		free(ctx);
	}
```

agkaminski (commented):
I'm afraid than pthread_ctx structs should also be treated as global data - many thread can hold pointer to these and make operations concurrently. So mutex should be kept when performing read/writes on these structs.

I suggest making version of pthread_ctx_get/put that need outside locking (_pthread_ctx_get/put), so you don't need to release and reacquire lock when getting/putting context:

_get() { ctx->refcount++; }
get() { mutexLock; _get(); mutexUnlock; }

pthread/pthread.c
```diff
@@ -51,6 +55,38 @@ static const pthread_attr_t pthread_attr_default = {
 };
 
 
+static void __pthread_ctx_get(pthread_ctx *ctx)
+{
+	++ctx->refcount;
+}
+
+
+static void __pthread_ctx_put(pthread_ctx *ctx)
```
agkaminski, line 64: Nitpick - we use single '_' to mark unlocked function. Or is there some reason for two '_'s?
```suggestion
static void _pthread_ctx_get(pthread_ctx *ctx)
{
	++ctx->refcount;
}


static void _pthread_ctx_put(pthread_ctx *ctx)
```

pthread/pthread.c
```diff
@@ -51,6 +55,38 @@ static const pthread_attr_t pthread_attr_default = {
...
+{
+	int refcnt = --ctx->refcount;
+	mutexUnlock(pthread_common.pthread_list_lock);
```
agkaminski, line 67: This is unorthodox, but TBH I don't see any better solution, so for me it's ok

agkaminski (commented):
Nitpick, otherwise LGTM

agkaminski (approved):
LGTM

PR #197, opened 2022-08-17
pthread: implement missing functions

lukileczo:
## Description

Implements functions required for running posix test suite
## Motivation and Context

- JIRA: RTOS-236
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `ia32-generic-qemu.`

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work https://github.com/phoenix-rtos/libphoenix/pull/196.

### DIFF
--- a/include/pthread.h
+++ b/include/pthread.h
@@ -20,6 +20,7 @@
 #include <sys/threads.h>
 #include <sched.h>
 #include <time.h>
+#include <signal.h>
 
 
 #define EDEADLK 38
@@ -190,5 +191,24 @@ static inline pthread_cond_t _pthread_cond_init(void)
 	pthread_cond_init(&cond, NULL);
 	return cond;
 }
+
+
+extern int pthread_sigmask(int how, const sigset_t *__restrict__ set, sigset_t *__restrict__ oset);
+
+
+extern int pthread_kill(pthread_t thread, int sig);
+
+
+extern int pthread_key_create(pthread_key_t *key, void (*destructor)(void *));
+
+
+extern int pthread_key_delete(pthread_key_t key);
+
+
+extern int pthread_setspecific(pthread_key_t key, const void *value);
+
+
+extern void *pthread_getspecific(pthread_key_t key);
+
 
 #endif
--- a/include/sys/types.h
+++ b/include/sys/types.h
@@ -71,6 +71,7 @@ typedef struct pthread_condattr_t {
 	clockid_t clock_id;
 } pthread_condattr_t;
 
+typedef struct __pthread_key_t *pthread_key_t;
 
 /* BSD legacy types permitted by POSIX */
 typedef uint8_t   u_int8_t;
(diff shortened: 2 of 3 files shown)


include/pthread.h
```diff
@@ -191,4 +192,23 @@ static inline pthread_cond_t _pthread_cond_init(void)
...
+extern int pthread_key_delete(pthread_key_t key);
+
+
+extern int pthread_setspecific(pthread_key_t key, const void *value);
+
+
+extern void *pthread_getspecific(pthread_key_t key);
+
+
+#endif
```
agkaminski, line 214: Missing new line at the end of the file

include/sys/types.h
```diff
@@ -79,4 +80,4 @@ typedef uint32_t  u_int32_t;
 typedef uint64_t  u_int64_t;
 typedef int register_t;
 
-#endif
+#endif
```
agkaminski, line 83: Missing new line at the end of the file

pthread/pthread.c
```diff
@@ -201,8 +258,22 @@ void pthread_exit(void *value_ptr)
...
 	pthread_ctx *ctx = (pthread_ctx *)pthread_self();
 
-	if (ctx != NULL)
+	if (ctx != NULL) {
 		ctx->retval = value_ptr;
+		mutexLock(pthread_common.pthread_key_lock);
+		while (ctx->key_data_list != NULL) {
+			pthread_key_data_t *thread_data = ctx->key_data_list;
+			void *value = thread_data->value;
+			void (*destructor)(void *) = thread_data->key->destructor;
```
agkaminski, line 267: Our style is still variable declaration on the beginning of the function, not beginning of the block. This might change, but for now it is

pthread/pthread.c
```diff
@@ -594,3 +665,131 @@ int pthread_cond_timedwait(pthread_cond_t *restrict cond,
...
+	else {
+		(*key)->destructor = destructor;
+	}
+	return err;
+}
+
+
+int pthread_key_delete(pthread_key_t key)
+{
+	int err = EOK;
```
agkaminski, line 700: It seems to me that we never modify err, it always stays EOK. Maybe we should return error if key hasn't been found (and skip free(key) if so)?

lukileczo: I don't think there is a way to determine if key is valid or not, and whether it should be freed. Someone could call `pthread_key_create` - which mallocs memory for` __pthread_key_t` - and then not use this key at all but when `pthread_key_delete` is called the memory still has to be freed. I'll just return `EOK`.

pthread/pthread.c
```diff
@@ -101,27 +142,31 @@ int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
 		return -EAGAIN;
 	}
 
+	pthread_ctx_get(ctx);
```
kemonats, line 145: In function `pthread_ctx_get(ctx)` you increment uninitialized `ctx->refcount` variable.
Why not:
```suggestion
	ctx->refcount = 1;
```

lukileczo: Of course, thanks.

agkaminski (approved):
LGTM

PR #201, opened 2022-09-07
include/math.h: add HUGE_VAL and other missing C99 macros

damianloew:
JIRA: PD-251

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


include/math.h
```diff
@@ -24,6 +24,12 @@
 #define isinf(x)   __builtin_isinf(x)
 #define signbit(x) __builtin_signbit(x)
 
+#define HUGE_VAL  __builtin_huge_val()
+#define HUGE_VALF (__builtin_huge_valf())
+#define HUGE_VALL (__builtin_huge_vall())
+#define INFINITY  (__builtin_inff())
+#define NAN       (__builtin_nanf(""))
```
agkaminski, line 31: Why these need (..) around them?

nalajcie: I think it was made that way because of how it's defined in `glibc`:
https://github.com/bminor/glibc/blob/master/math/math.h#L59

and `newlib`:
https://github.com/bminor/newlib/blob/master/newlib/libc/include/math.h#L19


IMHO this is a coding convention, not a requirement so we can safely omit `(..)`.

damianloew: Yeah, that was the reason. In fact they aren't needed here, so I removed them.

agkaminski (approved):

PR #202, opened 2022-09-19
Add missing functions stubs/declarations for lsb-vsx-2.0-1 (posix test suite)

adamdebek:
## Description

Add few defines and empty function declarations.
## Motivation and Context

Changes needed to add lsb-vsx-2.0-1 (posix test suite) port. This functions should also be implemented because they are part of POSIX standard.
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (list targets here).

## Checklist:

## Special treatment


include/stdlib.h
```diff
@@ -27,6 +27,8 @@
 
 #define ATEXIT_MAX 32
 
+#define MB_CUR_MAX 4
```
nalajcie, line 30: is this needed to be `4`? IMHO the real answer is `1`. Also, maybe define `MB_LEN_MAX` at the same time?

include/ulimit.h
```diff
@@ -0,0 +1,15 @@
+
```
nalajcie, line 1: please add valid file header

include/stdio.h
```diff
@@ -46,6 +46,11 @@
 #define SEEK_END 2
 #endif
 
+#define FOPEN_MAX 16
+#define L_ctermid 9
```
nalajcie, line 50: strange value, please explain in the comment why it's valid

nalajcie (changes requested):
pointed out some small issues to be fixed before the merge

include/stdio.h
adamdebek: FOPEN_MAX is described on opengroup page as follows - Number of streams which the implementation guarantees can be open simultaneously. The value is at least eight.

Open group says about L_ctermid - The symbolic constant L_ctermid is defined in [<stdio.h>], and shall have a value greater than 0.
 
On my ubuntu is defined as 9.

include/stdlib.h
adamdebek: Now I realised MB_CUR_MAX should be 1. MB_LEN_MAX might be 4?

include/stdio.h
nalajcie: I meant - add comment in code :).

This function defines the maximum output len for `ctermid()` function (without `\0`) - I know it's not implemented but it's directly related to device naming convention in our OS. It needs to be able to hold UART device paths  (so `/dev/uartXX` assuming no more than 99 uarts on single HW) and pseudoterminal paths (`/dev/pts/XX...`) defined in `posixsrv` (see https://github.com/phoenix-rtos/phoenix-rtos-posixsrv/blob/master/pty.c#L613). Right now even `/dev/pts/1` would exceed Your limit.

Please make sure it fits at least `/dev/pts/XX` and write so in comment in the code.

posix/stubs.c
```diff
@@ -161,3 +161,28 @@ int flock(int fd, int operation)
...
+
+char *gets(char *str)
+{
+	return 0;
+}
+
+char *tmpnam(char *str)
+{
+	return 0;
+}
```
agkaminski, line 183: I bet there're warnings 
```suggestion
char *gets(char *str)
{
	return NULL;
}

char *tmpnam(char *str)
{
	return NULL;
}
```

include/sys/mount.h
```diff
@@ -16,7 +16,26 @@
...
+#define MS_NOSUID      1 << 1   /* Ignore suid and sgid bits.  */
+#define MS_NODEV       1 << 2   /* Disallow access to device special files.  */
+#define MS_NOEXEC      1 << 3   /* Disallow program execution.  */
+#define MS_SYNCHRONOUS 1 << 4   /* Writes are synced at once.  */
+#define MS_REMOUNT     1 << 5   /* Alter flags of a mounted FS.  */
+#define MS_MANDLOCK    1 << 6   /* Allow mandatory locks on an FS.  */
+#define MS_DIRSYNC     1 << 7   /* Directory modifications are synchronous.  */
+#define MS_NOSYMFOLLOW 1 << 8   /* Do not follow symlinks.  */
+#define MS_NOATIME     1 << 10   /* Do not update access times.  */
+#define MS_NODIRATIME  1 << 11  /* Do not update directory access times.  */
```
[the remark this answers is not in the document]
nalajcie: please write the values in parentheses, eg `(1 << 1)` as otherwise writing eg. `MS_NODEV + MS_NOEXEC` will be resolved to: `1 << 2 + 1 <<  3`   - which (due to c operators precedence) will result in unexpected value 64 (`+` is being more important than `<<`) instead of expected 12.

Hopefully nobody would use `+` in this context (as `|` should be used) but it's always good to use parentheses for non-trivial macro values.

nalajcie (commented):
:exclamation:  please remove `include/.stdio.h.swp` from the PR

The code looks good now, please remove periods (`.`) at the end of the commit message titles -and `needed to compile test suite` suffix (IMHO this is irrelevant to the change and the PR / JIRA ID clearly describes why it's needed).

nalajcie (approved):

PR #203, opened 2022-09-23
pthread: implement missing functions

lukileczo:
JIRA: RTOS-236

## Description

- implement missing pthread functions
## Motivation and Context

JIRA: RTOS-236
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/345 https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/143

### DIFF
--- a/include/sys/types.h
+++ b/include/sys/types.h
@@ -61,10 +61,22 @@ typedef struct pthread_attr_t {
 
 typedef uintptr_t pthread_t;
 
-typedef handle_t pthread_mutex_t;
-typedef void *pthread_mutexattr_t;
+typedef struct {
+	handle_t mutexh;
+	volatile int initialized;
+} pthread_mutex_t;
 
-typedef handle_t pthread_cond_t;
+
+typedef struct pthread_mutexattr_t {
+	int type;
+} pthread_mutexattr_t;
+
+
+typedef struct {
+	handle_t condh;
+	volatile int initialized;
+} pthread_cond_t;
+
 
 typedef struct pthread_condattr_t {
 	int pshared;
@@ -72,6 +84,8 @@ typedef struct pthread_condattr_t {
 } pthread_condattr_t;
 
 typedef struct __pthread_key_t *pthread_key_t;
+
+typedef uint32_t pthread_once_t;
 
 /* BSD legacy types permitted by POSIX */
 typedef uint8_t   u_int8_t;
--- a/pthread/pthread.c
+++ b/pthread/pthread.c
@@ -211,13 +211,13 @@ int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
 		PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, NULL, 0);
 
 	if (stack == MAP_FAILED || stack == NULL)
-		return -EAGAIN;
+		return EAGAIN;
 
 	pthread_ctx *ctx = (pthread_ctx *)malloc(sizeof(pthread_ctx));
 
 	if (ctx == NULL) {
 		munmap(stack, attrs->stacksize);
-		return -EAGAIN;
+		return EAGAIN;
 	}
 
 	ctx->refcount = 1;
@@ -234,7 +234,7 @@ int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
 	int err = beginthreadex(start_point, attrs->priority, stack,
 		attrs->stacksize, (void *)ctx, &ctx->id);
 
-	if (err != EOK) {
+	if (err != 0) {
 		munmap(stack, attrs->stacksize);
 		pthread_ctx_put(ctx);
 		thread = NULL;
@@ -248,19 +248,19 @@ int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
 		mutexUnlock(pthread_common.pthread_list_lock);
 	}
 
-	return err;
+	return -err;
 }
 
 
 int pthread_join(pthread_t thread, void **value_ptr)
 {
 	if (thread == pthread_self())
-		return -EDEADLK;
+		return EDEADLK;
 
 	pthread_ctx *ctx = (pthread_ctx *)thread;
 
 	if (ctx == NULL) {
-		return -EINVAL;
+		return EINVAL;
 	}
 
 	mutexLock(pthread_common.pthread_list_lock);
@@ -273,7 +273,7 @@ int pthread_join(pthread_t thread, void **value_ptr)
 			mutexLock(pthread_common.pthread_list_lock);
 			if (err < 0) {
 				mutexUnlock(pthread_common.pthread_list_lock);
-				return err;
+				return -err;
 			}
 			else if (ctx->id != err) {
 				mutexUnlock(pthread_common.pthread_list_lock);
@@ -298,7 +298,7 @@ int pthread_join(pthread_t thread, void **value_ptr)
 	munmap(ctx->stack, ctx->stacksize);
 	pthread_ctx_put(ctx);
 
-	return EOK;
+	return 0;
 }
 
 
@@ -309,29 +309,29 @@ int pthread_detach(pthread_t thread)
 
 	if (ctx == NULL) {
 		mutexUnlock(pthread_common.pthread_list_lock);
-		return -ESRCH;
+		return ESRCH;
 	}
 
 	if (ctx->detached != PTHREAD_CREATE_JOINABLE) {
 		mutexUnlock(pthread_common.pthread_list_lock);
-		return -EINVAL;
+		return EINVAL;
 	}
 
 	LIST_REMOVE(&pthread_common.pthread_list, ctx);
 
 	_pthread_ctx_put(ctx);
 
-	return EOK;
+	return 0;
 }
 
 
 int pthread_setcancelstate(int state, int *oldstate)
 {
-	int err = EOK;
+	int err = 0;
 	pthread_ctx *ctx = (pthread_ctx *)pthread_self();
 
 	if (state != PTHREAD_CANCEL_ENABLE && state != PTHREAD_CANCEL_DISABLE) {
-		err = -EINVAL;
+		err = EINVAL;
 	}
 	else {
 		mutexLock(pthread_common.pthread_list_lock);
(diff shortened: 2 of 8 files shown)


pthread/pthread.c
```diff
@@ -260,6 +280,81 @@ int pthread_detach(pthread_t thread)
...
+				destructor = thread_data->key->destructor;
+				free(thread_data);
+				if (value != NULL && destructor != NULL) {
+					destructor(value);
+				}
+				ctx->key_data_list = ctx->key_data_list->next;
+			}
+			mutexUnlock(pthread_common.pthread_key_lock);
+			ctx->retval = (void *)PTHREAD_CANCELED;
+			int id = ctx->id;
```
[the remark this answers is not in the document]
agkaminski: What it meant: move `int id;` to the beginning

pthread/pthread.c
```diff
@@ -483,44 +583,138 @@ int pthread_setschedparam(pthread_t thread, int policy,
...
-	return mutexLock(*mutex);
+	if (mutex == NULL) {
+		err = -EINVAL;
+	}
+	else {
+		if (mutex->initialized == 0) {
+			mutexLock(pthread_common.mutex_cond_init_lock);
+			if (mutex->initialized == 0) {
+				mutex->initialized = 1;
+				err = mutexCreate(&mutex->mutexh);
```
agkaminski, line 603: set initialized to 1 only if err == EOK. We might've got -EINTR or some temporary error here, so it might succeed on the next try. Applies to other lazy constructors too

pthread/pthread.c
```diff
@@ -483,44 +583,138 @@ int pthread_setschedparam(pthread_t thread, int policy,
...
 int pthread_mutex_trylock(pthread_mutex_t *mutex)
 {
-	if (mutex == NULL)
-		return -EINVAL;
+	int err = EOK;
 
-	return mutexTry(*mutex);
+	if (mutex == NULL) {
+		err = -EINVAL;
+	}
```
agkaminski, line 621: IMHO there's no reason for this check to be performed

pthread/pthread.c
```diff
@@ -483,44 +583,138 @@ int pthread_setschedparam(pthread_t thread, int policy,
 int pthread_mutex_init(pthread_mutex_t *mutex, const pthread_mutexattr_t *attr)
 {
 	(void)attr;
-
-	return mutexCreate(mutex);
+	mutex->initialized = 1;
```
agkaminski, line 586: We should return -EINVAL on existing mutex

pthread/pthread.c
```diff
@@ -659,20 +908,41 @@ int pthread_cond_timedwait(pthread_cond_t *restrict cond,
...
+				mutex->initialized = 1;
+				err = mutexCreate(&mutex->mutexh);
+			}
+			mutexUnlock(pthread_common.mutex_cond_init_lock);
+		}
+
+		if (err == EOK) {
+			err = condWait(cond->condh, mutex->mutexh, timeout_us);
+
+			while (err == -EINTR) {
```
agkaminski, line 933: I've mislead you - it's handled in the wrapper: https://github.com/phoenix-rtos/libphoenix/blob/6e57d09a03ed8c30af3751abf4c36fb9bc8af59a/sys/threads.c#L34

but, sadly, it's handled very wrong. It returns 0 on EINTR... So I propose two solutions:
- either replace condWait with phCondWait here,
- or fix condWait implementation in scope of this PR (so mutex is taken and -EINTR is returned).

I favour the second option ;)

pthread/pthread.c
```diff
@@ -803,9 +1074,101 @@ void *pthread_getspecific(pthread_key_t key)
...
+	mutexUnlock(pthread_common.pthread_atfork_lock);
+}
+
+
 void _pthread_init(void)
 {
 	mutexCreate(&pthread_common.pthread_key_lock);
 	mutexCreate(&pthread_common.pthread_list_lock);
+	mutexCreate(&pthread_common.pthread_atfork_lock);
+	mutexCreate(&pthread_common.mutex_cond_init_lock);
```
agkaminski, line 1172: Missing pthread_fork_handlers init (in theory it's ok as is, because it's in .bss, but we do that for pthread_list anyway, so let's do that for pthread_fork_handlers too)

pthread/pthread.c
lukileczo: I don't think we can check `mutex->initialized` here because it might not have been initialized at all - `pthread_mutex_init` should be called only after `pthread_mutex_t` variable declaration, without assigning it `PTHREAD_MUTEX_INITIALIZER`

agkaminski: Ok, so it would be just an error from user and we can't really do anything about it (as mutex can have random data if it was created on stack)

include/sys/types.h
```diff
@@ -61,10 +61,18 @@ typedef struct pthread_attr_t {
...
+	volatile int initialized;
+} pthread_mutex_t;
+typedef struct pthread_mutexattr_t {
+	int type;
+} pthread_mutexattr_t;
+
+typedef struct {
+	handle_t condh;
+	volatile int initialized;
+} pthread_cond_t;
```
agkaminski, line 75:
```suggestion
typedef struct {
	handle_t mutexh;
	volatile int initialized;
} pthread_mutex_t;


typedef struct pthread_mutexattr_t {
	int type;
} pthread_mutexattr_t;


typedef struct {
	handle_t condh;
	volatile int initialized;
} pthread_cond_t;

```

pthread/pthread.c
```diff
@@ -105,6 +134,49 @@ static void start_point(void *args)
...
+
+	mutexLock(pthread_common.pthread_key_lock);
+
+	while (ctx->key_data_list != NULL) {
+		thread_data = ctx->key_data_list;
+		ctx->key_data_list = ctx->key_data_list->next;
+
+		next = malloc(sizeof(_pthread_key_cleanup_t));
+		if (next == NULL) {
+			return head;
```
agkaminski, line 150: Exit with mutex taken

pthread/pthread.c
```diff
@@ -105,6 +134,49 @@ static void start_point(void *args)
...
+
+		free(thread_data);
+	}
+	mutexUnlock(pthread_common.pthread_key_lock);
+
+	return head;
+}
+
+
+static void full_key_cleanup(_pthread_key_cleanup_t *head)
```
agkaminski, line 165: Let's use module prefix for static functions too, for consistency

pthread/pthread.c
```diff
@@ -251,6 +324,68 @@ int pthread_detach(pthread_t thread)
...
+int pthread_cancel(pthread_t thread)
+{
+	int err = EOK, id;
+	pthread_ctx *ctx = (pthread_ctx *)thread;
+	_pthread_key_cleanup_t *cleanup;
+
+	if (ctx == NULL) {
+		err = -ESRCH;
+	}
+	else if (thread == pthread_self()) {
```
agkaminski, line 357: Mutex is not taken only on the first `if` path. You could simplify mutex management by going with:
```
if (ctx == NULL) {
 err = -ESRCH;
}
else {
 mutexLock
 if (thread == pthread_self()) {
  ...
 }
 else {
  ...
 }
 mutexUnlock
}
```

pthread/pthread.c
```diff
@@ -482,45 +605,136 @@ int pthread_setschedparam(pthread_t thread, int policy,
...
+	return EOK;
+}
+
+
+int pthread_mutexattr_settype(pthread_mutexattr_t *attr, int type)
+{
+	int err = EOK;
 
-	return resourceDestroy(*mutex);
+	if (type != PTHREAD_MUTEX_DEFAULT && type != PTHREAD_MUTEX_ERRORCHECK &&
```
agkaminski, line 724: You could invert this condition (using de morgan's law) and only on valid params do `attr->type = type` and `err = EOK`, this way there's no need for `else`

pthread/pthread.c
```diff
@@ -603,37 +818,95 @@ int pthread_condattr_getclock(const pthread_condattr_t *restrict attr, clockid_t
 
 int pthread_cond_init(pthread_cond_t *restrict cond, const pthread_condattr_t *restrict attr)
 {
-	return condCreate(cond);
+	cond->initialized = 1;
```
agkaminski, line 821: Should be cond->initialized = 1 only if err == 0, as with mutex init

pthread/pthread.c
```diff
@@ -603,37 +818,95 @@ int pthread_condattr_getclock(const pthread_condattr_t *restrict attr, clockid_t
...
+	if (cond->initialized == 0) {
+		mutexLock(pthread_common.mutex_cond_init_lock);
+		if (cond->initialized == 0) {
+			err = condCreate(&cond->condh);
+			if (err == EOK) {
+				cond->initialized = 1;
+			}
+		}
+		mutexUnlock(pthread_common.mutex_cond_init_lock);
+	}
```
agkaminski, line 893: How about refactoring the common part of lazy resource allocation? Right now there is code duplication between this and other functions that can cause lazy allocation.

pthread/pthread.c
```diff
@@ -603,37 +818,95 @@ int pthread_condattr_getclock(const pthread_condattr_t *restrict attr, clockid_t
...
+				mutex->initialized = 1;
+			}
 		}
+		mutexUnlock(pthread_common.mutex_cond_init_lock);
+	}
+
+	if (err == EOK) {
+		do {
+			err = condWait(cond->condh, mutex->mutexh, 0);
+		} while (err == -EINTR);
```
agkaminski, line 909: Handled in condWait (phCondWait wrapper)

pthread/pthread.c
```diff
@@ -803,9 +1099,102 @@ void *pthread_getspecific(pthread_key_t key)
...
+	}
+	mutexUnlock(pthread_common.pthread_atfork_lock);
+}
+
+
+void _pthread_atfork_child(void)
+{
+	mutexLock(pthread_common.pthread_atfork_lock);
+	pthread_fork_handlers_t *first = pthread_common.pthread_fork_handlers,
+							*curr = pthread_common.pthread_fork_handlers;
```
agkaminski, line 1178:
```suggestion
	pthread_fork_handlers_t *first = pthread_common.pthread_fork_handlers,
		*curr = pthread_common.pthread_fork_handlers;
```

PR #208, opened 2022-10-10
unistd/getopt.c: fix getopt() function

kemonats:
- added condition to check optind <= 0
 - improved handling of options --
 - for return getopt() to initial state set optind = 0

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/unistd/getopt.c
+++ b/unistd/getopt.c
@@ -34,13 +34,18 @@ int getopt(int argc, char * const argv[], const char *optstring)
 	char *optspec;
 	int leading_colon;
 
+	if (optind <= 0) {
+		optind = 1;
+		optwhere = 1;
+	}
+
 	if (argc == 0 || argv == NULL || optind >= argc)
 		return -1;
 
 	if (argv[optind] == NULL || argv[optind][0] != '-' || argv[optind][1] == '\0')
 		return -1;
 
-	if (argv[optind][1] == '-') {
+	if ((argv[optind][1] == '-') && (argv[optind][2] == '\0')) {
 		optind++;
 		return -1;
 	}


nalajcie (approved):
IMHO a valid fix - `newlib` and `glibc` does the same (resets internal state on `optind == 0`), also busybox sets `optind = 0` to reset `getopt` state in case of NOFORK aplets.

`--` was handled improperly (`--x` was parsed as `--`), now it's better. Still, I'm not sure how we should react to `--x` - right now we'll write `illegal option -- -` which is not really helpful :).

I would like to have these fixed code paths tested, @damianloew can You change Your `getopt test` PR test for propper `getopt` resetting and for `--x` case?

LGTM.

damianloew:
> IMHO a valid fix - `newlib` and `glibc` does the same (resets internal state on `optind == 0`), also busybox sets `optind = 0` to reset `getopt` state in case of NOFORK aplets.
> 
> `--` was handled improperly (`--x` was parsed as `--`), now it's better. Still, I'm not sure how we should react to `--x` - right now we'll write `illegal option -- -` which is not really helpful :).
> 
> I would like to have these fixed code paths tested, @damianloew can You change Your `getopt test` PR test for propper `getopt` resetting and for `--x` case?
> 
> LGTM.

Yes, of course, I will cover it in the getopt test. Didn't tested yet, but I suppose we should set optind to 0 instead of 1 now in psh applets, like history. I can change it too.

damianloew:
> > IMHO a valid fix - `newlib` and `glibc` does the same (resets internal state on `optind == 0`), also busybox sets `optind = 0` to reset `getopt` state in case of NOFORK aplets.
> > `--` was handled improperly (`--x` was parsed as `--`), now it's better. Still, I'm not sure how we should react to `--x` - right now we'll write `illegal option -- -` which is not really helpful :).
> > I would like to have these fixed code paths tested, @damianloew can You change Your `getopt test` PR test for propper `getopt` resetting and for `--x` case?
> > LGTM.
> 
> Yes, of course, I will cover it in the getopt test. Didn't tested yet, but I suppose we should set optind to 0 instead of 1 now in psh applets, like history. I can change it too.

Yeah, I verified that this PR and setting optind to 0 in pshaapp resolves the #208 #521 issue :)

damianloew:
> 

About `--x`, it works the same on host:
Screenshot from 2022-10-10 17-00-06

I've added new test cases (the link below) and it seems that after adding changes from this PR our `getopt` will be POSIX compilant :)

https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/139

PR #213, opened 2022-10-30
fprintf.c: fix handling of the variable returned by the write() function

kemonats:
## Description

In the current code, if the `write()` function sets the `errno` variable to the value `EINTR` there will be an exit from the `while` loop, and I guess this loop in such case should be contine.
What remains to be considered is what to do in case the `write()` function returns a value of `0`.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


gerard5 (approved):
Thanks. Your fix is correct. In the case when `write()` returns zero, unpublished yet part (needs discussion about POSIX compliance) it set `errno` to `EIO`. I'll push it after merging your fix, so we all can discuss it.

PR #214, opened 2022-10-31
fprintf: Handle case when write returns zero length

gerard5:
Handle undefined behavior when `wlen = write(.., todo)` returns `wlen == 0` while `todo != 0`. I assume `EIO` for `errno` as it is closest to that some kind of hardware level error occurred. It's also possible that the application tried to perform some operation that is not possible because the device is in wrong state.

JIRA: RTOS-279

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/stdio/fprintf.c
+++ b/stdio/fprintf.c
@@ -44,15 +44,22 @@ static size_t safe_write(int fd, const char *buff, size_t size)
 
 	while (todo != 0) {
 		wlen = write(fd, buff, todo);
-		if (wlen < 0) {
+		if (wlen > 0) {
+			todo -= wlen;
+			buff += wlen;
+			continue;
+		}
+		else if (wlen < 0) {
 			if ((errno == EINTR) || (errno == EAGAIN)) {
 				continue;
 			}
-
 			break;
 		}
-		todo -= wlen;
-		buff += wlen;
+		else {
+			/* Undefined behaviour (wlen==0) */
+			errno = EIO;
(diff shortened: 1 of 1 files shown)


stdio/fprintf.c
```diff
@@ -44,15 +44,22 @@ static size_t safe_write(int fd, const char *buff, size_t size)
...
 				continue;
 			}
-
 			break;
 		}
-		todo -= wlen;
-		buff += wlen;
+		else {
+			/* Undefined behaviour (wlen==0) */
+			errno = EIO;
```
agkaminski, line 60: Imho there's no need to set errno in this case - it's not checked anyway and errno should matter (and be set) in case of returning -1.

agkaminski (commented):
I'm not sure about errno setting, otherwise LGTM

stdio/fprintf.c
gerard5: It's because of line 81: https://github.com/phoenix-rtos/libphoenix/blob/a80d7fd2f00beb2883c2c31ab4eeb39a94d629c8/stdio/fprintf.c#L77-L82
in that case `ctx->error` is set to `-errno`, so if in case `wlen==0` the `errno` is not set some previous value of `errno` or at best `errno=0` may result in invalid `SETERR_NO(ctx->error)`, and may not return `-1` at the end.

agkaminski (approved):

PR #216, opened 2022-11-25
pthread: fix pthread_once mutex locking

lukileczo:
JIRA: RTOS-288

## Description

fixes `pthread_once` deadlocking
## Motivation and Context

JIRA: RTOS-288
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `ia32-generic-qemu`

## Checklist:

## Special treatment

### DIFF
--- a/pthread/pthread.c
+++ b/pthread/pthread.c
@@ -25,6 +25,8 @@
 
 #define CEIL(value, size)	((((value) + (size) - 1) / (size)) * (size))
 
+#define PTHREAD_ONCE_DONE        0
+#define PTHREAD_ONCE_IN_PROGRESS 2
 
 typedef struct pthread_ctx {
 	handle_t id;
@@ -57,6 +59,8 @@ static struct {
 	handle_t pthread_list_lock;
 	handle_t pthread_atfork_lock;
 	handle_t mutex_cond_init_lock;
+	pthread_mutex_t pthread_once_lock;
+	pthread_cond_t pthread_once_cond;
 	pthread_ctx *pthread_list;
 	pthread_fork_handlers_t *pthread_fork_handlers;
 } pthread_common;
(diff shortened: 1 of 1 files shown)


include/sys/types.h
```diff
@@ -85,7 +85,11 @@ typedef struct pthread_condattr_t {
 
 typedef struct __pthread_key_t *pthread_key_t;
 
-typedef uint32_t pthread_once_t;
+typedef struct pthread_once_t {
+	uint32_t once;
+	handle_t lock;
+	volatile int initialized;
```
agkaminski, line 91: Maybe use pthread mutex here and use lazy init code already in place for it?

agkaminski (commented):
Is there any way to free resource (lock) after this? I see that it's not freed right now

include/sys/types.h
```diff
@@ -85,7 +85,10 @@ typedef struct pthread_condattr_t {
 
 typedef struct __pthread_key_t *pthread_key_t;
 
-typedef uint32_t pthread_once_t;
+typedef struct pthread_once_t {
```
nalajcie, line 88: should be unnamed struct or name it differently than the type

agkaminski (approved):

lukileczo:
> can creating new mutex every time when `pthread_once` is called be avoided (especially the most common path - the `init_routine` was already called and finished)?

I don't think it's possible, if we were to check `once_control->once` we'd have to do it under mutex specific to this `pthread_once` (which may not exist at this point) or not destroying the mutex on exit, which results in leaking this resource

nalajcie:
> > can creating new mutex every time when `pthread_once` is called be avoided (especially the most common path - the `init_routine` was already called and finished)?
> 
> I don't think it's possible, if we were to check `once_control->once` we'd have to do it under mutex specific to this `pthread_once` (which may not exist at this point) or not destroying the mutex on exit, which results in leaking this resource

You sure about that? :)

https://github.com/lattera/glibc/blob/master/nptl/pthread_once.c#L139

IMHO using atomic variables in pthreads generally would make the implementation faster but that's a separate topic.

If we don't want to use atomic variables right now we should be able to do it with global lock and three-state once variable (similarly to `glibc` implementation)

agkaminski (approved):

PR #224, opened 2022-11-29
Maxez/feature umount

Maxez:
## Description

Renamed `mount_msg_t` to `mount_i_msg_t`.
Added `mount_o_msg_t` as `mtMount` message result.
Added checking `mtMount` result in `mount()`.

Fixed `mount()` `mode` and `data` parameter types.
Please note that `data` parameter type is `const void *` on Linux. We're using `const char *` type in order to obtain `data` size
with `strlen()` in `mount()` implementation (we assume that `data` is a `NULL` terminated string, no use case for now).

Added `oid_t mnt` field to `mount_msg_t`. Changed mounted device `long id` field to `oid_t dev`.
Added new `mtMountPoint` message for retrieving mountpoint oid from device.

Added `umount()` implementation. The `target` parameter must be a mounted device. Unmounting by mountpoint `target` is not yet supported.

## Motivation and Context

[JIRA: RTOS-290](https://jira.phoenix-rtos.com/browse/RTOS-290)
[JIRA: RTOS-291](https://jira.phoenix-rtos.com/browse/RTOS-291)
[JIRA: RTOS-293](https://jira.phoenix-rtos.com/browse/RTOS-293)
[JIRA: DTR-101](https://jira.phoenix-rtos.com/browse/DTR-101)

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7a7-imx6ull

## Checklist:

## Special treatment

### DIFF
--- a/include/sys/file.h
+++ b/include/sys/file.h
@@ -30,9 +30,15 @@ enum { mtMount = 0xf50, mtUmount, mtSync, mtStat };
 
 typedef struct {
 	long id;
-	unsigned mode;
+	unsigned long mode;
 	char fstype[16];
-} mount_msg_t;
+} mount_i_msg_t;
+
+
+typedef struct {
+	int err;
+	oid_t oid;
+} mount_o_msg_t;
 
 
 extern int fileAdd(unsigned int *h, oid_t *oid, unsigned mode);
--- a/include/sys/file.h
+++ b/include/sys/file.h
@@ -25,11 +25,12 @@ enum { otDir = 0, otFile, otDev, otSymlink, otUnknown };
 enum { atMode = 0, atUid, atGid, atSize, atBlocks, atIOBlock, atType, atPort, atPollStatus, atEventMask, atCTime, atMTime, atATime, atLinks, atDev };
 
 
-enum { mtMount = 0xf50, mtUmount, mtSync, mtStat };
+enum { mtMount = 0xf50, mtUmount, mtSync, mtStat, mtMountPoint };
 
 
 typedef struct {
-	long id;
+	oid_t dev;
+	oid_t mnt;
 	unsigned long mode;
 	char fstype[16];
 } mount_i_msg_t;
--- a/sys/mount.c
+++ b/sys/mount.c
@@ -22,7 +22,7 @@
 #include <sys/msg.h>
 
 
-int mount(const char *source, const char *target, const char *fstype, long mode, char *data)
+int mount(const char *source, const char *target, const char *fstype, unsigned long mode, const char *data)
 {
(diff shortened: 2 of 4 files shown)


include/sys/file.h
```diff
@@ -25,14 +25,21 @@ enum { otDir = 0, otFile, otDev, otSymlink, otUnknown };
...
+	unsigned long mode;
 	char fstype[16];
-} mount_msg_t;
+} __attribute__((packed)) mount_i_msg_t;
+
+
+typedef struct {
+	int err;
+	oid_t oid;
+} __attribute__((packed)) mount_o_msg_t;
```
agkaminski, line 42: What do you think? One pascal less. Just a thought 
```suggestion
} __attribute__((packed)) mount_imsg_t;


typedef struct {
	int err;
	oid_t oid;
} __attribute__((packed)) mount_omsg_t;
```

sys/mount.c
```diff
@@ -64,30 +65,147 @@ int mount(const char *source, const char *target, const char *fstype, long mode,
...
+		msg.i.size = sizeof(dev);
+
+		if (msgSend(dev.port, &msg) < 0) {
+			return SET_ERRNO(-EIO);
+		}
+
+		if (omnt->err < 0) {
+			/* Check for rootfs */
+			if (omnt->err == -ENOENT) {
+				rootfs = 1;
```
agkaminski, line 168: Hmm, why does this indicate rootfs? Perhaps it might, but it can also indicate other things I guess

sys/mount.c
```diff
@@ -22,15 +22,16 @@
 #include <sys/msg.h>
 
 
-int mount(const char *source, const char *target, const char *fstype, long mode, char *data)
+int mount(const char *source, const char *target, const char *fstype, unsigned long mode, const char *data)
 {
 	struct stat buf;
-	oid_t toid, soid, doid, *doidp;
+	oid_t toid, soid, root;
```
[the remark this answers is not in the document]
Maxez: Restored previous name `doid`. Maybe `fsroot` would be better but I don't mind `doid`.

include/sys/file.h
Maxez: I tried to stick to the convention already used in our codebase, e.g. here: [link](https://github.com/phoenix-rtos/phoenix-rtos-devices/blob/9e1c8296894d8f33e7f8f67758cac2353c1b9a51/storage/imx6ull-flash/imx6ull-flashsrv.h#L74-L80)
I don't mind either approach.

agkaminski: Ack

agkaminski (dismissed):

PR #233, opened 2023-02-17
stdlib/strtoul: Refactor implementation

gerard5:
Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/603

This change refactors implementation of `strtoul()` in the following way
* the function returns either the result of the conversion or, if there was a leading minus sign the negation of the result of the conversion represented as an unsigned value unless the original (non-negated) value would overflow.
* in the latter case the `strtoul()` function returns `ULONG_MAX` and sets `errno` to `ERANGE`.
* any white space before the number is eaten.
* in case base value is unsupported errno is set to `EINVAL`.
* in case of no conversion was performed `EINVAL` errno is set.

JIRA: RTOS-367

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic, imxrt1176).

## Checklist:

## Special treatment

### DIFF
--- a/stdlib/strtoul.c
+++ b/stdlib/strtoul.c
@@ -5,58 +5,99 @@
  *
  * strtoul, strtol
  *
- * Copyright 2017-2018 Phoenix Systems
- * Author: Jakub Sejdak, Pawel Pisarczyk, Michal Miroslaw
+ * Copyright 2017-2018, 2023 Phoenix Systems
+ * Author: Jakub Sejdak, Pawel Pisarczyk, Michal Miroslaw, Gerard Swiderski
  *
  * This file is part of Phoenix-RTOS.
  *
  * %LICENSE%
  */
 
+#include <errno.h>
 #include <stdlib.h>
 #include <ctype.h>
+#include <limits.h>
 
 
 unsigned long int strtoul(const char *nptr, char **endptr, int base)
 {
-	unsigned long int t, v = 0;
+	unsigned long int cutoff, result = 0;
+	int cutlim, t, width = 0, negative = 0;
+	const char *sptr = nptr;
 
-	if ((base == 16 || base == 0) && nptr[0] == '0' && (nptr[1] | 0x20) == 'x') {
-		base = 16;
-		nptr += 2;
-	} else if (base == 0 && nptr[0] == '0')
-		base = 8;
-	else if (base == 0)
-		base = 10;
-
-	for (; isdigit(*nptr) || isalpha(*nptr); ++nptr) {
-		t = *nptr - '0';
-		if (t > 9)
-			t = (*nptr | 0x20) - 'a' + 10;
-
-		if (t >= (unsigned long int)base)
-			break;
-
-		v = (v * base) + t;
+	while (isspace(*sptr) != 0) {
+		sptr++;
 	}
 
-	if (endptr != NULL)
-		*endptr = (char *)nptr;
+	if (*sptr == '-') {
+		negative = 1;
+		sptr++;
+	}
+	else if (*sptr == '+') {
+		sptr++;
+	}
 
-	return v;
+	if ((base == 16 || base == 0) && sptr[0] == '0' && (sptr[1] | 0x20) == 'x') {
+		base = 16;
+		sptr += 2;
+	}
+	else if (base == 0 && sptr[0] == '0') {
+		base = 8;
+	}
+	else if (base == 0) {
+		base = 10;
+	}
+
+	if (base < 2 || base > (11 + 'z' - 'a')) {
+		errno = EINVAL;
+		return 0;
+	}
+
+	cutoff = (unsigned long int)(ULONG_MAX) / (unsigned long int)base;
+	cutlim = (unsigned long int)(ULONG_MAX) - (cutoff * (unsigned long int)base);
+
+	while (isalnum(*sptr) != 0) {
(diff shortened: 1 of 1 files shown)


stdlib/strtoul.c
```diff
@@ -5,58 +5,99 @@
...
+			t = (*sptr | 0x20) - 'a' + 10;
+		}
 
-		if (t >= (unsigned long int)base)
+		if (t >= (unsigned long int)base) {
 			break;
+		}
 
-		v = (v * base) + t;
+		if (t > cutoff || (result == cutoff && t > cutlim)) {
```
agkaminski, line 69: Hmm, I can't imagine a condition when t would be bigger than cutoff

stdlib/strtoul.c
```diff
@@ -5,58 +5,99 @@
...
+		result = (result * base) + t;
+		width++;
+		sptr++;
+	}
+
+	if (width < 0) {
+		errno = ERANGE;
+		result = ULONG_MAX;
+	}
+	else if (width == 0 && result == 0) {
```
agkaminski, line 83: I think width == 0 is enough alone

agkaminski (commented):
Shouldn't strtoul give up on '-'?

stdlib/strtoul.c
gerard5: Yup, right.

stdlib/strtoul.c
gerard5: Good catch, should be `result` instead of `t`.

gerard5:
> Shouldn't strtoul give up on '-'?

You may verify on *nix host that `strtoul("-1234") == 18446744073709550382` which is `-1234` (on x86_64 arch), tried on Linux and :heart: OpenBSD :)

agkaminski:
> > Shouldn't strtoul give up on '-'?
> 
> You may verify on *nix host that `strtoul("-1234") == 18446744073709550382` which is `-1234` (on x86_64 arch), tried on Linux and ❤️ OpenBSD :)

Ok, I wasn't sure ;)

agkaminski (dismissed):
LGTM

stdlib/strtoul.c
```diff
@@ -5,58 +5,99 @@
...
 #include <stdlib.h>
 #include <ctype.h>
+#include <limits.h>
 
 
 unsigned long int strtoul(const char *nptr, char **endptr, int base)
 {
-	unsigned long int t, v = 0;
+	unsigned long int cutoff, t, result = 0;
+	int cutlim, width = 0, negative = 0;
```
kemonats, line 25: `cutlim` should probably be like `cutoff` of type `unsigned long int`

stdlib/strtoul.c
```diff
@@ -5,58 +5,99 @@
...
 		base = 10;
+	}
+
+	if (base < 2 || base > (11 + 'z' - 'a')) {
+		errno = EINVAL;
+		return 0;
+	}
+
+	cutoff = (unsigned long)(ULONG_MAX) / (unsigned long)base;
+	cutlim = (unsigned long)(ULONG_MAX) % (unsigned long)base;
```
kemonats, line 57:
```suggestion
	cutlim = (unsigned long)(ULONG_MAX) - (cutoff * (unsigned long)base);
```
On target `a9` one less division.

stdlib/strtoul.c
```diff
@@ -5,58 +5,99 @@
...
 
-	for (; isdigit(*nptr) || isalpha(*nptr); ++nptr) {
-		t = *nptr - '0';
-		if (t > 9)
-			t = (*nptr | 0x20) - 'a' + 10;
+	while (isdigit(*sptr) || isalpha(*sptr)) {
```
kemonats, line 59:
```suggestion
	while (isalnum(*sptr)) {
```

stdlib/strtoul.c
```diff
@@ -5,58 +5,99 @@
...
-	else if (base == 0)
+	}
+	else if (base == 0) {
 		base = 10;
+	}
+
+	if (base < 2 || base > (11 + 'z' - 'a')) {
+		errno = EINVAL;
+		return 0;
+	}
```
kemonats, line 54: Move this check to the begining of the function.

stdlib/strtoul.c
gerard5: The remainder of division`cutlim` is in range of `2 ≤ base ≤ 36`. The right side of assignment is `unsigned long int`, after assignment it is truncated to `int` and because of the range 2 .. 36, truncation does not occur. The result of modulo is `0 ≤ cutlim ≤ 35` (depending on `base`), for me`int` type here is ok, but ... The other thing is variable `t` which shall be `int` and not `unsigned long int`, depending on single digit range `'0' ≤ t ≤ 'z'`, the check `t > cutlim` is then valid.

stdlib/strtoul.c
gerard5: That's right, thanks.

stdlib/strtoul.c
gerard5: Agree.

stdlib/strtoul.c
gerard5: Good catch, thanks.

agkaminski (approved):

agkaminski (approved):

PR #242, opened 2023-03-13
stdio: Add `fgetpos()` and `fsetpos()`

gerard5:
Only the `fgetpos()`, `fsetpos()` function prototypes were available in `libphoenix`, but implementations of the functions were missing. This PR adds implementation of both functions.

There is placeholder for the tests: [phoenix-rtos-tests/libc/stdio_file.c#L689-L696](https://github.com/phoenix-rtos/phoenix-rtos-tests/blob/3ead4a4ed91b28f50e020413ebd2dcf8c0558dde/libc/stdio_file.c#L689-L696) 
Added PR with test https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/175

JIRA: RTOS-392

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/stdio/file.c
+++ b/stdio/file.c
@@ -616,11 +616,17 @@ int fflush(FILE *stream)
 }
 
 
-int fseek_unlocked(FILE *stream, long offset, int whence)
+static off_t fseektell_unlocked(FILE *stream, off_t offset, int whence)
 {
 	fflush_unlocked(stream);
 	stream->flags &= ~F_EOF;
 	return lseek(stream->fd, offset, whence);
+}
+
+
+int fseek_unlocked(FILE *stream, long offset, int whence)
+{
+	return (int)fseektell_unlocked(stream, offset, whence);
 }
 
 
@@ -636,7 +642,7 @@ int fseek(FILE *stream, long offset, int whence)
 
 long int ftell_unlocked(FILE *stream)
 {
-	return (long)fseek_unlocked(stream, 0, SEEK_CUR);
+	return (long int)fseektell_unlocked(stream, 0, SEEK_CUR);
 }
 
 
@@ -652,7 +658,7 @@ long int ftell(FILE *stream)
 
 off_t ftello_unlocked(FILE *stream)
 {
-	return (off_t)fseek_unlocked(stream, 0, SEEK_CUR);
+	return fseektell_unlocked(stream, 0, SEEK_CUR);
 }
 
 
@@ -663,6 +669,17 @@ off_t ftello(FILE *stream)
 	ret = ftello_unlocked(stream);
 	mutexUnlock(stream->lock);
 	return ret;
+}
+
+
+int fgetpos(FILE *stream, fpos_t *pos)
+{
+	off_t ret = ftello(stream);
+	if (ret >= 0) {
+		*pos = (fpos_t)ret;
+		return 0;
+	}
+	return -1;
 }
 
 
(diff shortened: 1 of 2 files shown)


stdio/file.c
```diff
@@ -666,6 +672,21 @@ off_t ftello(FILE *stream)
 }
 
 
+int fgetpos(FILE *stream, fpos_t *pos)
+{
+	off_t ret;
+	if ((stream == NULL) || (pos == NULL)) {
+		return set_errno(EINVAL);
+	}
```
gerard5, line 680: @agkaminski , perhaps these checks are redundant (not necessary). For example, on host, `getpos(f, NULL)` causes a segfault, so if you agree I'll remove them.

agkaminski: Yeah, there's no need for NULL checks

stdio/file.c
```diff
@@ -855,9 +872,25 @@ void rewind(FILE *file)
...
+	mutexLock(stream->lock);
+	ret = fseeko_unlocked(stream, offset, whence);
+	mutexUnlock(stream->lock);
+	return (ret >= 0) ? 0 : -1;
+}
+
+
+int fsetpos(FILE *stream, const fpos_t *pos)
+{
+	return (fseeko(stream, *(const off_t *)pos, SEEK_SET) >= 0) ? 0 : -1;
```
agkaminski, line 893: This cast would become unsafe should `sizeof(off_t) != sizeof(fpos_t)`.
```suggestion
	off_t off = (off_t)(*pos);
	return (fseeko(stream, off, SEEK_SET) >= 0) ? 0 : -1;
```

gerard5: I agree with you that an intermediate variable on the stack would be safer. But what appeals to me more is why the direct cast might be unreadable: "On  some non-NIX systems, an `fpos_t` object may have a complex representation", and on my host it's actually a structure actually.

I've checked that, in case of `libphoenix` currently is true:
`sizeof(off_t) == sizeof(fpos_t)`

##

`off_t` is `long long int` a.k.a. `off64_t`
https://github.com/phoenix-rtos/libphoenix/blob/b2bbdb12dccc098f32981f761ada8d6e05af7eb0/include/sys/types.h#L41-L42

##

`fpos_t` is `long long int` a.k.a. `offs_t` (on all arch)
https://github.com/phoenix-rtos/libphoenix/blob/b2bbdb12dccc098f32981f761ada8d6e05af7eb0/include/stdio.h#L61

https://github.com/phoenix-rtos/libphoenix/blob/b2bbdb12dccc098f32981f761ada8d6e05af7eb0/include/arch/ia32/types.h#L20
https://github.com/phoenix-rtos/libphoenix/blob/b2bbdb12dccc098f32981f761ada8d6e05af7eb0/include/arch/armv7/types.h#L20
https://github.com/phoenix-rtos/libphoenix/blob/b2bbdb12dccc098f32981f761ada8d6e05af7eb0/include/arch/arm-imx/types.h#L20
https://github.com/phoenix-rtos/libphoenix/blob/b2bbdb12dccc098f32981f761ada8d6e05af7eb0/include/arch/riscv64/types.h#L20

agkaminski: If `fpos_t` would become a structure, casting a pointer will cause a nasty bug (hopefully a warning too)

agkaminski (approved):
LGTM

PR #243, opened 2023-03-14
stdio: fix scanf setting position indicator

gerard5:
### Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/639

This change introduces the followng principle of ungetting characters to the stream: first use `ungetc()`, eg. when used with `stdin`, if it fails in case of buffer limit `BUFSIZ` as fallback continue using `fseek` (negative offset from the current position).

NOTE: on `NOMMU` targets `BUFSIZ` is very small (80 characters) when comparing to `MMU` targets where it's size is set to `_PAGE_SIZE`.

JIRA: RTOS-393

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-generic, imxrt1176-nil.

## Checklist:

## Special treatment

### DIFF
--- a/stdio/scanf.c
+++ b/stdio/scanf.c
@@ -100,17 +100,17 @@ static const unsigned char *__sccl(char *tab, const unsigned char *fmt)
 }
 
 
-static int scanf_parse(char *ccltab, const char *inp, char const *fmt0, va_list ap)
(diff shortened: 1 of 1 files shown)


agkaminski (dismissed):
LGTM

PR #248, opened 2023-03-23
stdio/printf: Fix some issues with printf

astalke:
## Description

1. Fix an issue with `format_parse` failing silently. After this change, it returns an error code.
2. Fix an issue with %f allocating too much memory.

## Motivation and Context

`printf` causes issues on some platforms with no MMU and small amount of memory.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/stdio/bignum.c
+++ b/stdio/bignum.c
@@ -15,7 +15,7 @@
 
 #include "bignum.h"
 #include <errno.h>
-
+#include <string.h>
 
 static void bignum_shiftrInternal(bignum_t *bn, const size_t shift)
 {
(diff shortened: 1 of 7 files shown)


stdio/bignum.c
```diff
@@ -37,13 +38,28 @@ static void bignum_shiftrInternal(bignum_t *bn, const size_t shift)
...
+	else if (bn->size <= BIGNUM_STACK_BUFFER_SIZE) {
+		ptr = calloc(newSize, sizeof(*bn->data));
+		if (ptr != NULL) {
+			(void)memcpy(ptr, bn->data, bn->size * sizeof(*bn->data));
+			bn->size = newSize;
+			bn->data = ptr;
+			return 0;
+		}
+	}
+	else {
```
agkaminski, line 54: Case of `bn->size > newSize` and `bn->size == newSize` aren't handled

agkaminski (commented):
One (imho small) remark about `bignum_resize` (I guess we don't really use it to shrink it's size, so it's generally ok, until someone uses it in some other place).

stdio/bignum.c
gerard5: For me missing cases are:
* `bn->size == newSize`  -> no operation 
* `(newSize < bn->Size) && (bn->data != bn->stackBuffer)` -> shrink from `re/c/alloc` to `use-stack`.

I see that way, until `bn->size` and `newSize` are grater than `BIGNUM_STACK_BUFFER_SIZE`, the check `bn->size > newSize` (shrink) and `bn->size < newSize` (grow) are handled (catched) by this `else` block with `realloc()`, but `bn->size == newSize` case should be handled explicitly `just return 0`, of course `realloc()` with reassignement in this block will handle that but it's waste of cycles. However, the case of freeing `realloc'ed` memory when returning to the `use stack memory`case is definitely not handled: `newSize < bn->Size` when `bn->data != bn->stackBuffer`.

stdio/bignum.c
```diff
@@ -104,21 +158,26 @@ static int bignum_cmpInternal(const bignum_t *first, const bignum_t *second)
 int bignum_init(bignum_t *bn, size_t size, bignum_data_t val)
 {
 	bignum_t result = { .data = NULL, .len = 1, .size = size };
-	result.data = calloc(result.size, sizeof(*result.data));
-	if (result.data != NULL) {
-		result.data[0] = val;
-		*bn = result;
-		return 0;
+	if (size > BIGNUM_STACK_BUFFER_SIZE) {
```
agkaminski, line 161: To simplify the logic, how about:
- init bn->data = bn->stackBuffer,
- init bn->size = BIGNUM_STACK_BUFFER_SIZE
- do bignum_resize(bn, size).

sidenote - `result` temp var seems redundant to me

agkaminski (approved):
LGTM

gerard5 (approved):
Verified on HW-NIL, looks good to me.

gerard5:
@astalke @agkaminski After merging there is an issue with this commit https://github.com/phoenix-rtos/libphoenix/pull/249/files function should now return an `int` but returns no value (issue appeared probably due to automatic rebase during merge).

See https://github.com/phoenix-rtos/libphoenix/blob/master/stdio/format.c#L888

astalke:
> @astalke @agkaminski After merging there is an issue with this commit https://github.com/phoenix-rtos/libphoenix/pull/249/files function should now return an `int` but returns no value (issue appeared probably due to automatic rebase during merge).
> 
> See https://github.com/phoenix-rtos/libphoenix/blob/master/stdio/format.c#L888

Fix here: https://github.com/phoenix-rtos/libphoenix/pull/252

PR #259, opened 2023-06-15
file: fflush all output streams on fflush(NULL)

agkaminski:
DONE: RTOS-494

## Description

Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/734

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/stdio/file.c
+++ b/stdio/file.c
@@ -21,6 +21,7 @@
 #include <sys/minmax.h>
 #include <sys/mman.h>
 #include <sys/threads.h>
+#include <sys/list.h>
 
 #include <arch.h>
 #include <stdio.h>
@@ -45,6 +46,12 @@ typedef struct {
 
 
 FILE *stdin, *stdout, *stderr;
+
+
+static struct {
+	FILE *list;
+	handle_t lock;
+} file_common;
 
 
 static int string2mode(const char *mode)
@@ -136,11 +143,16 @@ static void buffFree(void *ptr, size_t size)
 
 static void file_free(FILE *file)
 {
+	mutexLock(file_common.lock);
+	LIST_REMOVE(&file_common.list, file);
+	mutexUnlock(file_common.lock);
+
 	if (file->buffer != NULL && !(file->flags & F_USRBUF)) {
 		buffFree(file->buffer, file->bufsz);
 	}
 
 	resourceDestroy(file->lock);
+
 	free(file);
 }
 
@@ -229,6 +241,11 @@ FILE *fopen(const char *filename, const char *mode)
 	f->fd = fd;
 	f->mode = m;
 	fflush(f);
+
+	mutexLock(file_common.lock);
+	LIST_ADD(&file_common.list, f);
+	mutexUnlock(file_common.lock);
+
 	return f;
 }
 
@@ -268,6 +285,11 @@ FILE *fdopen(int fd, const char *mode)
 	f->fd = fd;
 	f->mode = m;
 	fflush(f);
+
+	mutexLock(file_common.lock);
+	LIST_ADD(&file_common.list, f);
+	mutexUnlock(file_common.lock);
+
 	return f;
 }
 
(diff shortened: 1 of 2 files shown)


stdio/file.c
```diff
@@ -581,37 +603,61 @@ char *fgets(char *str, int n, FILE *stream)
...
+			} while (iter != file_common.list);
+		}
+		mutexUnlock(file_common.lock);
 	}
-
-	if (stream->flags & F_WRITING) {
-		if (stream->bufpos) {
-			full_write(stream->fd, stream->buffer, stream->bufpos);
-			stream->bufpos = 0;
+	else if ((stream->flags & F_WRITING) != 0) {
```
nalajcie, line 626: maybe move to separate func `fflush_single`  or move stream == NULL semantics to `fflush` for better readability (and avoid recursion-at-a-glance)?

stdio/file.c
```diff
@@ -581,37 +603,61 @@ char *fgets(char *str, int n, FILE *stream)
...
 	if (stream == NULL) {
-		fflush_unlocked(stdout);
-		return 0; /* TODO: flush all FILE's */
+		mutexLock(file_common.lock);
+		if (file_common.list != NULL) {
+			FILE *iter = file_common.list;
+			do {
+				mutexLock(iter->lock);
+				if ((iter->flags & F_WRITING) != 0) {
+					ret = fflush_unlocked(iter);
```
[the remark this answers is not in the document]
agkaminski: Yeah, but it's stated that for `fflush(NULL)` _only_ **output** streams should be `fflush`ed. Or do am I missing something?

stdio/file.c
agkaminski: I'd rather not create a new function, really, less better.

Hmmm, I placed this in unlocked to handle `fflush_unlocked(NULL`) too, but I guess I handle it wrong, as I get locks anyway. I wonder, should `fflush_unlocked(NULL)` work too?

agkaminski: I don't see it stated explicitly, but I guess `fflush_unlocked(NULL)` should be handled

agkaminski: "I'd rather not", but in the end I added not one but two functions, lol

nalajcie: hehh, `fflush_unlocked` is linux-specific and implemented only in `glibc` so IMHO we don't need to focus on correctness here. I've checked the `glibc` implementation and `fflush_unlocked(NULL)` is the same as `fflush(NULL)` (acquires per-file locks - https://github.com/lattera/glibc/blob/master/libio/iofflush_u.c ). Surprisingly - there is an unlocked version which is used in cleanup (`fcloseall` implemetation and probably destructors).

I think You can either leave it as-is or remove unlocked NULL version for code simplicity.

stdio/file.c
```diff
@@ -141,6 +148,11 @@ static void file_free(FILE *file)
 	}
 
 	resourceDestroy(file->lock);
+
+	mutexLock(file_common.lock);
```
nalajcie, line 152: first remove from the list, then start destroying. Right now there might be a race with `fflush(NULL)` which will try to acquire destroyed lock?

nalajcie (approved):

stdio/file.c
```diff
@@ -579,39 +601,85 @@ char *fgets(char *str, int n, FILE *stream)
...
-int fflush_unlocked(FILE *stream)
+static int __fflush_one(FILE *stream)
 {
-	if (stream == NULL) {
-		fflush_unlocked(stdout);
-		return 0; /* TODO: flush all FILE's */
+	int ret = 0;
+
+	if ((stream->flags & F_WRITING) != 0) {
+		if (stream->bufpos != 0) {
```
lukileczo, line 609: Why not merge this into one `if`?

agkaminski: I don't want to go to `else` if `stream->bufpos == 0`

agkaminski: I don't want to go to else if stream->bufpos == 0

lukileczo: okay, ofc

lukileczo (approved):

adamdebek (approved):
Now test case with flushed streams in exit() test passes.

PR #260, opened 2023-06-16
 math: Add optional per arch sqrt implementation 

agkaminski:
DONE: RTOS-495

## Description

Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/738

## Motivation and Context

To provide simple, platform independent and hw accelerated if possible implementation of sqrt()

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic-qemu).

## Checklist:

## Special treatment

### DIFF
--- a/math/power.c
+++ b/math/power.c
@@ -115,3 +115,24 @@ double sqrt(double x)
 	return xn * x;
 #endif
 }
+
+
+float sqrtf(float x)
+{
+#ifdef __IEEE754_SQRTF
+	float __ieee754_sqrtf(float x);
+
+	if (x < 0.0f) {
+		errno = EDOM;
+		return -NAN;
+	}
+
+	if (x == 0.0f || x == -0.0f) {
+		return x;
+	}
+
+	return __ieee754_sqrtf(x);
+#else
+	return (float)sqrt(x);
+#endif
+}
--- a/math/power.c
+++ b/math/power.c
@@ -13,9 +13,11 @@
  * %LICENSE%
  */
 
+#include <arch.h> /* Needed for __ieee754_sqrt */
 #include <math.h>
 #include <limits.h>
 #include <errno.h>
+#include <stdlib.h>
 #include "common.h"
 
 
@@ -61,7 +63,7 @@ double pow(double base, double exponent)
 	return res;
 }
 
-/* Uses reciprocal square root method: */
+
 double sqrt(double x)
 {
 	if (x < 0.0) {
@@ -69,22 +71,30 @@ double sqrt(double x)
 		return -NAN;
 	}
 
-	if (x == 0.0 || x == -0.0)
+	if (x == 0.0 || x == -0.0) {
 		return x;
+	}
 
+#ifdef __IEEE754_SQRT
+	double __ieee754_sqrt(double x);
+
+	return __ieee754_sqrt(x);
+#else
+	/* Use reciprocal square root method: */
 	double xn = (double)(1.0f / (float)x);
 
 	/* IEEE-754 compliant: */
 	conv_t *init = (conv_t *)&xn;
 
 	/* +Infinity : */
-	if (init->i.mantisa == 0 && init->i.exponent == 0x7FF)
+	if (init->i.mantisa == 0 && init->i.exponent == 0x7FF) {
 		return x;
+	}
 
 	/* Subnormals: */
-	if (init->i.exponent == 0)
+	if (init->i.exponent == 0) {
 		init->i.exponent = 0x1;
-
+	}
 
 	/* Initial guess: */
 	init->i.mantisa = (init->i.mantisa >> 1);
@@ -97,10 +107,11 @@ double sqrt(double x)
 		init->i.mantisa |= (0x1ull << 51);
 	}
 
-	int i;
 	/* Reciprocal sqare root iters (avoiding division): */
-	for (i = 0; i < 4; ++i)
+	for (int i = 0; i < 4; ++i) {
 		xn = xn * (1.5 -  0.5 * x * xn * xn);
+	}
 
 	return xn * x;
+#endif
 }
--- a/include/arch/arm-imx/arch.h
+++ b/include/arch/arm-imx/arch.h
@@ -40,7 +40,9 @@
 
 static inline double __ieee754_sqrt(double x)
 {
+	/* clang-format off */
 	__asm__ volatile ("vsqrt.f64 %P0, %P1" : "=w"(x) : "w"(x));
+	/* clang-format on */
 
 	return x;
 }
@@ -50,7 +52,9 @@ static inline double __ieee754_sqrt(double x)
 
 static inline float __ieee754_sqrtf(float x)
 {
+	/* clang-format off */
 	__asm__ volatile ("vsqrt.f32 %0, %1" : "=t"(x) : "t"(x));
+	/* clang-format on */
 
 	return x;
 }
--- a/include/arch/arm-imx/arch.h
+++ b/include/arch/arm-imx/arch.h
@@ -33,6 +33,17 @@
 #define __STRNCPY
 #define __MEMMOVE
 
+#if (__ARM_PCS_VFP || (__VFP_FP__ && !__SOFTFP__)) && (__ARM_FP & 8)
+#define __IEEE754_SQRT
+
+static inline double __ieee754_sqrt(double x)
+{
+	__asm__ volatile ("vsqrt.f64 %P0, %P1" : "=w"(x) : "w"(x));
+
+	return x;
+}
+#endif
+
 #define _PAGE_SIZE 0x1000
 #define SIZE_PAGE _Pragma("GCC warning \"'SIZE_PAGE' is deprecated. Use _PAGE_SIZE from arch.h or PAGE_SIZE from limits.h (POSIX only)\"") _PAGE_SIZE
 
--- a/include/arch/arm-imx/arch.h
+++ b/include/arch/arm-imx/arch.h
@@ -33,12 +33,24 @@
 #define __STRNCPY
 #define __MEMMOVE
 
-#if (__ARM_PCS_VFP || (__VFP_FP__ && !__SOFTFP__)) && (__ARM_FP & 8)
+
+#if __ARM_PCS_VFP || (__VFP_FP__ && !__SOFTFP__)
+#if __ARM_FP & 8
 #define __IEEE754_SQRT
 
 static inline double __ieee754_sqrt(double x)
 {
 	__asm__ volatile ("vsqrt.f64 %P0, %P1" : "=w"(x) : "w"(x));
+
+	return x;
+}
+#endif
+
+#define __IEEE754_SQRTF
+
+static inline float __ieee754_sqrtf(float x)
+{
+	__asm__ volatile ("vsqrt.f32 %0, %1" : "=t"(x) : "t"(x));
 
 	return x;
 }
--- a/include/arch/armv7/arch.h
+++ b/include/arch/armv7/arch.h
@@ -33,12 +33,23 @@
 #define __STRNCPY
 #define __MEMMOVE
 
-#if (__ARM_PCS_VFP || (__VFP_FP__ && !__SOFTFP__)) && (__ARM_FP & 8)
+#if __ARM_PCS_VFP || (__VFP_FP__ && !__SOFTFP__)
+#if __ARM_FP & 8
 #define __IEEE754_SQRT
 
 static inline double __ieee754_sqrt(double x)
 {
 	__asm__ volatile ("vsqrt.f64 %P0, %P1" : "=w"(x) : "w"(x));
+
+	return x;
+}
+#endif
+
+#define __IEEE754_SQRTF
+
+static inline float __ieee754_sqrtf(float x)
+{
+	__asm__ volatile ("vsqrt.f32 %0, %1" : "=t"(x) : "t"(x));
 
 	return x;
 }
--- a/include/arch/armv7/arch.h
+++ b/include/arch/armv7/arch.h
@@ -39,7 +39,9 @@
 
 static inline double __ieee754_sqrt(double x)
 {
+	/* clang-format off */
 	__asm__ volatile ("vsqrt.f64 %P0, %P1" : "=w"(x) : "w"(x));
+	/* clang-format on */
 
 	return x;
 }
@@ -49,7 +51,9 @@ static inline double __ieee754_sqrt(double x)
 
 static inline float __ieee754_sqrtf(float x)
 {
+	/* clang-format off */
 	__asm__ volatile ("vsqrt.f32 %0, %1" : "=t"(x) : "t"(x));
+	/* clang-format on */
 
 	return x;
 }
--- a/include/arch/armv7/arch.h
+++ b/include/arch/armv7/arch.h
@@ -33,6 +33,17 @@
 #define __STRNCPY
 #define __MEMMOVE
 
+#if (__ARM_PCS_VFP || (__VFP_FP__ && !__SOFTFP__)) && (__ARM_FP & 8)
+#define __IEEE754_SQRT
+
+static inline double __ieee754_sqrt(double x)
+{
+	__asm__ volatile ("vsqrt.f64 %P0, %P1" : "=w"(x) : "w"(x));
+
+	return x;
+}
+#endif
+
 #define _PAGE_SIZE 0x200
 #define SIZE_PAGE _Pragma("GCC warning \"'SIZE_PAGE' is deprecated. Use _PAGE_SIZE from arch.h or PAGE_SIZE from limits.h (POSIX only)\"") _PAGE_SIZE
 
--- a/include/arch/ia32/arch.h
+++ b/include/arch/ia32/arch.h
@@ -33,6 +33,7 @@ static inline double __ieee754_sqrt(double x)
 {
 	double result;
 
+	/* clang-format off */
 	__asm__ volatile ("fldl %1\n\t" /* put value */
 		"fsqrt\n\t"                 /* calc sqrt */
 		"fxtract\n\t"               /* extract exponent */
@@ -47,6 +48,7 @@ static inline double __ieee754_sqrt(double x)
 		"fstpl %0"                  /* get the result */
 		: "=m"(result)
 		: "m"(x));
+	/* clang-format on */
 
 	return result;
 }
--- a/include/arch/ia32/arch.h
+++ b/include/arch/ia32/arch.h
@@ -26,6 +26,33 @@
 #define __MEMCPY
 #define __MEMSET
 
+#if (!__SOFTFP__)
+#define __IEEE754_SQRT
+
+static inline double __ieee754_sqrt(double x)
+{
+	double result;
+
+	__asm__ volatile ("fldl %1\n\t" /* put value */
+		"fsqrt\n\t"                 /* calc sqrt */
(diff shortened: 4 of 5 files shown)


nalajcie:
I think we need to check if the metrology on DTR project won't be affected by the change. @gerard5 can you compile `metersrv` with this PR so we could test it?

gerard5 (commented):
> I think we need to check if the metrology on DTR project won't be affected by the change. @gerard5 can you compile `metersrv` with this PR so we could test it?

Verified: on DTR HW - works ok, on imxrt1176 works.

kemonats:
@agkaminski look at these cases.

Code without set errno.
```c
#define NAN       __builtin_nanf("")

double sqrt(double x) {
    if (x < 0.0) {
        return -NAN;
    }
    return __builtin_sqrt(x);
}
```
1. armv7m4 optimization `-O2`
```00000000 <sqrt>:
   0:   b538            push    {r3, r4, r5, lr}
   2:   2200            movs    r2, #0
   4:   2300            movs    r3, #0
   6:   4604            mov     r4, r0
   8:   460d            mov     r5, r1
   a:   f7ff fffe       bl      0 <__aeabi_dcmplt>      a: R_ARM_THM_CALL       __aeabi_dcmplt
   e:   b928            cbnz    r0, 1c <sqrt+0x1c>
  10:   4620            mov     r0, r4
  12:   4629            mov     r1, r5
  14:   e8bd 4038       ldmia.w sp!, {r3, r4, r5, lr}
  18:   f7ff bffe       b.w     0 <sqrt>        18: R_ARM_THM_JUMP24    sqrt
  1c:   4901            ldr     r1, [pc, #4]    ; (24 <sqrt+0x24>)
  1e:   2000            movs    r0, #0
  20:   bd38            pop     {r3, r4, r5, pc}
  22:   bf00            nop
  24:   fff80000        .word   0xfff80000
```

2. armv7a7 optimization `-Os`
```
00000000 <sqrt>:
   0:   eeb5 0bc0       vcmpe.f64       d0, #0.0
   4:   eef1 fa10       vmrs    APSR_nzcv, fpscr
   8:   d401            bmi.n   e <sqrt+0xe>
   a:   f7ff bffe       b.w     0 <sqrt>        a: R_ARM_THM_JUMP24     sqrt
   e:   ed9f 0b02       vldr    d0, [pc, #8]    ; 18 <sqrt+0x18>
  12:   4770            bx      lr
  14:   f3af 8000       nop.w
  18:   00000000        .word   0x00000000
  1c:   fff80000        .word   0xfff80000
```

3. armv7a7 optimization `-O2`
```
00000000 <sqrt>:
   0:   eeb5 0bc0       vcmpe.f64       d0, #0.0
   4:   b508            push    {r3, lr}
   6:   ed2d 8b02       vpush   {d8}
   a:   eef1 fa10       vmrs    APSR_nzcv, fpscr
   e:   d40b            bmi.n   28 <sqrt+0x28>
  10:   eeb5 0b40       vcmp.f64        d0, #0.0
  14:   eeb1 8bc0       vsqrt.f64       d8, d0
  18:   eef1 fa10       vmrs    APSR_nzcv, fpscr
  1c:   d40b            bmi.n   36 <sqrt+0x36>
  1e:   eeb0 0b48       vmov.f64        d0, d8
  22:   ecbd 8b02       vpop    {d8}
  26:   bd08            pop     {r3, pc}
  28:   ed9f 8b05       vldr    d8, [pc, #20]   ; 40 <sqrt+0x40>
  2c:   eeb0 0b48       vmov.f64        d0, d8
  30:   ecbd 8b02       vpop    {d8}
  34:   bd08            pop     {r3, pc}
  36:   f7ff fffe       bl      0 <sqrt>        36: R_ARM_THM_CALL      sqrt
  3a:   e7f0            b.n     1e <sqrt+0x1e>
  3c:   f3af 8000       nop.w
  40:   00000000        .word   0x00000000
  44:   fff80000        .word   0xfff80000
```

nalajcie:
@kemonats thnks for staying viliglant!

We've also detected the anomaly internally and narrowed it down to `-fno-math-errno` gcc feature.

Basically - the GCC knows which math functions may change errno values and changes builtin calls to `libc` calls unless explicit `-fno-math-errno` is set. The `-O2` version works by accident (I think the call to `sqrt` is inlined and optimized - see that `-O2 -fno-math-errno` produces slightly shorter code), `-Os` doesn't inline the call which results in infinite loop.

Other math functions behave the same way, we just haven't found out until this change.

I think we should build `libphoenix` with `-fno-math-errno` as we're explicitly handling the errno by ourselfs.

agkaminski:
`__builtin_sqrt()` turned out not to be a way to go - it will cause random `sqrt` -> `sqrt` reference, e.g. when softfloat is used

math/power.c
```diff
@@ -97,10 +108,10 @@ double sqrt(double x)
 		init->i.mantisa |= (0x1ull << 51);
 	}
 
-	int i;
 	/* Reciprocal sqare root iters (avoiding division): */
-	for (i = 0; i < 4; ++i)
+	for (int i = 0; i < 4; ++i) {
```
nalajcie, line 112: :heart:

anglov (dismissed):
Unfortunately your solution doesn't work on armv7a7-imx6ull-evk

Please don't be misleaded by arm code generation - it generates vsqrt.f64 call even without your changes (generation same with and without your change):


```c
#include <math.h>

double my_sqrt(double a) {
        return sqrt(a);
}
```

```asm
00008038 <my_sqrt>:
    8038:       eeb5 0b40       vcmp.f64        d0, #0.0
    803c:       b508            push    {r3, lr}
    803e:       ed2d 8b02       vpush   {d8}
    8042:       eef1 fa10       vmrs    APSR_nzcv, fpscr
    8046:       eeb1 8bc0       vsqrt.f64       d8, d0
    804a:       d404            bmi.n   8056 <my_sqrt+0x1e>
    804c:       eeb0 0b48       vmov.f64        d0, d8
    8050:       ecbd 8b02       vpop    {d8}
    8054:       bd08            pop     {r3, pc}
    8056:       f000 f80b       bl      8070 <sqrt>
    805a:       e7f7            b.n     804c <my_sqrt+0x14>
```

BTW - it's interesting. If I correctly understand generated code, compiler do **exactly** what we want: when positive result of asm command, then use it, in other case call soft implementation to get an errno


Anyways, sqrt function after your changes is generated as
```asm
00008070 <sqrt>:
    8070:       eeb5 0bc0       vcmpe.f64       d0, #0.0
    8074:       b510            push    {r4, lr}
    8076:       b082            sub     sp, #8
    8078:       eef1 fa10       vmrs    APSR_nzcv, fpscr
    807c:       d477            bmi.n   816e <sqrt+0xfe>
    807e:       eeb5 0b40       vcmp.f64        d0, #0.0
    8082:       eef0 0b40       vmov.f64        d16, d0
    8086:       eef1 fa10       vmrs    APSR_nzcv, fpscr
    808a:       d05a            beq.n   8142 <sqrt+0xd2>
    808c:       f240 0300       movw    r3, #0
    8090:       f2c0 0300       movt    r3, #0
    8094:       2b00            cmp     r3, #0
    8096:       d150            bne.n   813a <sqrt+0xca>
    8098:       eef7 7bc0       vcvt.f32.f64    s15, d0
    809c:       f647 70f0       movw    r0, #32752      ; 0x7ff0
    80a0:       eef7 6a00       vmov.f32        s13, #112       ; 0x3f800000  1.0
    80a4:       ee86 7aa7       vdiv.f32        s14, s13, s15
    80a8:       eeb7 7ac7       vcvt.f64.f32    d7, s14
    80ac:       ed8d 7b00       vstr    d7, [sp]
    80b0:       9c01            ldr     r4, [sp, #4]
    80b2:       9a00            ldr     r2, [sp, #0]
    80b4:       f3c4 0313       ubfx    r3, r4, #0, #20
    80b8:       ee17 1a90       vmov    r1, s15
    80bc:       ea52 0403       orrs.w  r4, r2, r3
    80c0:       ea00 4111       and.w   r1, r0, r1, lsr #16
    80c4:       d101            bne.n   80ca <sqrt+0x5a>
    80c6:       4281            cmp     r1, r0
    80c8:       d03b            beq.n   8142 <sqrt+0xd2>
    80ca:       2900            cmp     r1, #0
    80cc:       d03d            beq.n   814a <sqrt+0xda>
    80ce:       f8bd 0006       ldrh.w  r0, [sp, #6]
    80d2:       9901            ldr     r1, [sp, #4]
    80d4:       0852            lsrs    r2, r2, #1
    80d6:       ea42 72c3       orr.w   r2, r2, r3, lsl #31
    80da:       f3c0 100a       ubfx    r0, r0, #4, #11
    80de:       085b            lsrs    r3, r3, #1
    80e0:       f363 0113       bfi     r1, r3, #0, #20
    80e4:       e9cd 2100       strd    r2, r1, [sp]
    80e8:       0841            lsrs    r1, r0, #1
    80ea:       07c0            lsls    r0, r0, #31
    80ec:       d436            bmi.n   815c <sqrt+0xec>
    80ee:       f8bd 0006       ldrh.w  r0, [sp, #6]
    80f2:       f201 11ff       addw    r1, r1, #511    ; 0x1ff
    80f6:       9200            str     r2, [sp, #0]
    80f8:       f443 2300       orr.w   r3, r3, #524288 ; 0x80000
    80fc:       f361 100e       bfi     r0, r1, #4, #11
    8100:       f8ad 0006       strh.w  r0, [sp, #6]
    8104:       9901            ldr     r1, [sp, #4]
    8106:       f363 0113       bfi     r1, r3, #0, #20
    810a:       9101            str     r1, [sp, #4]
    810c:       eef6 2b00       vmov.f64        d18, #96        ; 0x3f000000  0.5
    8110:       2304            movs    r3, #4
    8112:       ed9d 0b00       vldr    d0, [sp]
    8116:       eef7 3b08       vmov.f64        d19, #120       ; 0x3fc00000  1.5
    811a:       ee60 2ba2       vmul.f64        d18, d16, d18
    811e:       eef0 4b63       vmov.f64        d20, d19
    8122:       3b01            subs    r3, #1
    8124:       ee60 1b62       vnmul.f64       d17, d0, d18
    8128:       eee1 4b80       vfma.f64        d20, d17, d0
    812c:       ee20 0b24       vmul.f64        d0, d0, d20
    8130:       d1f5            bne.n   811e <sqrt+0xae>
    8132:       ee20 0b20       vmul.f64        d0, d0, d16
    8136:       b002            add     sp, #8
    8138:       bd10            pop     {r4, pc}
    813a:       f3af 8000       nop.w
    813e:       b002            add     sp, #8
    8140:       bd10            pop     {r4, pc}
    8142:       eeb0 0b60       vmov.f64        d0, d16
    8146:       b002            add     sp, #8
    8148:       bd10            pop     {r4, pc}
    814a:       9801            ldr     r0, [sp, #4]
    814c:       0852            lsrs    r2, r2, #1
    814e:       ea42 72c3       orr.w   r2, r2, r3, lsl #31
    8152:       085b            lsrs    r3, r3, #1
    8154:       f363 0013       bfi     r0, r3, #0, #20
    8158:       e9cd 2000       strd    r2, r0, [sp]
    815c:       f8bd 3006       ldrh.w  r3, [sp, #6]
    8160:       f501 7100       add.w   r1, r1, #512    ; 0x200
    8164:       f361 130e       bfi     r3, r1, #4, #11
    8168:       f8ad 3006       strh.w  r3, [sp, #6]
    816c:       e7ce            b.n     810c <sqrt+0x9c>
    816e:       f003 fbf1       bl      b954 <__errno_location>
    8172:       2321            movs    r3, #33 ; 0x21
    8174:       ed9f 0b02       vldr    d0, [pc, #8]    ; 8180 <sqrt+0x110>
    8178:       6003            str     r3, [r0, #0]
    817a:       b002            add     sp, #8
    817c:       bd10            pop     {r4, pc}
    817e:       bf00            nop
    8180:       00000000        .word   0x00000000
    8184:       fff80000        .word   0xfff80000
```
From above snippet,  call to __ieee754_sqrt is generated as 
```asm
(...)
    808c:       f240 0300       movw    r3, #0
    8090:       f2c0 0300       movt    r3, #0
    8094:       2b00            cmp     r3, #0
(...)
```

So, always null and no __ieee754_sqrt in example application code.

When I removed weak symbol for verification, sqrt is now good:
```asm
00008070 <sqrt>:
    8070:       eeb5 0bc0       vcmpe.f64       d0, #0.0
    8074:       eef1 fa10       vmrs    APSR_nzcv, fpscr
    8078:       d407            bmi.n   808a <sqrt+0x1a>
    807a:       eeb5 0b40       vcmp.f64        d0, #0.0
    807e:       eef1 fa10       vmrs    APSR_nzcv, fpscr
    8082:       d100            bne.n   8086 <sqrt+0x16>
    8084:       4770            bx      lr
    8086:       f003 bbf5       b.w     b874 <__ieee754_sqrt>
    808a:       b508            push    {r3, lr}
    808c:       f003 fbf6       bl      b87c <__errno_location>
    8090:       eddf 0b03       vldr    d16, [pc, #12]  ; 80a0 <sqrt+0x30>
    8094:       2321            movs    r3, #33 ; 0x21
    8096:       eeb0 0b60       vmov.f64        d0, d16
    809a:       6003            str     r3, [r0, #0]
    809c:       bd08            pop     {r3, pc}
    809e:       bf00            nop
    80a0:       00000000        .word   0x00000000
    80a4:       fff80000        .word   0xfff80000
```
I suspect  that strong symbol was removed due to --gc-sections flag or by discarding whole .o file from .a archive (need of __ieee754_sqrt symbol was satisfied by weak symbol and it was the only symbol needed from cpu.o file)

Weak symbols can be really troublesome when used without enough care. Result of compilation can depend on linking order.

I don't have good solution for now. I know that current one not working as expected.

Full reproduction code below (modified hello app in _user dir)

==> _user/hello/main.c <==
```c
/*
 * Phoenix-RTOS
 *
 * Hello World
 *
 * Example of user application
 *
 * Copyright 2021 Phoenix Systems
 * Author: Hubert Buczynski
 *
 * This file is part of Phoenix-RTOS.
 *
 * %LICENSE%
 */

#include <stdio.h>


extern double my_sqrt(double);

int main(void)
{
        double d = my_sqrt(2);
        printf("Hello World of sqrt2: %lf!!\n", d);

        return 0;
}
```

==> _user/hello/Makefile <==
```makefile
#
# Makefile for user application
#
# Copyright 2021 Phoenix Systems
#

LD_FLAGS += -verbose
NAME := hello
LOCAL_SRCS := sqrt.c main.c

include $(binary.mk)
```
==> _user/hello/sqrt.c <==
```c
#include <math.h>

double my_sqrt(double a) {
        return sqrt(a);
}
```

agkaminski:
@anglov Yeah, noticed that, will change tommorow to #if, weak symbols are unreliable, I don't trust it anymore, even if I get it to work.

nalajcie:
Yep, we were researching it today - I think weak symbols are intendent to be used across libraries not within single one.

Musl and newlib does it in other ways:
 - they both implement single external API function by single `.c` file
 - the same file name is being used for soft generic implementation and for machine-specific one (if exists)
 - the resulting `.o` file is either overwritten based on order of compilation (newlib: https://github.com/bminor/newlib/blob/master/newlib/libm/Makefile.inc#L38) or explicitly replaced in makefile (musl: https://github.com/bminor/musl/blob/master/Makefile#L28)

As we're linking with `--gc-sections`, we don't need to place each function in separate file - but maybe we should do it (and don't expect user to always use `--gc-sections` while linking)?

Side note: glibc used to provide specific per-arch `sqrt` but 3 days ago switched to builtin (lol) - https://github.com/bminor/glibc/commit/cbf3571f49134388bc589b74445b5d6bdd21ef02

Maybe using per-arch defines for builtin (and `-fno-math-errno` - glibc uses it for libm compilation - https://github.com/bminor/glibc/blob/master/Makeconfig#L879) + soft implementation would be enough?

agkaminski:
@gerard5 I'm more inclined towards a roboust solution, this seems like "works by miracle" case to me

nalajcie:
Yeah, this works because `Ofast` implies `ffast-math` which enables `fno-math-errno`. 

Not sure if softfpu version would be correct as GCC doesn't provide soft sqrt and calls libc (so might result in an infinite loop or no-op).

kemonats:
What do you think about such a solution:
```c
#if defined(__arm__)
#if (__ARM_PCS_VFP || (__VFP_FP__ && !__SOFTFP__)) && (__ARM_FP & 8)
#define SQRT_HW 1
static inline double sqrt_hw(double x)
{
	__asm__ volatile ("vsqrt.f64 %P0, %P1" : "=w"(x) : "w"(x));
	return x;
}
#endif
#endif /* __arm__ */

#if defined(__i386__)
/* sqrt_hw() definition for i386 */
#endif

#if defined(__riscv__)
/* sqrt_hw() definition for riscv */
#endif

double sqrt(double x)
{
	if (x < 0.0) {
		errno = EDOM;
		return -NAN;
	}

#if defined(SQRT_HW)
	return sqrt_hw(x);
#else
/* rest of code */
#endif
```
`sqrt_hw()` function definition can be moved to the `arch.h` file.

gerard5:
I've checked ia32 speed on i686, and you may add i386 implementation to `include/arch/ia32/arch.h`:
```c
#if (!__SOFTFP__)
#define __IEEE754_SQRT

static inline double __ieee754_sqrt(double x)
{
	double result;

	asm("fldl %1\n"
		"fsqrt\n"
		"fstpl %0"
		: "=m"(result)
		: "m"(x));

	return result;
}

#endif
```
since we don't support SSE yet, this code seems to run faster on i686 than the software implementation (I don't count old i386 with i387).

gerard5:
Ok checked with border cases, updated version which corrects the mantisa of 80bit/64bit, also I've described what it do to be more clear. Still faster than software implementation.
```asm
static inline double __ieee754_sqrt(double x)
{
	double result;

	asm("fldl %1\n\t"        /* put value */
		"fsqrt\n\t"      /* calc sqrt */
		"fxtract\n\t"    /* extract exponent */
		"fisttp %0\n\t"  /* round the exponent */
		"fld %0\n\t"     /* load the rounded exponent */
		"fcomp\n\t"      /* compare with the original exponent */
		"fstsw %%ax\n\t" /* save FPU status to AX register */
		"sahf\n\t"       /* copy status to CPU flags */
		"jp 1f\n\t"      /* jump if the result was exact */
		"fsubrp\n"       /* adjust mantissa down */
		"1:\n\t"
		"fstpl %0"       /* get the result */
		: "=m"(result)
		: "m"(x));

	return result;
}
```

lukileczo:
On SPARC there are instructions `fsqrt(s/d/q)`, more info here: https://www.gaisler.com/doc/sparcv8.pdf#page=143
At the moment we're not using FPU, so the GCC defines `_SOFT_FLOAT 1`

kemonats:
Maybe add such a line at the beginning of the file for information.
Now `arch.h` is included in the file `limits.h`
```c
#include <arch.h> /* for __ieee754_sqrt */
```

agkaminski:
@kemonats normally I would be reluctant, but in this case it is essential to include this header, as lack of it would cause silent fallback to soft sqrt implementation. I also added explicit `arch.h` include, thanks

gerard5 (approved):
Looks good to me. If there are no objections, let's merge it.
btw. as for the mentioned issue that [for small small values sqrt returns wrong values](https://github.com/phoenix-rtos/phoenix-rtos-project/issues/766), the fix can be done in a dedicated pull-request.

PR #265, opened 2023-06-20
refactor path resolving

xvuko:
Main point of this is to remove `char tmp_buf[PATH_MAX];` from stack. I refactored rest of the function to make it (hopefully) easier to read. 

This should give same output as before. Only change is case of empty symlink as it was setting errno to 0.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/unistd/dir.c
+++ b/unistd/dir.c
@@ -154,111 +154,118 @@ char *canonicalize_file_name(const char *path)
  * path might be rewritten as a part of the path resolving */
 static int _resolve_abspath(char *path, char *result, int resolve_last_symlink, int allow_missing_leaf)
 {
-	char *p = path + 1; /* skip first '/' */
-	char *r = result;
-	char *slash;
-	char *const res_end = result + PATH_MAX;
-	int is_leaf = 0;
+	char *r = result; /* r -  first unused result byte */
+	char *p = path;   /* p - first byte of remaining unresolved path */
 	int symlinkcnt = 0;
 
 	assert(path && path[0] == '/');
 
-	*r++ = '/';
+	/* path is shifted right to use its's start for symlink resolution */
+	const size_t path_len = strlen(path) + 1; /* including '\0' */
+	if (path_len < PATH_MAX) {
+		p = path + PATH_MAX - path_len;
+		memmove(p, path, path_len);
+	}
+	p += 1; /* skip initial '/' */
 
-	while (!is_leaf) {
-		slash = strchrnul(p, '/');
-		if (*slash == 0) {
-			is_leaf = 1; /* last path node */
-		}
-		else {
-			char *rem = slash;
-			while (*rem == '/')
-				rem++;
-			if (*rem == 0)
-				is_leaf = 1; /* trailing '/' */
+	while (*p != '\0') {
+		char *const node = p;
+		p = strchrnul(node, '/');
+		const size_t node_len = p - node;
 
-			*slash = 0;
+		/* skip trailing slashes to properly detect leaf node */
+		while (*p == '/') {
+			p += 1;
 		}
+
+		const int is_leaf = (*p == '\0') ? 1 : 0;
 
 		/* check for '.' and '..' */
-		if (p[0] == 0) { /* multiple '/' */
-			r -= 1;
+		if (node_len == 0) { /* multiple '/' */
+			continue;
 		}
-		else if (p[0] == '.' && p[1] == 0) { /* '.' dir */
-			r -= 1;
+		if ((node_len == 1) && (node[0] == '.')) { /* '.' dir */
+			continue;
 		}
-		else if (p[0] == '.' && p[1] == '.' && p[2] == 0) { /* '..' dir */
-			r -= 1;
+		if ((node_len == 2) && (node[0] == '.') && (node[1] == '.')) { /* '..' dir */
 			while (r > result && *--r != '/')
 				;
+			continue;
 		}
-		else { /* 'normal' path node */
-			char tmp_buf[PATH_MAX];
-			ssize_t symlink_len;
 
-			/* check for buffer overflows */
-			if (r + (slash - p + 1) > res_end)
-				return SET_ERRNO(-ENAMETOOLONG);
+		if ((r + 1 + node_len + 1) > (result + PATH_MAX)) {
+			return SET_ERRNO(-ENAMETOOLONG);
+		}
 
-			memcpy(r, p, (slash - p + 1));
+		*r++ = '/';
+		memcpy(r, node, node_len);
+		r += node_len;
+		*r = '\0';
 
-			if (is_leaf && !resolve_last_symlink) {
-				r += slash - p; /* commit node path */
+		if ((is_leaf != 0) && (resolve_last_symlink == 0)) {
+			/* WARN: slight inconsistency: we're not checking if the path actually exists in this case (TODO?) */
+			break;
+		}
 
-				/* WARN: slight inconsistency: we're not checking if the path actually exists in this case (TODO?) */
-			}
-			/* (hackish) save some messsaging by not calling lstat, but directly readlink() and checking error code */
-			else if ((symlink_len = _readlink_abs(result, tmp_buf, sizeof(tmp_buf) - 1)) < 0) { /* reserve space for \0 */
+		/*
+		 * Number of initial unused `path` bytes that can be used to resolve current node. If link content is
+		 * longer we will not be able to merge it with remaining path and still fit in PATH_MAX buffer
+		 */
+		const size_t readlink_max_len = p - path;
 
-				if ((errno == EINVAL) || ((errno == ENOENT) && is_leaf && allow_missing_leaf)) { /* not a link or non-esixting leaf */
-					r += slash - p;
-				}
-				else {
-					/* errno set by _readlink_abs */
-					return -1;
-				}
-			}
-			else {
-				if (++symlinkcnt > SYMLOOP_MAX)
-					return SET_ERRNO(-ELOOP);
+		/* (hackish) save some messsaging by not calling lstat, but directly readlink() and checking error code */
+		ssize_t symlink_len = _readlink_abs(result, path, readlink_max_len);
 
-				if (!is_leaf) { /* append remaining original path suffix */
-					*slash = '/';
-					int rempath_len = strlen(slash);
-					if (symlink_len + rempath_len + 1 >= PATH_MAX) {
-						return SET_ERRNO(-ENAMETOOLONG);
-					}
-					memcpy(tmp_buf + symlink_len, slash, rempath_len + 1);
-					symlink_len += rempath_len;
-				}
-
-				/* reuse full path buffer */
-				memcpy(path, tmp_buf, symlink_len);
-				path[symlink_len] = 0;
-				p = path;
-
-				if (tmp_buf[0] == '/') { /* ABS symlink - reset result to '/' */
-					r = result + 1;
-				}
-				/* else: relative symlink - keep current result */
-
-				/* keep parsing resolved symlink */
-				is_leaf = 0;
+		if (symlink_len < 0) {
+			if (errno == EINVAL) { /* not a symlink */
 				continue;
 			}
+			else if ((errno == ENOENT) && (is_leaf != 0) && (allow_missing_leaf != 0)) { /* non-esixting leaf */
+				break;
+			}
+			else {
+				/* errno set by _readlink_abs */
+				return -1;
+			}
 		}
 
-		if (!is_leaf) {
-			*r++ = '/';
-			p = slash + 1;
+		if (symlink_len == 0) {
+			/* NOTE: this case is not defined in POSIX */
+			/* see: https://lwn.net/Articles/551224/ */
+			return SET_ERRNO(-ENOENT);
+		}
+
+		symlinkcnt += 1;
+		if (symlinkcnt > SYMLOOP_MAX) {
+			return SET_ERRNO(-ELOOP);
+		}
+
+		assert(symlink_len <= readlink_max_len);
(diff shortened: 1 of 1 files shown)


niewim19 (commented):
Hey, I have done some research on this change and I found some issues within this change and in current implementation. @damianloew I have described current issue in section 2 of this comment. I will also make an issue on this.

# Section 1, issue in proposed change

Using `readlink` instead of `stat` looks neat, considering how many messages are sent by `stat` (~10). Proposed change uses spare place in `path` argument to expand and process symlinks content. This may be erroneous if original path is sufficiently long so that not all data read from symlink can be stored. Such clipped symlink (`readlink` does not fail if place to write is too small) content can be data-wise valid and resolve to _some place_ in filesystem. It would be a byzantine error, and very hard to find.

Test code below works on host and current implementation but causes segmentation fault on proposed change.

# Section 2, error in current implementation

Makeshift code that I have produced to check Section 1 error has found a discrepancy between current implementation of `resolve_path()` (via `realpath()`) and the host one.

Test code below resolves to `/etc` on host, but `/symShort` on phoenix, which is in violation with goal of `realpath()` which should remove symlinks.

# Test code

Test code tries to resolve very long path that starts with symlink. It instantiates:
 - symlink `symNameLong` with long name that points to "/"
 - symlink `symNameShort` that points to `symNameLong`
 - `basePath` path that goes like: `"symShort/etc/../etc/<and so on up to almost PATH_MAX length>/../etc"` that should finally land in `etc`.

```c
#include <stdio.h>
#include <unistd.h>
#include <sys/stat.h>
#include <stdlib.h>
#include <limits.h>
#include <string.h>
#include <errno.h>

#define NAME_LEN (NAME_MAX / 2)

int main(int argc, char **argv) {
	char symNameShort[] = "symShort";
	char symNameLong[NAME_LEN];
	char basePath[] = "symShort/etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/../etc/";
	char resolved[PATH_MAX + 1];

	memset(symNameLong, 'a', NAME_LEN);
	symNameLong[NAME_LEN - 1] = 0;

	if (symlink("/", symNameLong) < 0)
		printf("%s\n", strerror(errno));

	if (symlink(symNameLong, symNameShort) < 0)
		printf("%s\n", strerror(errno));

	realpath(basePath, resolved);
	resolved[PATH_MAX] = 0;
	printf("%s\n", resolved);

	/* cleanup */
	remove(symNameShort);
	remove(symNameLong);
}
```

`basePath` is resolved to:
 - `/etc` on host
 - `/symShort` on phoenix (tested on zturn)
 - `segmentation fault` on proposed change

nalajcie:
small notes from me:
 - the code looks sane (but I haven't tested it), I was thinking about similar approach at the initial approach but decided the increased code complexity wasn't worth the case
 - please write in commit message body regarding WHY this change is needed (or what we gained) to leave trail
 - @niewim19 the `_readlink_abs` instead of `lstat` was in the original approach so this is not a functionality change
 - if there is a test that highlights possible error in current / new implementation - please try to add it to unit tests :)
 - I'll try to test it in the common use cases on DTR project, will follow up then

unistd/dir.c
```diff
@@ -154,111 +154,109 @@ char *canonicalize_file_name(const char *path)
...
-			is_leaf = 1; /* last path node */
-		}
-		else {
-			char *rem = slash;
-			while (*rem == '/')
-				rem++;
-			if (*rem == 0)
-				is_leaf = 1; /* trailing '/' */
+	while (*p != '\0') {
+		char *node = p;
```
nalajcie, line 172: `const`?

unistd/dir.c
```diff
@@ -154,111 +154,109 @@ char *canonicalize_file_name(const char *path)
...
+		p = strchrnul(node, '/');
+		size_t node_len = p - node;
```
nalajcie, line 174: `const`?

niewim19:
>     * @niewim19 the `_readlink_abs` instead of `lstat` was in the original approach so this is not a functionality change
> 
>     * if there is a test that highlights possible error in current / new implementation - please try to add it to unit tests :)
> 
>     * I'll try to test it in the common use cases on DTR project, will follow up then

- I`m sorry for that - it is indeed not a functionality change. I have misunderstood diff. I will amend previous message
- I have spoke with test team (@damianloew) about this issue, they will take it from here.

damianloew:
> > small notes from me:
> > ```
> > * the code looks sane (but I haven't tested it), I was thinking about similar approach at the initial approach but decided the increased code complexity wasn't worth the case
> > 
> > * please write in commit message body regarding WHY this change is needed (or what we gained) to leave trail
> > 
> > * @niewim19 the `_readlink_abs` instead of `lstat` was in the original approach so this is not a functionality change
> > 
> > * if there is a test that highlights possible error in current / new implementation - please try to add it to unit tests :)
> > 
> > * I'll try to test it in the common use cases on DTR project, will follow up then
> > ```
> 
> * I`m sorry for that - it is indeed not a functionality change. I have misunderstood diff. I will amend previous message
> * I have spoke with test team (@damianloew) about this issue, they will take it from here.

Yeah, the issue will be created and we will add tests to cover this issue.

agkaminski:
@xvuko Hmmm, I don't know... Doesn't this solution cause us to be unable to on a long path not being able to solve even longer symlink? Total path length being shorter than PATH_MAX in both cases, but us being unable to solve it anyway

nalajcie:
> @xvuko Hmmm, I don't know... Doesn't this solution cause us to be unable to on a long path not being able to solve even longer symlink? Total path length being shorter than PATH_MAX in both cases, but us being unable to solve it anyway

I think the original solution would also fail if symlink resolved length + remaining (unresolved) path len exceeds PATH_MAX (while resolving, even if the expected resolve result would be < PATH_MAX).

damianloew:
> > @xvuko Hmmm, I don't know... Doesn't this solution cause us to be unable to on a long path not being able to solve even longer symlink? Total path length being shorter than PATH_MAX in both cases, but us being unable to solve it anyway
> 
> I think the original solution would also fail if symlink resolved length + remaining (unresolved) path len exceeds PATH_MAX (while resolving, even if the expected resolve result would be < PATH_MAX).

We checked it and in fact there is no issue in the current libphoenix implementation as you mentioned @nalajcie So I've updated the issue and closed it. However the exception shouldn't be raised when applying changes from the PR when running code from the mentioned case.

xvuko:
>     * `segmentation fault` on proposed change

Thanks for testing this. There was a bug in my code:
```c
assert(symlink_len <= path - p);                                                                                                                                                                                                                                                                                                                                          
if (symlink_len == path - p) {                                                                                                                                                                                                                                                                                                                                            
      return SET_ERRNO(-ENAMETOOLONG);                                                                                                                                                                                                                                                                                                                                  
}
```
I changed `path - p` to `p - path` and now `/symShort` is printed.

nalajcie:
- ok, so maybe adding the mentioned test would still be beneficial?
- we could introduce `symlink_max_len = p - path`  but I'm not sure if it would increase readability or not
- (Just analyzing code) I see that `/` is being added unconditionally (also if leaf) so we might end up in with eg. `/dir/realfile/` in `path` - but from reading the code it seems it would correctly be handled? (this reduces max resolved symlink len by one from the previous implementation but IMHO doesn't really matter)

xvuko:
> * ok, so maybe adding the mentioned test would still be beneficial?
Yes, I will add this test.

>     * we could introduce `symlink_max_len = p - path`  but I'm not sure if it would increase readability or not
Yeah, I was thinking about that. Added as it is a good place for comment.

>     * (Just analyzing code) I see that `/` is being added unconditionally (also if leaf) so we might end up in with eg. `/dir/realfile/` in `path` - but from reading the code it seems it would correctly be handled? (this reduces max resolved symlink len by one from the previous implementation but IMHO doesn't really matter)

'/' at the and is not an issue as trailing slashes are always ignored.

unistd/dir.c
```diff
@@ -154,111 +154,115 @@ char *canonicalize_file_name(const char *path)
...
-				return SET_ERRNO(-ENAMETOOLONG);
+		if (r + 1 + node_len + 1 > result + PATH_MAX)
+			return SET_ERRNO(-ENAMETOOLONG);
+
+		*r++ = '/';
+		memcpy(r, node, node_len);
+		r += node_len;
+		*r = '\0';
+
+		if (is_leaf && !resolve_last_symlink) {
```
agkaminski, line 204:
```suggestion
		if ((is_leaf != 0) && (resolve_last_symlink == 0)) {
```

unistd/dir.c
```diff
@@ -154,111 +154,115 @@ char *canonicalize_file_name(const char *path)
...
 		}
-		else { /* 'normal' path node */
-			char tmp_buf[PATH_MAX];
-			ssize_t symlink_len;
 
-			/* check for buffer overflows */
-			if (r + (slash - p + 1) > res_end)
-				return SET_ERRNO(-ENAMETOOLONG);
+		if (r + 1 + node_len + 1 > result + PATH_MAX)
+			return SET_ERRNO(-ENAMETOOLONG);
```
agkaminski, line 197:
```suggestion
		if ((r + 1 + node_len + 1) > (result + PATH_MAX)) {
			return SET_ERRNO(-ENAMETOOLONG);
		}
```

unistd/dir.c
```diff
@@ -154,111 +154,115 @@ char *canonicalize_file_name(const char *path)
...
+			continue;
 		}
-		else if (p[0] == '.' && p[1] == 0) { /* '.' dir */
-			r -= 1;
+		if (node_len == 1 && node[0] == '.') { /* '.' dir */
+			continue;
 		}
-		else if (p[0] == '.' && p[1] == '.' && p[2] == 0) { /* '..' dir */
-			r -= 1;
+		if (node_len == 2 && node[0] == '.' && node[1] == '.') { /* '..' dir */
```
agkaminski, line 190:
```suggestion
		if ((node_len == 2) && (node[0] == '.') && (node[1] == '.')) { /* '..' dir */
```

unistd/dir.c
```diff
@@ -154,111 +154,115 @@ char *canonicalize_file_name(const char *path)
...
-			else if ((symlink_len = _readlink_abs(result, tmp_buf, sizeof(tmp_buf) - 1)) < 0) { /* reserve space for \0 */
-
-				if ((errno == EINVAL) || ((errno == ENOENT) && is_leaf && allow_missing_leaf)) { /* not a link or non-esixting leaf */
-					r += slash - p;
-				}
-				else {
-					/* errno set by _readlink_abs */
-					return -1;
-				}
+			else if ((errno == ENOENT) && is_leaf && allow_missing_leaf) { /* non-esixting leaf */
```
agkaminski, line 222:
```suggestion
			else if ((errno == ENOENT) && (is_leaf != 0) && (allow_missing_leaf != 0)) { /* non-esixting leaf */
```

unistd/dir.c
```diff
@@ -154,111 +154,115 @@ char *canonicalize_file_name(const char *path)
...
-			*r++ = '/';
-			p = slash + 1;
+		if (symlink_len == 0) {
+			/* NOTE: this case is not defined in POSIX */
+			/* see: https://lwn.net/Articles/551224/ */
+			return SET_ERRNO(-ENOENT);
+		}
+
+		if (++symlinkcnt > SYMLOOP_MAX)
+			return SET_ERRNO(-ELOOP);
```
agkaminski, line 238:
```suggestion
		++symlinkcnt;
		if (symlinkcnt > SYMLOOP_MAX) {
			return SET_ERRNO(-ELOOP);
		}
```

agkaminski (commented):
Some MISRA-flavored nitpicks, otherwise LGTM

damianloew:
> > * ok, so maybe adding the mentioned test would still be beneficial?
> 
> Yes, I will add this test.
> 
> > ```
> > * we could introduce `symlink_max_len = p - path`  but I'm not sure if it would increase readability or not
> > ```
> 
> Yeah, I was thinking about that. Added as it is a good place for comment.
> 
> > ```
> > * (Just analyzing code) I see that `/` is being added unconditionally (also if leaf) so we might end up in with eg. `/dir/realfile/` in `path` - but from reading the code it seems it would correctly be handled? (this reduces max resolved symlink len by one from the previous implementation but IMHO doesn't really matter)
> > ```
> 
> '/' at the and is not an issue as trailing slashes are always ignored.

The test is already added and it's now running during CI.

agkaminski (approved):
LGTM

PR #272, opened 2023-07-21
complex: add commonly used complex number funcs

gerard5:
Provide required `complex.h` function prototypes and commonly used functions.

`cabs()` and `cabsf()` implementations, to calculate distance, temporarily uses `sqrt()` instead of `hypot()` which is required but missing now.

In `cexp()` implementation Euler equation is used.

In reference to https://github.com/phoenix-rtos/phoenix-rtos-project/issues/785
JIRA: RTOS-528

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `imxrt1176-nil`, `ia32-generic`.

## Checklist:

## Special treatment

### DIFF
--- a/include/complex.h
+++ b/include/complex.h
@@ -1,0 +1,255 @@
+/*
+ * Phoenix-RTOS
+ *
+ * libphoenix
+ *
+ * mathematical functions on complex numbers
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Gerard Swiderski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#ifndef _COMPLEX_H_
+#define _COMPLEX_H_
+
+/* define complex numbers specifers */
+#define complex    _Complex
+#define _Complex_I 1.0fi
+#define I          _Complex_I
+
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+static inline double creal(double complex z)
+{
+	return (__real__(z));
+}
+
+
+static inline float crealf(float complex z)
+{
+	return (__real__(z));
+}
+
+
+static inline double cimag(double complex z)
+{
+	return (__imag__(z));
+}
+
+
+static inline float cimagf(float complex z)
+{
+	return (__imag__(z));
+}
+
+
+static inline double complex conj(double complex z)
+{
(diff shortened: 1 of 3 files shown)


include/complex.h
```diff
@@ -0,0 +1,271 @@
...
+
+static inline float complex conjf(float complex z)
+{
+	union {
+		float complex z;
+		struct {
+			float re;
+			float im;
+		} part;
+	} u = { .z = z };
```
agkaminski, line 76: Uhh, I don't really like this union, specifically assuming _Complex implementation, on which we have no control.

include/complex.h
```diff
@@ -0,0 +1,271 @@
...
+{
+	union {
+		double complex z;
+		struct {
+			double re;
+			double im;
+		} part;
+	} u = { .z = z };
+	u.part.im = -u.part.im;
+	return u.z;
```
agkaminski, line 64: I'd rather avoid doing that stuff if not necessary

btw, I guess it's ok to do:
`# define CMPLX(x, y) __builtin_complex ((double) (x), (double) (y))`

```suggestion
	return CMPLX(creal(z), -cimag(z));
```

gerard5: What about this?
```C
float complex conjf2(float complex z)
{
    __imag__(z) = -(__imag__(z));
    return z;
}
```

gerard5: It looks like when `-O0` with `__builtin_complex()` is used the machine code is shorter (one-two instructions).

math/complex.c
```diff
@@ -0,0 +1,61 @@
...
+ * %LICENSE%
+ */
+
+#include <math.h>
+#include <complex.h>
+
+
+double cabs(double complex z)
+{
+	/* FIXME: due to missing hypot() implementation, temporarily sqrt() is used */
```
agkaminski, line 22: Fun fact - `hypot()` can be implemented via `cabs()` - which should be fundamental then? :)

I guess `hypot()` should be, to not require complex numbers if not needed

agkaminski (approved):
LGTM

math/complex.c
gerard5: More than that, however, the `hypot()` function calculate $sqrt(x^2+y^2)$ in such a way that underflow will not occur, and overflow will only occur if the end result deserves it.

PR #277, opened 2023-08-08
ctype: Provide fast macro implementation

agkaminski:
DONE: RTOS-549

## Description

Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/782

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/include/ctype.h
+++ b/include/ctype.h
@@ -5,8 +5,8 @@
  *
  * ctypes.h
  *
- * Copyright 2017 Phoenix Systems
- * Author: Pawel Pisarczyk
+ * Copyright 2017, 2023 Phoenix Systems
+ * Author: Pawel Pisarczyk, Adrian Kepka, Aleksander Kaminski
  *
  * This file is part of Phoenix-RTOS.
  *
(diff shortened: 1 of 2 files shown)


gerard5 (approved):
Quick check, works for me.
2023-08-08-160615_1137x467_scrot

include/ctype.h
```diff
@@ -22,65 +22,107 @@ extern "C" {
...
+{
+	return (c >= 'A' && c <= 'Z') ? 1 : 0;
+}
 
 
 /* This function checks whether the passed character is alphabetic. */
-extern int isalpha(int c);
+static inline int isalpha(int c)
+{
+	return (islower(c) || isupper(c)) ? 1 : 0;
```
xvuko, line 42:
```suggestion
	return ((islower(c) != 0) || (isupper(c) != 0)) ? 1 : 0;
```

include/ctype.h
```diff
@@ -22,65 +22,107 @@ extern "C" {
...
+	return (c >= '0' && c <= '9') ? 1 : 0;
+}
 
 
-/* This function checks whether the passed character has graphical representation using locale. */
-extern int isgraph(int c);
+/* This function checks whether the passed character is alphanumeric. */
+static inline int isalnum(int c)
+{
+	return (isalpha(c) || isdigit(c)) ? 1 : 0;
```
xvuko, line 63:
```suggestion
	return ((isalpha(c) != 0) || (isdigit(c) != 0)) ? 1 : 0;
```

include/ctype.h
```diff
@@ -22,66 +22,93 @@ extern "C" {
 #endif
 
 
-/* This function checks whether the passed character is alphanumeric. */
-extern int isalnum(int c);
+/* This function checks whether the passed character is lowercase letter. */
+int islower(int c);
+#define _islower(c) (((c) >= 'a' && (c) <= 'z') ? 1 : 0)
```
anglov, line 27: Small C nitpick.
C99 7.1.3 p 1 states:
— All identifiers that begin with an underscore and either an uppercase letter or another underscore are always reserved for any use.
— All identifiers that begin with an underscore are always reserved for use as identifiers with file scope in both the ordinary and tag name spaces.

C99 6.2.3 states that macros is not considered as identifiers in terms of scope and namespaces.

If my understanding of those is correct, then macros in headers should begin with __ (double _).

agkaminski: Done

anglov (approved):
Ok for me in terms of C99/POSIX purism. It would be great if you could rerun test. 

Looking at code generation (root @xvuko  cause)  would also be advisible.

include/ctype.h
```diff
@@ -22,66 +22,93 @@ extern "C" {
...
+#define toupper(c)   __toupper(c)
 
 
-static inline int isascii(int c)
-{
-	return (c & ~0x7f);
-}
+/* This function test if character is representable as ASCII value. */
+int isascii(int c);
+#define __isascii(c) ((((c) < 0) || ((c) > '\x7f')) ? 0 : 1)
```
gerard5, line 110: Why is `'\x7f'` instead of `0x7f` needed, although for iscntrl case `0x7f` is used not `'\x7f'` literal ?
```suggestion
#define __isascii(c) ((((c) < 0) || ((c) > 0x7f)) ? 0 : 1)
```

anglov (approved):

PR #284, opened 2023-08-23
strerror, strerror_r, gai_strerror, perror: Add full implementation

### LINKED ISSUE #181: libphoenix: perror implementation
For example in busybox test suite expected output for 'busybox bunzip2 nonexistentFile' is 'bunzip2: nonexistentFile: No such file or directory' instead of 'bunzip2: nonexistentFile: ENOENT'

agkaminski:
DONE: RTOS-569

## Description

Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/181

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic-qemu).

## Checklist:

## Special treatment

### DIFF
--- a/stdio/perror.c
+++ b/stdio/perror.c
@@ -5,7 +5,7 @@
  *
  * perror.c
  *
- * Copyright 2022 Phoenix Systems
+ * Copyright 2022, 2023 Phoenix Systems
  * Author: Aleksander Kaminski
  *
  * This file is part of Phoenix-RTOS.
@@ -17,7 +17,16 @@
 #include <errno.h>
 #include <string.h>
 
+
 void perror(const char *str)
 {
-	fprintf(stderr, "%s: %s\n", str, strerror(errno));
+	char errstr[64];
+
+	(void)strerror_r(errno, errstr, sizeof(errstr));
+
+	if (str != NULL) {
+		fprintf(stderr, "%s: ", str);
+	}
+
+	fprintf(stderr, "%s\n", errstr);
 }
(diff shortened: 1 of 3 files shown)


nalajcie:
IMHO this should be implementation of `strerror`, `perror` might keep using `strerror` :)

gerard5:
From my point of view, two implementations depending on the available platform resources (mem) should be considered. **For example, after this change, each program that uses `perror()` e.g. `psh` increases by ca. 2KiB (imxrt, stm32)**. The earlier solution `strerror()` already implements a table for shorter error codes, and this is where the functionality should be extended, so that `strerror()` returns either a short description (for small targets, nommu) or a long error description for (zynq, ia32, 6ull), .. or depending on build config.

agkaminski:
@gerard5 I've removed previous `strerror` string table which should somewhat reduce memory usage

stdio/perror.c
```diff
@@ -17,7 +17,12 @@
...
 
+
 void perror(const char *str)
 {
-	fprintf(stderr, "%s: %s\n", str, strerror(errno));
+	char errstr[64];
+
+	(void)strerror_r(errno, errstr, sizeof(errstr));
+
+	fprintf(stderr, "%s: %s\n", str, errstr);
```
kemonats, line 27: I suggest adding this:
```suggestion
	if ((str != NULL) && (*str != '\0')) {
		fprintf(stderr, "%s: ", str);
	}
	fprintf(stderr, "%s\n", errstr);
```

agkaminski: I've applied only NULL check, IMHO no need to check for empty string, it doesn't hurt anything

Issue #175, opened 2022-04-11
arch/armv7a/jmp.S: unaligned stack


kemonats:
I have several issuess:
1. In the `setjmp()` function stack is unaligned
https://github.com/phoenix-rtos/libphoenix/blob/841a31eff278626c02aa709a828ba7c86804119e/arch/armv7a/jmp.S#L76-L84
2. Shouldn't the `d8-d15` registers also be preserved ( in `armv7m` they are).
3. Maybe for `armv7a` make `setjmp(), longjmp()` functions like a they are for `armv7m`.

PR #289, opened 2023-09-08
armv7a: Use better set/longjmp implementation from armv7m 

agkaminski:
DONE: RTOS-600

## Motivation and Context

Fixes https://github.com/phoenix-rtos/libphoenix/issues/175

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/arch/armv7a/jmp.S
+++ b/arch/armv7a/jmp.S
@@ -6,90 +6,116 @@
  *
  * _setjmp, _longjmp, setjmp, sigsetjmp
  *
- * Copyright 2018 Phoenix Systems
- * Author: Kamil Amanowicz
+ * Copyright 2018, 2019, 2023 Phoenix Systems
+ * Author: Kamil Amanowicz, Aleksander Kaminski, Jakub Sarzynski
  *
  * This file is part of Phoenix-RTOS.
  *
  * %LICENSE%
  */
 
+ /* setjmp env buffer description
+  * Offset  Description
+  * 0       Signal mask flag - if != 0 then next value stores a signal mask
+  * 4       Signal mask
+  * 8       SP register
+  * 12-43   r4-r11 registers
+  * 44      LR register
+  * 48-79   d8-d15 registers
+  * 80      fpscr register
+  */
+
 #define __ASSEMBLY__
 
 .thumb
 .syntax unified
 
-.globl _setjmp
-.type _setjmp, %function
-_setjmp:
-	/* set jump address and stack pointer */
-	ldr r1, =(1f + 1)
-	mov r2, sp
-	/* store everythig in jmpbuf */
-	stm r0, {r1, r2, r4, r5, r6, r7, r8, r9, r10, r11, lr}
-	mov r0, #0
-	b 2f
-1:
-	/* this is where we land after the jump
-	 * just pass the return value which is in r3 */
-	mov r0, r3
-2:
-	bx lr
-.size _setjmp, .-_setjmp
+.text
+.align 2
 
-.globl _longjmp
-.type _longjmp, %function
-_longjmp:
-	/* move return value to r3.
-	 * longjmp should not make setjmp return 0 though */
-	cmp r1, #0
-	ite eq
-	moveq r3, #1
-	movne r3, r1
-	/* restore registers from jmpbuf */
-	ldmia r0!, {r1, r2, r4, r5, r6, r7, r8, r9, r10, r11, lr}
-	mov sp, r2
-	bx r1
-.size _longjmp, .-_longjmp
-
-.globl setjmp
-.type setjmp, %function
-setjmp:
-	push {r0, lr}
-	/* get signal mask */
-	mov r0, #0
-	mov r1, #0
-	bl signalMask
-	mov r3, r0
-	pop {r0, lr}
-	/* store everything in jmpbuf */
-	ldr r1, =(1f + 1)
-	mov r2, sp
-	stmia r0!, {r1, r2, r4, r5, r6, r7, r8, r9, r10, r11, lr}
-	str r3, [r0]
-	mov r0, #0
-	b 2f
-1:
-	/* this is where we land after the jump
-	 * r0 is now pointing to signal mask
-	 * which we want to restore */
-	push {r1, r4, lr}
-	ldr r4, [r0]
-	mov r0, r4
-	mov r4, r3
-	ldr r1, =0xffffffff
-	bl signalMask
-	/* set return value and pop the stack */
-	mov r0, r4
-	pop {r1, r4, lr}
-2:
-	bx lr
-.size setjmp, .-setjmp
 
 .globl sigsetjmp
 .type sigsetjmp, %function
+.align 2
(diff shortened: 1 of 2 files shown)


arch/armv7a/jmp.S
```diff
@@ -6,90 +6,107 @@
...
  */
 
+ /* setjmp env buffer description
+  * Offset	Description
+  * 0		Signal mask flag - if != 0 then next value stores a signal mask
+  * 4		Signal mask
+  * 8		SP register
+  * 12-44	r4-r11 registers
+  * 44		LR register
+  * 48-80	d8-d15 registers
```
kemonats, line 24:
```suggestion
  * 12-43	r4-r11 registers
  * 44		LR register
  * 48-79	d8-d15 registers
```

arch/armv7a/jmp.S
```diff
@@ -6,90 +6,107 @@
...
+  * 80		fpscr register
+  */
+
 #define __ASSEMBLY__
 
 .thumb
 .syntax unified
```
kemonats, line 31:
```suggestion
.thumb
.syntax unified

.text
```

arch/armv7a/jmp.S
```diff
@@ -6,90 +6,107 @@
...
-	ldr r1, =(1f + 1)
-	mov r2, sp
-	/* store everythig in jmpbuf */
-	stm r0, {r1, r2, r4, r5, r6, r7, r8, r9, r10, r11, lr}
+
+.globl sigsetjmp
+.type sigsetjmp, %function
+sigsetjmp:
+	/* Store signal mask if savesigs != 0 */
+	mov r2, #0
```
kemonats, line 38:
```suggestion
	movs r2, #0
```
2 bytes instead 4
same in line 44 
```
movs r0, #0
or 
mov r0, r2 /* r2 contains 0 */
```

arch/armv7a/jmp.S
```diff
@@ -6,90 +6,107 @@
...
+	cbz r1, 1f
+
+	/* Get signal mask */
+	/* Stack must be aligned to 8 bytes */
+	push {r0, r1, r3, lr}
 	mov r0, #0
```
kemonats, line 44:
```
movs r0, #0
or 
mov r0, r2 /* r2 contains 0 */
```

arch/armv7a/jmp.S
```diff
@@ -6,90 +6,107 @@
...
+	stmia r0!, {r1-r11, lr}
+
+#ifndef __SOFTFP__
+	/* Store fpu registers */
+	vstm r0!, {d8-d15}
+	vmrs r2, fpscr
+	str r2, [r0], #4
+#endif
+
+	mov r0, #0
```
kemonats, line 62:
```suggestion
	movs r0, #0
```

arch/armv7a/jmp.S
```diff
@@ -6,90 +6,107 @@
...
 	bx lr
+.size sigsetjmp, .-sigsetjmp
+
+.globl _setjmp
```
kemonats, line 66:
```suggestion
.size sigsetjmp, .-sigsetjmp


.globl _setjmp
```

arch/armv7a/jmp.S
```diff
@@ -6,90 +6,107 @@
...
+.type _setjmp, %function
+_setjmp:
+	mov r1, #0
```
kemonats, line 69:
```suggestion
	movs r1, #0
```

arch/armv7a/jmp.S
```diff
@@ -6,90 +6,107 @@
...
+	b sigsetjmp
 .size _setjmp, .-_setjmp
 
+.globl setjmp
```
kemonats, line 73:
```suggestion
.size _setjmp, .-_setjmp


.globl setjmp
```

arch/armv7a/jmp.S
```diff
@@ -6,90 +6,107 @@
...
+.type setjmp, %function
+setjmp:
+	mov r1, #1
```
kemonats, line 76:
```suggestion
	movs r1, #1
```

kemonats:
Two general comments:
1. You can add a `.align 2` directive before each function definition to align the function code to word size.
2. To consider defining the `jmp_buf` buffer size depending on the processor and the floating-point implementation used.
It is now defined like this:
https://github.com/phoenix-rtos/libphoenix/blob/345adf6103cb36c8fa9ef8b1c34e185759a93e75/include/setjmp.h#L25

arch/armv7a/jmp.S
agkaminski: According to docs it shouldn't matter: https://developer.arm.com/documentation/ddi0308/d/Thumb-Instructions/Alphabetical-list-of-Thumb-instructions/MOV--immediate-
and I'd rather not use `s` here, as it makes the code somewhat confusing

arch/armv7a/jmp.S
agkaminski: Again, I see no reason why should s produce different instruction encoding

arch/armv7a/jmp.S
agkaminski: Fixed, I also spotted tabs used for fancy alignment, changed to spaces

agkaminski:
Regarding general remarks:
1. It shouldn't be strictly needed, but might be beneficial from what I see, changed,
2. yeah, I've noticed that and I think that the common, largest size across all archs is good enough, at least until we get some exotic target with a huge `jmp_buff` needs

badochov (approved):

arch/armv7a/jmp.S
kemonats: Such instructions are used by the compiler.
A simple example:
```c
typedef unsigned int jmp_buf[34];
typedef jmp_buf sigjmp_buf;

extern int sigsetjmp(sigjmp_buf env, int savesigs);

int _setjmp(jmp_buf var)
{
    return sigsetjmp(var, 0);
}
```
and disassembly
```
Disassembly of section .text._setjmp:

00000000 <_setjmp>:
   0:   2100            movs    r1, #0
   2:   b508            push    {r3, lr}
   4:   f7ff fffe       bl      0 <sigsetjmp>
   8:   bd08            pop     {r3, pc}
   a:   bf00            nop
```

arch/armv7a/jmp.S
kemonats: You can use the second suggestion:
```
mov r0, r2 /* r2 contains 0 */
```

arch/armv7a/jmp.S
agkaminski: Ok, I see now. As it turns out there're only two 16-bit `mov` immediate instructions:
```
movs Rd, #imm8
mov<c> Rd, #imm8
```
There're no 16-bit `mov` instruction to be used outside IT block (https://developer.arm.com/documentation/dui0489/i/arm-and-thumb-instructions/mov). It all makes sense now to me, will change.

badochov (approved):

PR #293, opened 2023-09-19
Fix bug in formatting timestamps with gmtime_r

jmaksymowicz:
## Description
Timestamps that were far into the future (beyond about year 14000) were parsed incorrectly because the previous algorithm assumed too many leap years.

Related to https://github.com/phoenix-rtos/phoenix-rtos-project/issues/801 - in one example time on RTC was set correctly to maximum timestamp minus the given number of seconds, but displayed incorrectly due to this bug.

## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/255

## Checklist:

## Special treatment

### DIFF
--- a/time/time.c
+++ b/time/time.c
@@ -5,8 +5,8 @@
  *
  * time
  *
- * Copyright 2017 Phoenix Systems
- * Author: Andrzej Asztemborski
+ * Copyright 2017, 2023 Phoenix Systems
+ * Author: Andrzej Asztemborski, Jacek Maksymowicz
  *
  * This file is part of Phoenix-RTOS.
  *
@@ -20,6 +20,7 @@
 #include <string.h>
 #include <ctype.h>
 #include <stdio.h>
+#include <limits.h>
 
 
 char *tzname[2];
@@ -52,17 +53,6 @@ static int daysofmonth(int month, int leap)
 static inline int isleap(int year)
 {
 	return !(year % 400) || (!(year % 4) && (year % 100));
-}
-
-
-static inline int leapcount(int year)
-{
-	year -= 1;
-
-	if (year < 1972)
-		return 0;
-
-	return ((year - 1968) / 4) - ((year - 2000) / 100) + ((year - 2000) / 400);
 }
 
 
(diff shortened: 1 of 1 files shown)


time/time.c
```diff
@@ -22,6 +22,8 @@
 #include <stdio.h>
 
 
+#define DAYS_IN_400_YEARS 146097  // The exact number of days in a 400 year period, including all leap day rules.
```
agkaminski, line 25: Use C style comments
```suggestion
#define DAYS_IN_400_YEARS 146097  /* The exact number of days in a 400 year period, including all leap day rules. */
```

time/time.c
```diff
@@ -158,7 +160,7 @@ double difftime(time_t t1, time_t t2)
 struct tm *gmtime_r(const time_t *timep, struct tm *res)
 {
 	time_t sec = *timep, minute = 0, hou = 0, t;
-	int day = 0, leap, leapcnt;
+	int day = 0, leap, leapcnt, full_400_years;
```
agkaminski, line 163: Avoid snake_case, use camelCase
```suggestion
	int day = 0, leap, leapcnt, full400years;
```

time/time.c
```diff
@@ -181,8 +183,11 @@ struct tm *gmtime_r(const time_t *timep, struct tm *res)
 
 	res->tm_wday = (day + 4) % 7;
 
-	res->tm_year = 70 + day / 365;
-	leapcnt = leapcount(res->tm_year + 1900);
+	// Limit the following calculations to a 400 year period
+	full_400_years = day / DAYS_IN_400_YEARS;
+	day = day % DAYS_IN_400_YEARS;
+
+	leapcnt = leapcount(1970 + day / 365);  // This leap day count is approximate, but works within a 400 year period.
```
agkaminski, line 190: Perhaps it would be better to fix `leapcount()` instead

nalajcie:
please format the commit message title better:
 - use max 76 chars
 - use imperative mood
 - prefix with module name

For example: `time: fix counting leap years in gmtime_r`
(for more examples - see git history :))

You should also append JIRA ticket ID to the commit message trailer

time/time.c
jmaksymowicz: `leapcount()` is no longer used in `gmtime_r()` after my fixes, but it's still used in `_mktimeSkel()`. I can try to improve that function too.

time/time.c
```diff
@@ -155,59 +155,102 @@ double difftime(time_t t1, time_t t2)
...
+		int nYears;        /* Number of years this period has */
+		int isZeroDayLeap; /* Is day 0 of this period a leap day (unless a larger period was tried before) */
+	} periods[4] = {
+		{ 146097, 400, 1 },
+		{ 36524, 100, 0 },
+		{ 1461, 4, 1 },
+		{ 365, 1, 0 },
+	};
+
+	for (int i = 0; i < 4; i++) {
```
agkaminski, line 177:
```suggestion
	for (size_t i = 0; i < sizeof(periods) / sizeof(*periods); ++i) {
```

time/time.c
```diff
@@ -155,59 +155,102 @@ double difftime(time_t t1, time_t t2)
...
+struct tm *gmtime_r(const time_t *timep, struct tm *res)
+{
+	time_t seconds = *timep, minutes = 0, hours = 0, t;
+	int days = 0, year, isYearLeap;
 
-	res->tm_year = 70 + day / 365;
-	leapcnt = leapcount(res->tm_year + 1900);
-	day -= leapcnt;
-	res->tm_year = 70 + day / 365;
+	if ((t = seconds % 60) != seconds) {
```
agkaminski, line 216: MISRA
```suggestion
	t = seconds % 60;
	if (t != seconds) {
```

time/time.c
```diff
@@ -155,59 +155,102 @@ double difftime(time_t t1, time_t t2)
...
+		minutes = seconds / 60;
+		seconds = t;
+	}
 
-	if (leapcnt != leapcount(res->tm_year + 1900)) {
-		leapcnt = leapcount(res->tm_year + 1900);
-		++day;
+	if ((t = minutes % 60) != minutes) {
```
agkaminski, line 221:
```suggestion
	t = minutes % 60;
	if (t != minutes) {
```

time/time.c
```diff
@@ -155,59 +155,102 @@ double difftime(time_t t1, time_t t2)
...
+		hours = minutes / 60;
+		minutes = t;
 	}
 
-	day -= (res->tm_year - 70) * 365;
-	leap = isleap(res->tm_year + 1900);
-	res->tm_yday = day;
+	if ((t = hours % 24) != hours) {
```
agkaminski, line 226:
```suggestion
	t = hours % 24;
	if (t != hours) {
```

time/time.c
```diff
@@ -155,59 +155,102 @@ double difftime(time_t t1, time_t t2)
...
+	res->tm_wday = (days + 4) % 7;
+	year = epochDaysToYears(days, &days);
+	isYearLeap = isleap(year);
+	res->tm_year = year - 1900;
+	res->tm_yday = days;
 
-	for (res->tm_mon = 0; day >= 0 && res->tm_mon < 12; ++res->tm_mon)
-		day -= daysofmonth(res->tm_mon, leap);
+	for (res->tm_mon = 0; days >= 0 && res->tm_mon < 12; ++res->tm_mon)
+		days -= daysofmonth(res->tm_mon, isYearLeap);
```
agkaminski, line 242:
```suggestion
	for (res->tm_mon = 0; days >= 0 && res->tm_mon < 12; ++res->tm_mon) {
		days -= daysofmonth(res->tm_mon, isYearLeap);
	}
```

agkaminski (commented):
Is leapcount() still needed? It shoudn't be, as it's buggy, so it should be removed

time/time.c
```diff
@@ -155,59 +155,102 @@ double difftime(time_t t1, time_t t2)
...
 
-	res->tm_sec = sec;
-	res->tm_min = minute;
-	res->tm_hour = hou;
 
-	res->tm_wday = (day + 4) % 7;
+struct tm *gmtime_r(const time_t *timep, struct tm *res)
+{
+	time_t seconds = *timep, minutes = 0, hours = 0, t;
+	int days = 0, year, isYearLeap;
```
agkaminski, line 214: Nitpick - `int` days and year still can be overflowed by `time_t`. I see no point in handling times hundred of thousands of years into the future and past, so perhaps some check to reject too big values?

jmaksymowicz: It's a good idea, other parts of the system can't handle timestamps that large anyway (for example timer counts the number of microseconds in a 64-bit variable, so `time_t` generated by it will be 45 bits max). Using signed 32-bit int allows us to handle time_t that are about 46 bits long.

time/time.c
```diff
@@ -155,59 +144,114 @@ double difftime(time_t t1, time_t t2)
...
-	res->tm_min = minute;
-	res->tm_hour = hou;
+	return year;
+}
+
 
-	res->tm_wday = (day + 4) % 7;
+struct tm *gmtime_r(const time_t *timep, struct tm *res)
+{
+	if (*timep > (UINT64_MAX / 1000000)) {
```
agkaminski, line 202: Hmmm, I don't think this condition is ok, I think something along these lines:
```suggestion
	/* We need to be able to represent days as int */
	if ((sizeof(time_t) > sizeof(int)) && (*timep > ((time_t)INT_MAX * 1314000)) {
```

jmaksymowicz: OK, but this suggestion isn't correct - the number should be equal to the number of seconds in a day which is 86400.

agkaminski (approved):
LGTM

kemonats:
Some issues:
1. `time_t` is of `unsigned long long`, why check `seconds, minutes, hours < 0`, the compiler will skip these `if` statements.
2. if `gmtime_r()` can return `NULL` then this should be checked in the `ctime_r()` function:
```c
char *ctime_r(const time_t *timep, char *buf)
{
	struct tm t, *tp;

	tp = localtime_r(timep, &t);

	return (tp == NULL) ? NULL : asctime_r(&t, buf);
}
```
3. suggestion:
for 32-bit arm platform one 64-bit divide ( function __aeabi_uldivmod ) instead of six ( for this version of the compiler )
```c

	int seconds, minutes, hours;
	int days, month, year, isYearLeap;

	days = *timep / (24 * 60 * 60);
	seconds = *timep - (time_t)days * (24 * 60 * 60);
#if 0 /* insert it if `time_t` is signed */
	if (seconds < 0) {
		days--;
		seconds += (24 * 60 * 60);
	}
#endif

	hours = seconds / (60 * 60);
	seconds = seconds % (60 * 60);

	minutes = seconds / 60;
	seconds = seconds % 60;

	res->tm_sec = seconds;
	res->tm_min = minutes;
	res->tm_hour = hours;

/* the rest of the code */
```

agkaminski:
@kemonats 
1. We decided to slowly prepare the code for signed `time_t`. As you've noticed, these checks are optimized-out by the compiler, so it doesn't hurt anything (apart from perhaps some confusion and desire to remove them for now),
2. I agree, this needs to be fixed,
3. This way you sacrifice resolution of time_t - yes, it's still about 4000 years we're able to represent. It's your call @jmaksymowicz, entry check need to be changed in that case.

jmaksymowicz:
1. As @agkaminski said, we're preparing to use a signed type for `time_t`, so while correcting the algorithm I also made it ready for negative numbers.
2. I will fix it
3. It's a good point to change it, fewer 64-bit divisions is better. I don't see how it sacrifices resolution, the resulting variable `days` is same as before and the variable `seconds` will not exceed +/-86400.

agkaminski:
3. I misunderstood the code, there's no reduction in resolution. So I see no cons to kemonats's approach

time/time.c
```diff
@@ -155,59 +145,104 @@ double difftime(time_t t1, time_t t2)
...
+{
+	/* We need to be able to represent days as int */
+	if ((sizeof(time_t) > sizeof(int)) && (*timep > ((time_t)INT_MAX * (24 * 60 * 60)))) {
+		return NULL;
+	}
 
-	if (leapcnt != leapcount(res->tm_year + 1900)) {
-		leapcnt = leapcount(res->tm_year + 1900);
-		++day;
+	int seconds, minutes, hours;
```
agkaminski, line 208: Mind the warnining

time/time.c
```diff
@@ -155,59 +145,104 @@ double difftime(time_t t1, time_t t2)
...
 
-	res->tm_year = 70 + day / 365;
-	leapcnt = leapcount(res->tm_year + 1900);
-	day -= leapcnt;
-	res->tm_year = 70 + day / 365;
+struct tm *gmtime_r(const time_t *timep, struct tm *res)
+{
+	/* We need to be able to represent days as int */
+	if ((sizeof(time_t) > sizeof(int)) && (*timep > ((time_t)INT_MAX * (24 * 60 * 60)))) {
+		return NULL;
```
agkaminski, line 205: https://pubs.opengroup.org/onlinepubs/000095399/functions/gmtime_r.html

This will handle all other functions that use gmtime_r internally too

```suggestion
		errno = EOVERFLOW;
		return NULL;
```

jmaksymowicz: Done.

agkaminski (approved):
LGTM

PR #296, opened 2023-10-03
ubsan: provide C handlers for libubsan

gerard5:
## Description

Provides small but full implementation of C runtime handlers for the undefined behavior sanitisation mechanism.

The output log is only colored when printing to a tty device, if the output is a file the log will be plain text.

The goal of the implementation is to provide a minimalist implementation of ubsan that can be run on small targets, that's why this implementation does not provide C++ instrumentation and other handlers like:
  `uptr __ubsan_vptr_type_cache[];`
  `void __ubsan_handle_dynamic_type_cache_miss(..);`
  `void __ubsan_handle_dynamic_type_cache_miss_abort(..);`

## Example output

Buggy snippet used in this example
```C
int main(int argc, char **argv)
{
    unsigned int v = 1 << 31;
    int tab[] = { 0 };
    return (tab[argc] != v) ? 0 : 1;
}
```

imxrt117x (ubsan implementation from this PR):
2023-10-03-112508_1503x193_scrot

ia32-generic (ubsan implementation from this PR):

2023-10-03-105754_738x231_scrot

host-pc (default ubsan implementation provided with gcc):
2023-10-03-105954_840x284_scrot

## Section sizes

Comparison of the imxrt117x elf, compiled from the above "buggy example" (left with `-fsanitize=undefined`, right without)

2023-10-03-113543_2245x651_scrot

## Motivation and Context

JIRA: RTOS-633

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `ia32-generic`, `imxrt117x-nil`, `imxrt1064-evk`.

## Checklist:

## Special treatment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ include ../phoenix-rtos-build/Makefile.common
 SYSROOT := $(shell $(CC) $(CFLAGS) -print-sysroot)
 MULTILIB_DIR := $(shell $(CC) $(CFLAGS) -print-multi-directory)
 LIBC_INSTALL_DIR := $(SYSROOT)/lib/$(MULTILIB_DIR)
-LIBC_INSTALL_NAMES := libc.a libm.a libg.a libpthread.a
+LIBC_INSTALL_NAMES := libc.a libm.a libg.a libpthread.a libubsan.a
 HEADERS_INSTALL_DIR := $(SYSROOT)/usr/include
 LIBNAME := libphoenix.a
 
@@ -57,6 +57,7 @@ include termios/Makefile
 include time/Makefile
 include unistd/Makefile
 include wchar/Makefile
+include ubsan/Makefile
 
 #include test/Makefile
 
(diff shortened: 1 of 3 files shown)


agkaminski (approved):
LGTM

gerard5:
> Great to have this kind of functionality, thanks for the contribution!
> 
> Some notes to consider (not necessarily a as a part of this PR):
> 
>     * maybe it would be good to leave a comment in the source code regarding limitations of the current implementation (like lack of support for C++-specific features)?
> 
>     * did you try to compile llvm's libubsan (it's embedded in gcc)? maybe that's the way to support more features on bigger targets?
> 
>     * idea for the future - make libubsan a separate lib with external linkage to `ubsan_print` making it possible to be reused in kernel /plo / userspace apps

* Added descriptive comment about limitations in the commit message.
* I tried with LLVM's libubsan, but the resulting ELF size was huge on IA32 (so I gave up, and implemented something lighter), LLVM's libubsan was implemented in C++.
* Agree, it's good idea

PR #297, opened 2023-10-03
atexit: allow for calling atexit inside of atexit

badochov:
JIRA: RTOS-636
FIxes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/871

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32

## Checklist:

## Special treatment

### DIFF
--- a/stdlib/atexit.c
+++ b/stdlib/atexit.c
@@ -21,8 +21,14 @@
 /* The atexit_node structure uses uint32_t for flags that
  indicate whether to call a function with arguments or not, therefore
  why ATEXIT_MAX must be 32 as type "fntype" */
+#define ATEXIT_MAX 32
+
+
+typedef void (*destructor_t)(void);
+
+
 struct atexit_node {
-	void (*destructors[ATEXIT_MAX])(void);
+	destructor_t destructors[ATEXIT_MAX];
 	void *args[ATEXIT_MAX];
 	uint32_t fntype;
 	struct atexit_node *prev;
@@ -33,6 +39,7 @@ struct atexit_node {
 static struct {
 	handle_t lock;
 	struct atexit_node *head;
+	unsigned int idx;
 } atexit_common = { .head = &((struct atexit_node) {}) };
 
 
@@ -41,6 +48,7 @@ void _atexit_init(void)
 {
 	mutexCreate(&atexit_common.lock);
 	memset(atexit_common.head, 0, sizeof(struct atexit_node));
+	atexit_common.idx = 0;
 }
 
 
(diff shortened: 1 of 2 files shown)


adamdebek (approved):
LGTM, but I would leave some constant in stdlib like #define `ATEXIT_NODE_MAX 32`.

badochov:
> LGTM, but I would leave some constant in stdlib.h like #define `ATEXIT_NODE_MAX 32`.

why is it needed?

adamdebek:
If it will be needed in some source code, for example in atexit test I use this value to register 33 functions in order to allocate new node.

anglov:
> LGTM, but I would leave some constant in stdlib.h like #define `ATEXIT_NODE_MAX 32`.

Providing additional macros/symbols in standard library header that is not specified in documentation is against C/POSIX if not prefixed properly.

According to https://pubs.opengroup.org/onlinepubs/9699919799/functions/atexit.html it should be obtained via sysconf

As our sysconf is mostly stub, I identify removing ATEXIT_MAX from header as improvement.

nalajcie:
> According to https://pubs.opengroup.org/onlinepubs/9699919799/functions/atexit.html it should be obtained via sysconf

According to open group this should be defined in `limits.h` if it's invariant (compile-time const)  -https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html

IMHO we support more than 32 callbacks so the value is misleading, regarding other libc implementations:
 - `glibc` explicitly states in the comment of sub-include of limits.h that they don't provide fixed value for this const
 -  `newlib` defines the value only under cygwin
 - `musl`, `uclibc`, doesn't define the value at all

To summarize - I think we should remove the define from public headers and probably rename it in the implementation (as the name is misleading).

stdlib/atexit.c
```diff
@@ -87,28 +89,32 @@ static int _atexit_register(int isarg, void (*fn)(void), void *arg)
...
+			mutexUnlock(atexit_common.lock);
+			destructor();
+		}
+		mutexLock(atexit_common.lock);
+
+		if (atexit_common.idx == 0) {
+			struct atexit_node *last = atexit_common.head;
+			atexit_common.head = atexit_common.head->prev;
+			if (atexit_common.head != NULL) {
+				/* Free only if it's not the first node (statically allocated) */
```
agkaminski, line 111: This whole thing with first node being allocated statically is kinda a bad thing imho. We lose reference to it here (I guess no big deal, as we exit anyway). I'd consider lazy allocation (on first register ) instead. Just a suggestion to consider

agkaminski (approved):
My remark is not in scope of this PR, so feel free to ignore it for now

stdlib/atexit.c
nalajcie: Yeah, good idea. Just take note that the first `atexit()` calls can be done before main() - I think as a part of init array (C++ global objects). 

Side note: this is a good example of PR after which some C++ testing would be welcome (ensuring we won't break  C++ destructors for object of any storage class).

agkaminski: Ok, so this might be a reason for this solution being used here. This code was added as a part of C++ support, this further support this argument. So my idea would break C++

adamdebek:
> Providing additional macros/symbols in standard library header that is not specified in documentation is against C/POSIX if not prefixed properly.
> 
> According to https://pubs.opengroup.org/onlinepubs/9699919799/functions/atexit.html it should be obtained via sysconf
> 
> As our sysconf is mostly stub, I identify removing ATEXIT_MAX from header as improvement.

We can get ATEXIT_MAX via sysconf but in case of our implementation it is INT_MAX value since nodes are dynamically allocated. I think adding custom non POSIX defined constant symbol describing node size would be reasonable. I'm curios how it should be prefixed, never heard of that.

anglov:
> We can get ATEXIT_MAX via sysconf but in case of our implementation it is INT_MAX value since nodes are dynamically allocated. I think adding custom non POSIX defined constant symbol describing node size would be reasonable. I'm curios how it should be prefixed, never heard of that.

ISO/IEC 9899:TC3 7.1.3 Reserved identifiers

However - using special identifier (libc internals in fact)  will make your test not-portable. I believe that libc test in general should be possible to run and pass on host, Only exception is testing implementation defined behavior or phoenix extensions.

PR #298, opened 2023-10-04
pthread/cond: Use clock_id from condattr

agkaminski:
JIRA: RTOS-639

## Description

Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/876

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/include/sys/types.h
+++ b/include/sys/types.h
@@ -75,6 +75,7 @@ typedef struct pthread_mutexattr_t {
 typedef struct {
 	handle_t condh;
 	volatile int initialized;
+	clockid_t clock_id;
 } pthread_cond_t;
 
 
--- a/pthread/pthread.c
+++ b/pthread/pthread.c
@@ -25,8 +25,9 @@
 
 #define CEIL(value, size)	((((value) + (size) - 1) / (size)) * (size))
 
-#define PTHREAD_ONCE_DONE        0
-#define PTHREAD_ONCE_IN_PROGRESS 2
+#define PTHREAD_ONCE_DONE          0
+#define PTHREAD_ONCE_IN_PROGRESS   2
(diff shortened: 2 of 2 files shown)


agkaminski:
@lukileczo Did I understood correctly how it's supposed to work? I'm not so sure myself

include/sys/types.h
```diff
@@ -75,6 +75,7 @@ typedef struct pthread_mutexattr_t {
 typedef struct {
 	handle_t condh;
 	volatile int initialized;
+	clockid_t clock_id;
```
lukileczo, line 78: I don't think we should add `clockid_t` here, without `condattr` it should just use some default value

lukileczo: Ok nvm, I misunderstood this a little bit

lukileczo (dismissed):

pthread/pthread.c
```diff
@@ -845,6 +850,8 @@ int pthread_cond_init(pthread_cond_t *__restrict cond, const pthread_condattr_t
 {
 	int err;
 
+	cond->clock_id = attr->clock_id;
```
lukileczo, line 853: `attr` can be `NULL`, then `cond->clock_id` should have default value

lukileczo (changes requested):

pthread/pthread.c
anglov, line 853: attr can be null, and then default should be used.

anglov (commented):
Thanks for this change. It makes easier to write portable code and will secure from weird errors if potential changes of timers will be made in the future. 

One small remark

pthread/pthread.c
agkaminski: Thx, fixed

lukileczo (approved):

nalajcie:
pthread tests are now failing?

image

PR #304, opened 2023-10-19
pthread: fix key destructor calling and start function exit

badochov:
Fixes [https://github.com/phoenix-rtos/phoenix-rtos-project/issues/675](https://github.com/phoenix-rtos/phoenix-rtos-project/issues/675)

## Description
FIxes following posix imcompatibilities:
- `pthread_key_create`
```
An optional destructor function may be associated with each key value. At thread exit, if a key value has a non-NULL destructor pointer, and the thread has a non-NULL value associated with that key, the value of the key is set to NULL, and then the function pointed to is called with the previously associated value as its sole argument. The order of destructor calls is unspecified if more than one destructor exists for a thread when it exits.

If, after all the destructors have been called for all non-NULL values with associated destructors, there are still some non-NULL values with associated destructors, then the process is repeated. If, after at least {PTHREAD_DESTRUCTOR_ITERATIONS} iterations of destructor calls for outstanding non-NULL values, there are still some non-NULL values with associated destructors, implementations may stop calling destructors, or they may continue calling destructors until no non-NULL values with associated destructors exist, even though this might result in an infinite loop.
```
- `pthread_key_delete`
```
The pthread_key_delete() function shall be callable from within destructor functions. No destructor functions shall be invoked by pthread_key_delete(). Any destructor function that may have been associated with key shall no longer be called upon thread exit.
```
- `pthread_exit`
```
An implicit call to pthread_exit() is made when a thread other than the thread in which main() was first invoked returns from the start routine that was used to create it. The function's return value shall serve as the thread's exit status.
```
- `pthread_detach`
```
When a detached thread terminates, its
       resources are automatically released back to the system without
       the need for another thread to join with the terminated thread.
```
Also reworked how detached threads are handled

Code that previousle resoulted in faults:
```c
#include <pthread.h>
#include <stdio.h>
#include <string.h>
#include <assert.h>

static pthread_key_t key1;
static pthread_key_t key2;

static pthread_once_t key_once = PTHREAD_ONCE_INIT;

static int deleted = 0;

static void dstrctr(void *data)
{
	/* Constructor of deleted key should not be called. */
	assert(data != deleted);

	/* should be printed only once */
	printf("DSTR %p\n", data);

	/* Order of destructor calls is not specified. */
	/* Delete other key. */
	if (data == (void*)1) {
		pthread_key_delete(key2);
		deleted = 1;
	} else {
		pthread_key_delete(key1);
		deleted = 2;
	}
}

void *fn(void *arg)
{
	(void)pthread_key_create(&key1, dstrctr);
	(void)pthread_key_create(&key2, dstrctr);

	pthread_setspecific(key1, (void *)1);
	pthread_setspecific(key2, (void *)2);


	pthread_exit(0); /* Old implementation had a bug where pthread_exit wasn't called on a return from function. */
}

int main(void)
{
	pthread_t thread;
	pthread_create(&thread, NULL, fn, NULL);
	pthread_join(thread, NULL);
	return 0;
}

```

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia-32

## Checklist:

## Special treatment


pthread/pthread.c
```diff
@@ -1157,7 +1187,7 @@ void _pthread_atfork_child(void)
 {
 	mutexLock(pthread_common.pthread_atfork_lock);
 	pthread_fork_handlers_t *first = pthread_common.pthread_fork_handlers,
-		*curr = pthread_common.pthread_fork_handlers;
+							*curr = pthread_common.pthread_fork_handlers;
```
agkaminski, line 1190:
```suggestion
	*curr = pthread_common.pthread_fork_handlers;
```

agkaminski (commented):
Uh, lock releasing in functions make this code difficult, but I get it that it's necessary :)

agkaminski (dismissed):
LGTM

badochov:
After lastest change of pthreads by @lukileczo in #309 cleanup pop push were added, I had to adjust the code to the existence of them.
1) `pthread_exit` was adjusted to comply with linux's interpretation rules as they are confusing:
POSIX `pthread_exit` excerpt:
```
An implicit call to pthread_exit() is made when a thread other than the thread in which main() was first invoked returns from the start routine that was used to create it. The function's return value shall serve as the thread's exit status.
```
POSIX `pthread_cleanup_push` excerpt:
```
The cancellation cleanup handler shall be popped from the cancellation cleanup stack and invoked with the argument arg when:

The thread exits (that is, calls [pthread_exit()](https://pubs.opengroup.org/onlinepubs/009696699/functions/pthread_exit.html)).
```


LINUX interpretation:
```
•  When a thread terminates by calling [pthread_exit(3)](https://man7.org/linux/man-pages/man3/pthread_exit.3.html), all
          clean-up handlers are executed as described in the preceding
          point.  (Clean-up handlers are not called if the thread
          terminates by performing a return from the thread start
          function.)
```


2)
Fixed cleanup handlers not being called in case of cancelling not self thread

pthread/pthread.c
```diff
@@ -457,24 +480,37 @@ int pthread_equal(pthread_t t1, pthread_t t2)
...
+		if (cleanup != 0) {
+			mutexLock(pthread_common.pthread_list_lock);
+			_pthread_do_cleanup(ctx);
+			mutexUnlock(pthread_common.pthread_list_lock);
+		}
+
+		pthread_key_cleanup(ctx);
 		ctx->retval = value_ptr;
-		cleanup = pthread_key_cleanup(ctx);
 		_pthread_ctx_put(ctx);
```
lukileczo, line 494: We don't have `pthread_list_lock` here - use `pthread_ctx_put`? Or execute `_pthread_do_cleanup` after `pthread_key_cleanup` and then do `_pthread_ctx_put` or `pthread_ctx_put` accordingly to avoid lock/unlock twice.

badochov:
Also I spotted that `signal_cancel` is send only after everything is cleanup it seems racy to. Is this valid @agkaminski ?
https://github.com/phoenix-rtos/libphoenix/blob/bd3db90465b9ae2fab89ae72a0b8842ba7db2d13/pthread/pthread.c#L433

pthread/pthread.c
badochov: put should not be called here, mixup during rebase

lukileczo:
`signal_cancel` calls `proc_threadDestroy` immediately in kernel, so all resources should be freed before. Our cancel implementation is not fully POSIX compliant, as we have no cancellation points (which would require many changes across `libphoenix`)

lukileczo (dismissed):
LGTM

pthread/pthread.c
```diff
@@ -556,7 +590,7 @@ int pthread_attr_setschedparam(pthread_attr_t *attr,
 		return EINVAL;
 
 	if (param->sched_priority > sched_get_priority_max(SCHED_RR) ||
-			param->sched_priority < sched_get_priority_min(SCHED_RR))
+		param->sched_priority < sched_get_priority_min(SCHED_RR))
```
agkaminski, line 593: clang-format is mistaken ;)

pthread/pthread.c
```diff
@@ -618,9 +652,10 @@ int pthread_attr_getscope(const pthread_attr_t *attr,
 
 int pthread_attr_setdetachstate(pthread_attr_t *attr, int detachstate)
 {
-	if (detachstate != PTHREAD_CREATE_DETACHED ||
-			detachstate != PTHREAD_CREATE_JOINABLE)
+	if ((detachstate != PTHREAD_CREATE_DETACHED) ||
+		(detachstate != PTHREAD_CREATE_JOINABLE)) {
```
agkaminski, line 656: clang-format is mistaken ;)

agkaminski (commented):
Clang-format stuff nitpick, otherwise LGTM

agkaminski (approved):

PR #313, opened 2023-11-29
Change libphoenix initialization

badochov:
This change prepares libphoenix to be used in dynamic loader, where crt0 is not included.

Moved constructor and destrutor calling into crt0. Wrapped libphoenix initialization into a single function to reduce crt0 and libphoenix intertwine.

JIRA: RTOS-694

## Description
[RTOS-694] #done

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


agkaminski (commented):
LGTM, but add yourself to authors list

_startc.c
```diff
@@ -38,14 +79,10 @@ void _startc(int argc, char **argv, char **env)
...
-	_errno_init();
-	_malloc_init();
-	_env_init();
-	_signals_init();
-	_file_init();
+	_libc_init();
+
 	_init_array();
-	_pthread_init();
+	atexit(_fini_array);
```
nalajcie, line 85: not sure if 100% correct - is it allowed for constructors to call `atexit()`? If so - the destructors might not be called last? (probably not regulated by POSIX, but might be relevant for C++ constructors/destructors of global static objects?)

badochov: I cannot find anything on POSIX about constructors at all, however I see no reaseon not to change the order, thanks for spotting :rocket:

agkaminski (approved):
LGTM

badochov:
@agkaminski Is the CI failling on imx1060 a known issue or sth new creted by this pr?

badochov:
> > @agkaminski Is the CI failling on imx1060 a known issue or sth new creted by this pr?
> 
> I don't think I've seen it, looks like a user stack overflow. Might be caused by this PR I guess - maybe `_libc_init` was inlined along with function(s) it's calling? Perhaps `__attribute__((noinline))` on `_libc_init` would solve this. But it would be best to inspect build artifacts first to see if it indeed happened.

Locally it doesn't inline _libc_init, also seems to be too undeterministic for that to be the case

badochov:
@agkaminski I added noinlien to _init_array as it got inlined after rewrite and increased stack size of `_startc`.
I rerun test 3 times and the issue on imx1060 haven't reoccured

badochov:
> @agkaminski I added noinlien to _init_array as it got inlined after rewrite and increased stack size of `_startc`. I rerun test 3 times and the issue on imx1060 haven't reoccured

Through the night I run the CI 2 more times still no errors

agkaminski (approved):

PR #331, opened 2024-01-24
unistd/crypt: fix hash byte order

lukileczo:
## Description

Previously data passed to `base64_encode` was casted directly from `unsigned int` to `unsigned char` array.
This PR fixes this by explicitly copying appropriate values of the bytes.
Hash is always base64-encoded as little endian.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `ia32-generic-qemu`, `sparcv8leon3-gr716-mimas`

## Checklist:

## Special treatment

- [ ] This PR needs additional PRs to work

### DIFF
--- a/unistd/crypt.c
+++ b/unistd/crypt.c
@@ -18,7 +18,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 
-static void base64_encode(unsigned char *bytes, int len, unsigned char *res);
+static void base64_encode(uint8_t *bytes, size_t len, unsigned char *res);
 
 #define _SHA1_CRYPT_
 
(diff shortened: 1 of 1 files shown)


unistd/crypt.c
```diff
@@ -176,6 +178,15 @@ static void sha1_process(sha1_ctx *ctx, const char *msg)
 
 	if (ctx->ext)
 		sha1_processBlock(ctx);
+
+	/* Fix byte order */
```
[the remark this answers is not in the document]
lukileczo: Are we ok with `arpa/inet.h` dependency here?

nalajcie:
will this make hashes in `/etc/passwd` invalid? 

If so, we can't merge it in current way ("old" hashes need to be supported indefinitely).

unistd/crypt.c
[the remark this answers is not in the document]
agkaminski: Eh, perhaps it's not worth it, you're right

agkaminski:
> will this make hashes in `/etc/passwd` invalid?
> 
> If so, we can't merge it in current way ("old" hashes need to be supported indefinitely).

Sooo should we go with LE hashes from now on?

nalajcie:
> > will this make hashes in `/etc/passwd` invalid?
> > If so, we can't merge it in current way ("old" hashes need to be supported indefinitely).
> 
> Sooo should we go with LE hashes from now on?

I don't know specifics, but AFAIK the hash algorithm is embedded in the passwd hash - maybe we could introduce new HASH algorithm (real SHA1) and keep the old hashes functional? 

Currently I don't see any other reliable way of migrating `/etc/passwd` hashes to new algorithm :/.

agkaminski (approved):
LGTM

unistd/crypt.c
```diff
@@ -176,6 +178,15 @@ static void sha1_process(sha1_ctx *ctx, const char *msg)
...
+	for (size_t i = 0; i < sizeof(ctx->hash.h) / sizeof(ctx->hash.h[0]); i++) {
+		uint32_t tmp = ctx->hash.h[i];
+		ctx->hash.res[i * 4] = (tmp & 0xffu);
+		ctx->hash.res[i * 4 + 1] = (tmp & 0xff00u) >> 8;
+		ctx->hash.res[i * 4 + 2] = (tmp & 0xff0000u) >> 16;
+		ctx->hash.res[i * 4 + 3] = (tmp & 0xff000000u) >> 24;
```
kemonats, line 188: Maybe:
```c
uint32_t __builtin_bswap32 (uint32_t x)
```

lukileczo: I've already merged it - but regarding your comment, I'd prefer to write endian-independent code where possible.

lukileczo: This gets optimized on LE systems anyway.

PR #339, opened 2024-02-05
cleanup headers

lukileczo:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


unistd/alarm.c
```diff
@@ -26,7 +26,7 @@ static struct {
 	char stack[1024] __attribute__((aligned(8)));
 	handle_t cond, lock;
 	time_t wakeup;
-	unsigned tid;
+	handle_t tid;
```
agkaminski, line 29: thread id (tid) is just int (like pid, but it has it's own pid_t)

agkaminski (dismissed):

include/arch/armv7m/stdint.h
```diff
@@ -16,19 +16,21 @@
...
+typedef __s64 int64_t;
 
-typedef unsigned char uint8_t;
-typedef unsigned short uint16_t;
-typedef unsigned int uint32_t;
-typedef unsigned long long uint64_t;
+typedef __u8 uint8_t;
+typedef __u16 uint16_t;
+typedef __u32 uint32_t;
+typedef __u64 uint64_t;
```
anglov, line 33: Potentially it  could be moved to generic include/stdint.h 

Then we may want to not used it in least/fast/ptr/max types  defined below or firstly typedef and then include.

side note: least type could also be moved to generic include/stdint.h in section 
 USE_STANDARD_TYPES_STDINT

include/stdio.h
```diff
@@ -28,6 +28,9 @@ extern "C" {
 #endif
 
 
+#include <phoenix/posix-stdio.h>
```
anglov, line 31: SEEK_SET/CUR/END definitions can be removed from this file I think

include/sys/time.h
```diff
@@ -25,6 +25,9 @@ extern "C" {
 #endif
 
 
+#include <phoenix/posix-timespec.h>
```
anglov, line 28: timespec shouldn't be included in this header.

anglov (commented):
Major and nice refactoring. I am happy that we decided to clean up kernel-libphoenix includes.

You should probably add yourself as an author to files that you majorly changed.

This refactoring doesn't solve everything (eg. types.h should be modularized) to fully conform with POSIX, but it's major improvement.

I point small things that can improve your change a little.

arch/armv7m/reboot.c
```diff
@@ -17,11 +17,11 @@
 #include <sys/platform.h>
 
 #if defined(__CPU_STM32L4X6)
-#include <phoenix/arch/stm32l4.h>
+#include <phoenix/arch/armv7m/stm32/l4/stm32l4.h>
 #elif defined(__CPU_IMXRT117X)
-#include <phoenix/arch/imxrt1170.h>
+#include <phoenix/arch/armv7m/imxrt/11xx/imxrt1170.h>
 #else
```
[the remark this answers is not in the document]
lukileczo: Thanks, for now I'll leave the ifdefs as they are because they're only needed in one file, but if ever necessary elsewhere I agree that this could be moved to separate header.

agkaminski (dismissed):

include/stdint.h
```diff
@@ -26,10 +26,34 @@
...
+typedef int16_t int_least16_t;
+typedef int32_t int_least32_t;
+typedef int64_t int_least64_t;
+
+typedef uint8_t uint_least8_t;
+typedef uint16_t uint_least16_t;
+typedef uint32_t uint_least32_t;
+typedef uint64_t uint_least64_t;
+
+#ifdef _USE_STANDARD_TYPES_STDINT
```
anglov, line 55: To fully conform with idea of this define, this ifdef should be above  `/* 7.18.1.1 Exact-width integer types */`

in fact C doesn't require to have uint(8/16/32/64)_t types defined.

It's minor as we don't want to support weird architectures but would be great to preserve idea here on wipe it out :)

agkaminski (approved):

PR #342, opened 2024-02-09
arch: add armv8m

damianloew:
## Description

arch: add armv8m

## Motivation and Context

Finalizing the nrf9160 basic port

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (nrf9160 devkit).

## Checklist:

## Special treatment

### DIFF
--- a/arch/armv8m/jmp.S
+++ b/arch/armv8m/jmp.S
@@ -1,0 +1,121 @@
+/*
+ * Phoenix-RTOS
+ *
+ * libphoenix
+ *
+ *
+ * _setjmp, _longjmp, setjmp, sigsetjmp
+ *
+ * Copyright 2018, 2019, 2024 Phoenix Systems
+ * Author: Kamil Amanowicz, Aleksander Kaminski, Jakub Sarzyński, Damian Loewnau
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+ /* setjmp env buffer description
+  * Offset  Description
+  * 0       Signal mask flag - if != 0 then next value stores a signal mask
+  * 4       Signal mask
+  * 8       SP register
+  * 12-43   r4-r11 registers
+  * 44      LR register
+  * 48-79   d8-d15 registers
+  * 80      fpscr register
+  */
+
+#define __ASSEMBLY__
+
+.thumb
+.syntax unified
+
+.text
+.align 2
+
+
+.globl sigsetjmp
+.type sigsetjmp, %function
+.align 2
+sigsetjmp:
+	/* Store signal mask if savesigs != 0 */
+	movs r2, #0
+	cbz r1, 1f
+
+	/* Get signal mask */
+	/* Stack must be aligned to 8 bytes */
+	push {r0, r1, r3, lr}
+	movs r0, #0
+	mov r1, r0
+	bl signalMask
+	mov r2, r0
+	pop {r0, r1, r3, lr}
+
+1:
+	/* Store mask (fake or real) and general registers */
+	mov r3, sp
+	stmia r0!, {r1-r11, lr}
+
+#ifndef __SOFTFP__
+	/* Store fpu registers */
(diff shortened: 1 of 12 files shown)


kemonats:
I have a small comment. I suggest adding a `.cpu` directive to each `*.S` file at its begining.

damianloew:
> I have a small comment. I suggest adding a `.cpu` directive to each `*.S` file at its begining.

Hmm, but those files are intended for various armv8 cpus, like Cortex-M23 and so on (I know that now they are not provided, but maybe they will be available in the future)

agkaminski:
General remark - modify copyright headers. You are an author of all these files, of some (like Makefile) you are the only author

arch/armv8m/jmp.S
```diff
@@ -0,0 +1,121 @@
...
+  * 80      fpscr register
+  */
+
+#define __ASSEMBLY__
+
+.thumb
+.syntax unified
+
+.text
+.align 2
```
agkaminski, line 34: Alignment might cause some performance penalty. IMHO better to set it to 4, to make it predictable

arch/armv8m/jmp.S
```diff
@@ -0,0 +1,121 @@
...
+
+
+.globl sigsetjmp
+.type sigsetjmp, %function
+.align 2
```
agkaminski, line 39: Redundant
```suggestion
```

arch/armv8m/reboot.c
```diff
@@ -0,0 +1,50 @@
...
+int reboot_reason(uint32_t *val)
+{
+	platformctl_t pctl = {
+		.action = pctl_get,
+		.type = pctl_reboot,
+	};
+
+	*val = 0;
+	if (platformctl(&pctl) < 0)
+		return -1;
```
agkaminski, line 46:
```suggestion
	if (platformctl(&pctl) < 0) {
		return -1;
	}
```

arch/armv8m/jmp.S
damianloew: Do we want to leave 4 for other arm architectures?

arch/armv8m/tls.c
```diff
@@ -0,0 +1,34 @@
...
+
+void *read_tls_ptr(void)
+{
+	return (void *)arm_tls_ptr;
+}
+
+
+void __attribute__((naked)) __aeabi_read_tp(void)
+{
+	__asm__ volatile(
```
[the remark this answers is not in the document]
damianloew: Can we leave it without a space to be compliant with clang-format? Like for other architectures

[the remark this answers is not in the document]
agkaminski: I'd rather not, clang-format is confused in this case, as it thinks it's a function - it's not. It would be best to use clang-format off on every inline asm instead

arch/armv8m/jmp.S
agkaminski: Other arm archs aren't in the scope of this PR. Imho it's best to change alignment to 4 everywhere eventually, as @lukileczo found out it's faster indeed

damianloew: Oh I see that it has been changed recently

lukileczo:
With the merge of https://github.com/phoenix-rtos/libphoenix/pull/347, I've created conflicts for your port _again_. I've made good chunk of code previously split into `armv7a` and `armv7m` common - in `arm` directory, which has `v7a` and `v7m` subdirectories, so you can remove a fair amount of previously copied code from this PR.

agkaminski:
@damianloew support added in https://github.com/phoenix-rtos/libphoenix/commit/3375557c923188ec8e773bf3a7780cbb13aa560d can we close this PR?

damianloew:
> @damianloew support added in [3375557](https://github.com/phoenix-rtos/libphoenix/commit/3375557c923188ec8e773bf3a7780cbb13aa560d) can we close this PR?

Sure :+1:

PR #346, opened 2024-02-16
unistd/dir: fix resolve_path behavior on trailing slashes

### LINKED ISSUE #682: stat() completion successful and set wrong errno, when a component of the path prefix is not a directory
stat() completion successful and set wrong errno, when a component of the path is existing regular file, not a directory

Tested between ia32-generic and host-generic

Code to reproduction:
```
#include <stdio.h>
#include <string.h>
#include <fcntl.h>
#include <sys/stat.h>
#include <sys/types.h>
#include <unistd.h>
#include <errno.h>


int main(void)
{
	struct stat buf;

	FILE *fp = fopen("testfile.txt", "w");
	fclose(fp);

	errno = 0;
	stat("testfile.txt/", &buf);
	printf("Regular File: %d\n",S_ISREG(buf.st_mode));
 	printf("Directory: %d\n",S_ISDIR(buf.st_mode));
	printf("Errno: %d, %s\n", errno, strerror(errno));

	remove("testfile.txt");

	return 0;
}
```

Outcome from ia32-generic:
image

Outcome from host-generic:
image

Documentation:
image
image

jmaksymowicz:
## Description
Currently, any part of the system that uses the `resolve_path` function (including `stat`, `lstat`, `realpath` and many others) don't consider the trailing slashes at the end of the path. For directories this is correct behavior, but if the last component of a path is not a directory, then the function should return with error and set `errno` to `ENOTDIR`.

The change currently breaks CI, as one of the tests for `psh` checks that trailing slashes for files are ignored. 

## Motivation and Context
Closes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/682 and brings API more in line with [Open Group specification](https://pubs.opengroup.org/onlinepubs/9699919799/functions/realpath.html)

## Types of changes

This change may break existing software (including `psh` scripts) if it depends on the current trailing slash behavior. This may be difficult to detect if a path is generated automatically.

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/324

## Checklist:

## Special treatment

### DIFF
--- a/unistd/dir.c
+++ b/unistd/dir.c
@@ -173,13 +173,14 @@ static int _resolve_abspath(char *path, char *result, int resolve_last_symlink, 
 		p = strchrnul(node, '/');
 		const size_t node_len = p - node;
 
-		/* skip trailing slashes to properly detect leaf node */
+		const int is_leaf = (*p == '\0') ? 1 : 0;
 		while (*p == '/') {
 			p += 1;
 		}
 
-		const int is_leaf = (*p == '\0') ? 1 : 0;
-
+		/* TODO: this is too permissive, because it allows creating files (not just directories)
+		 * with a path that has trailing slashes */
+		const int is_leaf_after_slashes = (*p == '\0') ? 1 : 0;
 		/* check for '.' and '..' */
 		if (node_len == 0) { /* multiple '/' */
 			continue;
@@ -202,11 +203,6 @@ static int _resolve_abspath(char *path, char *result, int resolve_last_symlink, 
 		r += node_len;
 		*r = '\0';
 
-		if ((is_leaf != 0) && (resolve_last_symlink == 0)) {
-			/* WARN: slight inconsistency: we're not checking if the path actually exists in this case (TODO?) */
-			break;
-		}
-
 		/*
 		 * Number of initial unused `path` bytes that can be used to resolve current node. If link content is
 		 * longer we will not be able to merge it with remaining path and still fit in PATH_MAX buffer
@@ -217,16 +213,27 @@ static int _resolve_abspath(char *path, char *result, int resolve_last_symlink, 
 		ssize_t symlink_len = _readlink_abs(result, path, readlink_max_len);
 
 		if (symlink_len < 0) {
-			if (errno == EINVAL) { /* not a symlink */
+			if (symlink_len == -EISDIR) {
 				continue;
 			}
-			else if ((errno == ENOENT) && (is_leaf != 0) && (allow_missing_leaf != 0)) { /* non-esixting leaf */
+			if (symlink_len == -EINVAL) { /* not a symlink or directory */
+				if (is_leaf == 0) {
+					return SET_ERRNO(-ENOTDIR);
+				}
+				else {
+					continue;
+				}
+			}
+			else if ((symlink_len == -ENOENT) && (is_leaf_after_slashes != 0) && (allow_missing_leaf != 0)) { /* non-existing leaf */
 				break;
 			}
 			else {
-				/* errno set by _readlink_abs */
-				return -1;
+				return SET_ERRNO(symlink_len);
 			}
+		}
+
+		if ((is_leaf != 0) && (resolve_last_symlink == 0)) {
+			break;
 		}
 
 		if (symlink_len == 0) {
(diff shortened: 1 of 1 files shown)


unistd/dir.c
```diff
@@ -488,8 +500,10 @@ static ssize_t _readlink_abs(const char *path, char *buf, size_t bufsiz)
 	if (msg.o.attr.err != EOK)
 		return SET_ERRNO(msg.o.attr.err);
 
-	if (!S_ISLNK(msg.o.attr.val))
-		return SET_ERRNO(-EINVAL);
+	if (!S_ISLNK(msg.o.attr.val)) {
+		SET_ERRNO(-EINVAL);
+		return S_ISDIR(msg.o.attr.val) ? -EISDIR : -1;
```
nalajcie, line 505: ugly API (returning side info as an error). It's our internal function, maybe add explicit output param (it would increase the readability of the caller code)?

unistd/dir.c
```diff
@@ -247,8 +255,11 @@ static int _resolve_abspath(char *path, char *result, int resolve_last_symlink,
 		}
 
 		/* prepend resolved symlink to remaining unresolved path */
-		p -= 1;
-		*p = '/';
+		if (*p != '\0') {
```
nalajcie, line 258: is this a related change? is there any test for a case this is needed?

nalajcie:
Please also copy the problem description from PR to the commit message body (or write something similar there to leave a trail)

unistd/dir.c
jmaksymowicz: Yes. Previously if there was a path like `/path/to/symlink` and `symlink` had contents `/realpath`, after fetching symlink next iteration would look up `/realpath/`. It didn't matter because trailing slashes were always ignored during lookup. This change is necessary to prevent adding the slash when not necessary (when symlink is the final part of the path).

Adding the slash is necessary when looking up something like `/path/to/symlink/file_in_dir`, because we need to look up `/realpath/file_in_dir`.

No additional test cases are necessary, it is already covered by looking up a symlink to a file and trying to traverse through symlink to directory.

unistd/dir.c
jmaksymowicz: I think it's a good idea, I'll change the API a bit to do this more elegantly.

unistd/dir.c
```diff
@@ -229,6 +231,12 @@ static int _resolve_abspath(char *path, char *result, int resolve_last_symlink,
 			}
 		}
 
+		if ((is_leaf != 0) && (resolve_last_symlink == 0)) {
+			/* TODO: in this case it's more efficient to check if file exists
```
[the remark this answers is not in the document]
jmaksymowicz: 1. Checking if path exists and is a directory before call to `_readlink_abs` may be slightly more efficient in the case when `(is_leaf != 0) && (resolve_last_symlink == 0)` and the symlink exists. With this code the symlink's contents will be read even though it's not necessary. It's probably not a big performance concern.
2. You're right, when doing `lstat` on `/path/to/symlink/`, if `symlink` exists and is a symlink then the call will succeed despite the trailing slash. I think in this case we need to check `if (ends_with_slash != 0)` and exit with `errno == -ENOTDIR`.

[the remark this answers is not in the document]
nalajcie: > 1. Checking if path exists and is a directory before call to `_readlink_abs` may be slightly more efficient in the case when `(is_leaf != 0) && (resolve_last_symlink == 0)` and the symlink exists. With this code the symlink's contents will be read even though it's not necessary. It's probably not a big performance concern.

still, this should be **instead of**, not **before** as in the mentioned case you don't want to call `_readlink_abs` at all (if we really want to improve performance).  I think it's not worth the additional code complexity (not a hot path), so IMHO `TODO` should be removed

Regarding (2) -> maybe add a test for this particular "corner case" to avoid regressions in the future (this code is hard enough to understand with all special cases ;))

ziemleszcz:
Please rebase this PR.

unistd/dir.c
```diff
@@ -251,8 +262,11 @@ static int _resolve_abspath(char *path, char *result, int resolve_last_symlink,
 		}
 
 		/* prepend resolved symlink to remaining unresolved path */
-		p -= 1;
-		*p = '/';
+		if (*p != '\0') {
+			p -= 1;
+			*p = '/';
+		}
```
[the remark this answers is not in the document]
jmaksymowicz: Fixed

PR #357, opened 2024-04-22
libphoenx/scanf: Fix SEGFAULT caused by invalid processing of %n

astalke:
## Description

- Fix SEGFAULT caused by invalid processing of %n
- MISRA

Fixes: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1059

JIRA: RTOS-825

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/stdio/scanf.c
+++ b/stdio/scanf.c
@@ -48,6 +48,7 @@
 #define CT_STRING 2 /* %s conversion */
 #define CT_INT    3 /* %[dioupxX] conversion */
 #define CT_FLOAT  4 /* %[aefgAEFG] conversion */
+#define CT_NONE   5 /* No conversion (ex. %n) */
 
 
 static const unsigned char *__sccl(char *tab, const unsigned char *fmt)
@@ -113,7 +114,7 @@ static int scanf_parse(char *ccltab, const char *inp, int *inr, char const *fmt0
 	char *p, *p0;
 	char buf[32];
 
-	static short basefix[17] = { 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16 };
+	static const short basefix[17] = { 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16 };
 
 	*inr = strlen(inp);
 
@@ -226,6 +227,8 @@ static int scanf_parse(char *ccltab, const char *inp, int *inr, char const *fmt0
 				case '9':
 					width = width * 10 + c - '0';
 					continue;
+				default:
+					break;
 			}
 
 			/*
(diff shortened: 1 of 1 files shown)


agkaminski:
Hmm, tbh I don't get this change. Block handling `%n` has been moved later in the code, but I can't see any functional change in doing so. Could you please briefly explain this change?

astalke:
> Hmm, tbh I don't get this change. Block handling `%n` has been moved later in the code, but I can't see any functional change in doing so. Could you please briefly explain this change?

After reviewing my changes, I've noticed that there is an easier fix, so I will upload it soon and let CI check if it really is enough. Bug is caused by a use of [`continue` instead of `break`](https://github.com/phoenix-rtos/libphoenix/pull/357/files#diff-41d1b0f052c54b110b68dfe6c2b2a9cbe56fd85b9f67c7aed99373cf6d947226L305), lack of default label in this switch and reusing `c` variable in the next switch statement. Basically, `continue` causes loop to execute one too many times, so `c == 0`, there is no `default` label to catch that and due to reusing of `c` in the next switch statement, we execute [`case CT_CHAR`](https://github.com/phoenix-rtos/libphoenix/pull/357/files#diff-41d1b0f052c54b110b68dfe6c2b2a9cbe56fd85b9f67c7aed99373cf6d947226L330).

agkaminski (dismissed):
LGTM

stdio/scanf.c
```diff
@@ -318,7 +320,10 @@ static int scanf_parse(char *ccltab, const char *inp, int *inr, char const *fmt0
 					else {
 						*va_arg(ap, int *) = nread;
 					}
-					continue;
+					break;
```
Darchiv, line 323: This works because `c == 'n'` so it won't match any of `CT_*` values in the conversion `switch` below. Whitespace will still be skipped anyway. Maybe there should be `CT_NONE` value which performs no conversion is used for `'n'` and `default` cases, together with `NOSKIP`?

jmaksymowicz: After my changes `c` is not used in the switch statement and `CT_NONE` is used for handling `%%` and `%n`. Default case handles unknown conversion specifiers by ending the conversion (behavior is undefined [according to documentation](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fscanf.html), this is what glibc does).

stdio/scanf.c
```diff
@@ -154,8 +154,13 @@ static int scanf_parse(char *ccltab, const char *inp, int *inr, char const *fmt0
 
 		width = 0;
 		flags = 0;
+		int convType = CT_NONE;
```
Darchiv, line 157: Variable declarations should be placed at the start of the block, at least.

jmaksymowicz: Done

Darchiv (approved):

PR #363, opened 2024-06-20
locale: Fix incomplete locale.h

Darchiv:
DONE: RTOS-856

Currently, `struct lconv` has only one standard member, whereas there are many more. This change introduces all of them, modernizes and standardizes function declarations and definitions. ~~Unfortunately, this makes `fixed_locale` non-const, but modification of this struct by the pointer returned from `localeconv()` is prohibited by the standard.~~

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/locale/locale.c
+++ b/locale/locale.c
@@ -5,17 +5,49 @@
  *
  * locale.c
  *
- * Copyright 2018 Phoenix Systems
+ * Copyright 2018, 2024 Phoenix Systems
  * Author: Michal Miroslaw
  *
  * This file is part of Phoenix-RTOS.
  *
  * %LICENSE%
  */
-#include "locale.h"
+#include <limits.h>
+#include <locale.h>
 
+/* for regex/collate.c */
 char *_PathLocale;
 
-const struct lconv _fixed_locale = {
-	"."
+
+static const struct lconv fixed_locale = {
+	.decimal_point = ".",
+	.thousands_sep = "",
+	.grouping = "",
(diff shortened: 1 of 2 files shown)


nalajcie (dismissed):

locale/locale.c
```diff
@@ -5,17 +5,49 @@
...
+#include <limits.h>
+#include <locale.h>
 
+/* for regex/collate.c */
 char *_PathLocale;
 
-const struct lconv _fixed_locale = {
-	"."
+
+static struct lconv fixed_locale = {
```
agkaminski, line 22: Hmm, why has this be non-const? If it's because of the usage in `localeconv` (as it sadly returns non-const ptr), then perhaps we might just cast `const` away there?

agkaminski: Not the best option of course, so it's ok as is for me

agkaminski (dismissed):

locale/locale.c
nalajcie: I'm a little bit afraid of  user trying to modify the object residing on XIP (rodata) which might trigger flash write (possible waitstates/ realtime breakage). 

With MPU enabled it's not an attack vector (unsafe aplications can be prevented from writing to XIP) so maybe we can make it `const`?

Note: in C99/C11 modifying struct returned from `localeconv` is UB so we can safely throw `MemMgmtFault  :).

agkaminski (approved):

PR #370, opened 2024-07-23
stdlib, stdio: improve float string parsing

jmaksymowicz:
## Description
* `strtof`, `strtod` and `strtold` can parse infinity and NaN string representations, as well as hexadecimal floating point representations
* Code size was reduced by extracting operations done multiple times into functions.
* Performance of `strtof` and `strtod` was improved. Now `uint64_t` is used as an intermediate container for digits that were read, instead of `long double` used previously. This reduces the number of `long double` multiplications necessary.
* Reading of floating point numbers in `scanf` was reworked, fixing some bugs and improving performance (e.g. previously the conversion was performed twice, with one result being discarded).
* Fix buffer over-read bug in `scanf` for `%c` format.

## Motivation and Context
Fix https://github.com/phoenix-rtos/phoenix-rtos-project/issues/634, https://github.com/phoenix-rtos/phoenix-rtos-project/issues/657, https://github.com/phoenix-rtos/phoenix-rtos-project/issues/703, https://github.com/phoenix-rtos/phoenix-rtos-project/issues/704, https://github.com/phoenix-rtos/phoenix-rtos-project/issues/652, https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1116

Partially fix https://github.com/phoenix-rtos/phoenix-rtos-project/issues/681 (supports width specifier for "%f" up to 31)

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/stdlib/strtod.c
+++ b/stdlib/strtod.c
@@ -6,7 +6,7 @@
  * strtod.c
  *
  * Copyright 2018, 2022 Phoenix Systems
- * Author: Kamil Amanowicz, Damian Loewnau, Dawid Szpejna
+ * Author: Kamil Amanowicz, Damian Loewnau, Dawid Szpejna, Jacek Maksymowicz
  *
  * This file is part of Phoenix-RTOS.
  *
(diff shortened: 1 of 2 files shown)


stdlib/strtod.c
```diff
@@ -15,184 +15,539 @@
...
+	if (sign == 0) {
+		if (endptr != NULL) {
+			*endptr = (char *)str_wsin;
+		}
+
+		return result;
 	}
 
-	if (!is_number) {
+	bool is_hex;
```
Darchiv, line 485: Move variable declarations to the start of the scope?

stdlib/strtod.c
```diff
@@ -15,184 +15,539 @@
...
 	}
-	else if (isinf(result)) {
+
+	str = floatparse_number(str_zerox, is_hex, type, &result);
+
+	bool range_error;
+	switch (type) {
+		case TYPE_ID_FLT: {
+			float cast = (float)result;
+			range_error = isinf(cast) || ((result != 0) && (cast == 0));
```
Darchiv, line 502: nitpick:
```suggestion
			range_error = isinf(cast) || ((result != 0.0l) && (cast == 0.0f));
```
Applies to everywhere. Makes variables more "self-documenting".

stdlib/strtod.c
```diff
@@ -15,184 +15,539 @@
...
+		str++;
+	}
 
-			if (*str == '+' || *str == '-') {
-				str--;
+	*out = result;
+	return str;
+}
+
+static long double mul_exp10(long double in, int64_t exp)
```
Darchiv, line 257: Do you assume the value of `FLT_ROUNDS`? Which rounding direction do you use? By default (in our toolchain) `FLT_ROUNDS` is `FE_TONEAREST`. If that's what you implement, then maybe you could add some `#warning` if `FLT_ROUNDS != FE_TONEAREST`. Of course, implementing other rounding modes is unnecessary, but this check will avoid confusion if we ever use another toolchain and `FLT_ROUNDS` changes.

stdlib/strtod.c
```diff
@@ -15,184 +15,539 @@
...
+#define TYPE_ID_LDBL     2
+#define LDBL_MANT_10_DIG 34
+#else
+#error "Unknown floating point implementation"
+#endif
+
+typedef union {
+	struct {
+#if __FLOAT_WORD_ORDER__ == __ORDER_LITTLE_ENDIAN__
+		uint64_t mantisa : 52;
```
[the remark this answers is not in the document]
jmaksymowicz: I converted the bitfield to a bitwise operation. `__STDC_IEC_559__` currently isn't implemented by the toolchain.

stdlib/strtod.c
```diff
@@ -15,184 +15,539 @@
...
 #include <ctype.h>
 #include <math.h>
 #include <errno.h>
+#include <limits.h>
+
+
+#define TYPE_ID_FLT 0
+#define TYPE_ID_DBL 1
+
+#if (__LDBL_MANT_DIG__ == __DBL_MANT_DIG__) && (__LDBL_MAX_EXP__ == __DBL_MAX_EXP__)
```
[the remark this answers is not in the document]
jmaksymowicz: Done

stdlib/strtod.c
jmaksymowicz: Done

stdlib/strtod.c
jmaksymowicz: This code was tested with `FLT_ROUNDS == FE_TONEAREST`. I added a warning in case this is changed. So far setting the rounding mode is not implemented in libphoenix.

stdlib/strtod.c
jmaksymowicz: I can do it, but I prefer to have variable declarations close to the point where they are first needed.

stdlib/strtod.c
```diff
@@ -15,184 +15,539 @@
...
-		*endptr = (char *)str;
+	*out = result;
+	return str;
+}
+
+
+static long double strto_any(const char *__restrict str, char **__restrict endptr, uint8_t type)
+{
+	const char *decimal_point_str = localeconv()->decimal_point;
+	int decimal_point = decimal_point_str[0];
```
Darchiv, line 462: Maybe `decimal_point` should be a `char` instead of `int`? It is compared against chars in this source file. Also, functions work on char arrays, so a multi-character separator will be treated like a string and not a code-point.

stdlib/strtod.c
```diff
@@ -15,184 +15,539 @@
...
+	if ((decimal_point == 0) || (decimal_point_str[1] != 0)) {
```
Darchiv, line 463: If the decimal point is not defined in the current locale, then `.` shall be used as a default (according to POSIX), instead of aborting the conversion.

stdlib/strtod.c
Darchiv: Ok, can be left as-is.

stdlib/strtod.c
jmaksymowicz: This check is not meant for when the decimal point is not defined, but rather for when it is invalid. For example, glibc checks for this case using an assertion: https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/stdlib/strtod_l.c#L592

stdlib/strtod.c
jmaksymowicz: OK, done.

Darchiv (approved):

Darchiv (approved):

PR #376, opened 2024-08-21
stdio/file: fix numerous issues in the fread/fwrite functions

ziemleszcz:
## Types of changes

## How Has This Been Tested?

- [x] New test added: (https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/369).

## Checklist:

## Special treatment

### DIFF
--- a/stdio/file.c
+++ b/stdio/file.c
@@ -160,20 +160,24 @@ static void file_free(FILE *file)
 }
 
 
-int fclose(FILE *file)
+int fclose(FILE *stream)
 {
 	int err;
 
-	if (file == NULL) {
-		return -EINVAL;
+	if (stream == NULL) {
+		errno = EBADF;
+		return EOF;
 	}
 
-	fflush(file);
-	while ((err = __safe_close(file->fd)) < 0 && errno == EINTR) {
+	err = fflush(stream);
+	if (err == EOF) {
+		return EOF;
 	}
-	file_free(file);
 
-	return err;
+	err = __safe_close(stream->fd);
+	file_free(stream);
+
+	return (err < 0) ? EOF : 0;
 }
 
 
@@ -207,7 +211,6 @@ FILE *fopen(const char *filename, const char *mode)
 	f->bufsz = BUFSIZ;
 	f->fd = fd;
 	f->mode = m;
-	fflush(f);
 
 	mutexLock(file_common.lock);
 	LIST_ADD(&file_common.list, f);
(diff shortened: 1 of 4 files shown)


unistd/file.c
```diff
@@ -101,18 +101,26 @@ ssize_t __safe_read(int fd, void *buf, size_t size)
 			continue;
 		}
 		else {
-			break;
+			return -1;
```
Darchiv, line 104: > Also fixed error handling in blocking safe read to always return -1 on error.

How is this different? The `else` branch is taken when `rlen < 0`, so `if (rlen < 0)` below is still executed.

ziemleszcz: You're right  - disregard this comment.

stdio/file.c
```diff
@@ -311,98 +320,186 @@ static int full_read(int fd, void *ptr, size_t size)
...
+				ret = -1;
+			}
+			else {
+				stream->bufpos = 0;
+			}
+		}
+	}
+	else {
+		if (stream->bufpos != stream->bufeof) {
+			off = lseek(stream->fd, (off_t)stream->bufpos - stream->bufeof, SEEK_CUR);
```
Darchiv, line 366: Should this be as-is or the other way around?

```suggestion
			off = lseek(stream->fd, (off_t)stream->bufeof - stream->bufpos, SEEK_CUR);
```

stdio/file.c
```diff
@@ -416,67 +513,167 @@ size_t fread(void *ptr, size_t size, size_t nmemb, FILE *stream)
...
-	/* discard reading buffer */
-	if (!(stream->flags & F_WRITING)) {
-		fflush_unlocked(stream);
+	/* flush the read buffer if currently reading */
+	if ((stream->flags & F_WRITING) == 0) {
+		if (__fflush_one(stream) < 0) {
+			return 0;
+		}
+		if (stream->bufpos != stream->bufeof) {
+			/* abort if the read buffer cannot be flushed (non-seekable stream) */
```
Darchiv, line 595: Shouldn't `errno` be set then? (And maybe the error flag?)

stdio/file.c
ziemleszcz: This is correct, we are seeking backwards: `-(bufeof - bufpos) = bufpos - bufeof`.

stdio/file.c
ziemleszcz: It's implementation-specific behavior. In this case, I'm emulating glibc.

Darchiv: Ugh, case similar to `errno` in `malloc(0)`, so we probably can't solve this reasonably. Ack.

Darchiv (dismissed):

Darchiv (approved):

PR #388, opened 2024-10-10
sys/select: fix errnos

adamgreloch:
select() should fail with -1 and set errno to:
* EBADF when one or more of the fdsets specified an invalid fd
* EINVAL if nfds is incorrect and/or if timeout makes no sense (i.e. tv_usec is outside of [0, 999999])

JIRA: RTOS-945

## Types of changes

## How Has This Been Tested?

- [X] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/380.

## Checklist:

## Special treatment

### DIFF
--- a/sys/select.c
+++ b/sys/select.c
@@ -46,14 +46,29 @@ int select(int nfds, fd_set *rd, fd_set *wr, fd_set *ex, struct timeval *to)
 	struct pollfd *pfd;
 	time_t timeout;
 	size_t i, n;
-	int err;
+	int rv;
+
+	if ((nfds < 0) || (nfds > FD_SETSIZE)) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	if ((to != NULL) && ((to->tv_usec < 0) || (to->tv_usec > 999999))) {
+		errno = EINVAL;
+		return -1;
+	}
 
 	for (n = i = 0; i < nfds; ++i)
 		if (SFD_ISSET(i, rd) || SFD_ISSET(i, wr) || SFD_ISSET(i, ex))
 			++n;
 
-	if (!n)
-		return usleep(poll_timeout(to));
+	rv = poll_timeout(to);
+	timeout = rv < 0 ? 0 : (time_t)rv;
+
+	if (n == 0) {
+		rv = usleep(timeout);
+		return rv < 0 ? -1 : 0;
+	}
 
 	pfd = calloc(n, sizeof(*pfd));
 	if (!pfd) {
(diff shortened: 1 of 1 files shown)


sys/select.c
```diff
@@ -48,12 +48,26 @@ int select(int nfds, fd_set *rd, fd_set *wr, fd_set *ex, struct timeval *to)
...
+		return -1;
+	}
+
 	for (n = i = 0; i < nfds; ++i)
 		if (SFD_ISSET(i, rd) || SFD_ISSET(i, wr) || SFD_ISSET(i, ex))
 			++n;
 
-	if (!n)
-		return usleep(poll_timeout(to));
+	timeout = (time_t)poll_timeout(to);
```
agkaminski, line 65: `to == NULL` is accepted, but it will cause bad stuff below - `timeout = -1` then and `usleep(-1) `waits for few thousand years. It's not super bad right now, as `time_t` is unsigned, but it might become signed in the future and I guess negative time would mean no wait at all. I guess we don't want a timeout then, but I guess we should rather set `timeout` to zero if `to` is `NULL`

sys/select.c
```diff
@@ -76,22 +90,20 @@ int select(int nfds, fd_set *rd, fd_set *wr, fd_set *ex, struct timeval *to)
...
+			break;
+		}
+	}
+
 	if (err < 0) {
 		if (pfd)
 			free(pfd);
-		return err;
+		SET_ERRNO(err);
+		return -1;
```
agkaminski, line 106:
```suggestion
		return SET_ERRNO(err);
```

sys/select.c
```diff
@@ -48,12 +48,26 @@ int select(int nfds, fd_set *rd, fd_set *wr, fd_set *ex, struct timeval *to)
 	size_t i, n;
 	int err;
 
+	if (nfds < 0 || nfds > FD_SETSIZE) {
```
agkaminski, line 51: MISRA
```suggestion
	if ((nfds < 0) || (nfds > FD_SETSIZE)) {
```

sys/select.c
```diff
@@ -48,12 +48,26 @@ int select(int nfds, fd_set *rd, fd_set *wr, fd_set *ex, struct timeval *to)
...
+		errno = EINVAL;
+		return -1;
+	}
+
+	if ((to != NULL) && (to->tv_usec < 0 || to->tv_usec > 999999)) {
```
agkaminski, line 56: MISRA
```suggestion
	if ((to != NULL) && ((to->tv_usec < 0) || (to->tv_usec > 999999))) {
```

sys/select.c
agkaminski: Or even better, we should check retval for `< 0` before assigning the `timeout`

adamgreloch: Well, it depends whether we classify `timeout = -1` and `usleep(-1)`  as bad. I initially thought that current `usleep(-1)` is a bug, as using select just to emulate `usleep` by passing empty sets and some timeout is okay, but passing NULL timeout just to hang approx. forever seemed pointless to me. Then I checked how glibc handles `select(_, NULL, NULL, NULL, NULL)` and it hangs just like that. Actually, POSIX implies that a NULL timeout should cause select to block indefinitely waiting for any fd to become ready. It doesn't restrict the sets to be non-empty in this case, so the select is allowed to block forever, so the glibc behavior makes sense (although IMHO still seems a bit pointless)

agkaminski: Yeah, I agree that behaviour is ok, but our sleep expects 0 for such case, not negative (in reality really big, as time_t is unsigned) value

agkaminski: What I mean is something along the lines of:
```
int ret = poll_timeout(to);
timeout = ret < 0 ? 0 : (time_t)ret;
```

agkaminski: This is not ideal still, because if tv_sec >= INT_MAX/1000 then we will sleep indefinitely. Not a really use-case though

PR #393, opened 2024-11-08
math: improve POSIX compliance

adamdebek:
JIRA: RTOS-974

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (list targets here).

## Checklist:

## Special treatment

### DIFF
--- a/math/common.c
+++ b/math/common.c
@@ -14,6 +14,9 @@
  */
 
 #include <stdint.h>
+#include <errno.h>
+#include <math.h>
+#include <float.h>
 #include "common.h"
 
 
@@ -131,6 +134,11 @@ double quickPow(double x, int e)
 			x *= x;
 			eabs >>= 1;
 		}
+	}
+
+	/* ia32 use 80-bit extended precision registers so isinf() may not be true */
+	if ((isinf(res) != 0) || (fabs(res) > DBL_MAX)) {
+		errno = ERANGE;
 	}
 
 	return res;
--- a/math/common.c
+++ b/math/common.c
@@ -21,25 +21,26 @@ void normalizeSub(double *x, int *exp)
 {
 	conv_t *conv = (conv_t *)x;
 
-	if (conv->i.mantisa == 0)
+	if (conv->i.mantisa == 0) {
 		return;
+	}
 
-	while (!(conv->i.mantisa & 0xffffff0000000LL)) {
+	while ((conv->i.mantisa & 0xffffff0000000LL) == 0) {
 		conv->i.mantisa <<= 24;
 		*exp -= 24;
 	}
 
-	while (!(conv->i.mantisa & 0xff00000000000LL)) {
+	while ((conv->i.mantisa & 0xff00000000000LL) == 0) {
 		conv->i.mantisa <<= 8;
 		*exp -= 8;
 	}
 
-	while (!(conv->i.mantisa & 0xf000000000000LL)) {
+	while ((conv->i.mantisa & 0xf000000000000LL) == 0) {
 		conv->i.mantisa <<= 4;
 		*exp -= 4;
 	}
 
-	while (!(conv->i.mantisa & 0x8000000000000LL)) {
+	while ((conv->i.mantisa & 0x8000000000000LL) == 0) {
 		conv->i.mantisa <<= 1;
 		*exp -= 1;
 	}
@@ -89,8 +90,9 @@ void createSub(double *x, int exp)
 		++exp;
 	}
 
-	if (conv->i.mantisa == 0)
+	if (conv->i.mantisa == 0) {
 		*x = 0.0;
+	}
 }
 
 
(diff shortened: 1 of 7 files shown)


math/common.c
```diff
@@ -99,34 +104,42 @@ double quickPow(double x, int e)
...
 				res *= x;
+			}
 			x *= x;
 			eabs >>= 1;
 		}
 	}
 
+	if (res > DBL_MAX) {
+		errno = ERANGE;
+	}
```
agkaminski, line 141: This check is always false, double can never be greater than DBL_MAX.

Result should assume +-INF on overflow, according to the IEE754

math/hyper.c
```diff
@@ -14,29 +14,61 @@
...
+		return NAN;
+	}
+
+	if ((x == INFINITY) || (x == -INFINITY)) {
 		return INFINITY;
+	}
 
-	return (exp(x) + exp(-x)) / 2;
+	/* Make sure exp(x) is not infinity */
+	if (x < 709.78) {
```
agkaminski, line 35: It's 709.79 in exp.c:198, which value is correct one?

Please mind that 709.78 < 709.7899999 < 709.79

math/hyper.c
```diff
@@ -14,29 +14,61 @@
...
+		return NAN;
+	}
+
+	if ((x == INFINITY) || (x == -INFINITY)) {
 		return x;
+	}
 
-	return (exp(x) - exp(-x)) / 2;
+	/* Make sure exp(x) is not infinity */
+	if (x < 709.78) {
```
agkaminski, line 59: Same stuff as in cosh

math/power.c
```diff
@@ -27,38 +28,48 @@ double pow(double base, double exponent)
...
 		return quickPow(base, (int)exponent);
+	}
 
 	exponent *= log(base);
-	res = s * exp(exponent);
+	res = (double)s * exp(exponent);
+
+	if (res > DBL_MAX) {
+		errno = ERANGE;
+	}
```
agkaminski, line 72: Always false, check for +-INF instead

math/common.c
adamdebek: If `res` is `INFINITY` this check is true and I do this because on ia32 all floating point operations are done in 80-bit expanded precision registers, so checking if res is `INFINITY` using `isinf()` can be false even though after register spill its value will resolve to `INFINITY`. I agree I should add check for `-INFINITY`

math/hyper.c
adamdebek: The value of the argument at which exp(x) evaluates to INFINITY is approximately 709.7825. However, this is not defined as a constant in any standard, so I use a safe margin from this value.

math/common.c
Darchiv: @adamdebek Do you plan on adding `-INFINITY` check?

So `isinf()` can still be false even though it uses `__builtin_isinf()` underneath? Does `res > DBL_MAX` work in all cases, on all architectures we support? Maybe `res > DBL_MAX || isinf(res)` will be a more robust approach (or not)?

math/exp.c
```diff
@@ -22,9 +22,13 @@
 double frexp(double x, int *exp)
 {
 	conv_t *conv = (conv_t *)&x;
-
 	*exp = 0;
 
+	if (isnan(x) != 0) {
+		errno = EDOM;
```
Darchiv, line 28: Does any standard/specification dictate that `errno` should be set?

Applies to other functions.

math/exp.c
```diff
@@ -22,9 +22,13 @@
...
+		return NAN;
```
Darchiv, line 29: In case `isnan(x)` or `isinf(x)` hold, `*exp` can be left unmodified (its value is unspecified). Maybe it's worth to move these above `*exp = 0`, if no legacy test suites / applications rely on the value (glibc?).

math/exp.c
```diff
@@ -22,9 +22,13 @@
...
+	}
+
 	if ((x == 0.0) || (x == -0.0)) {
```
Darchiv, line 32: There are checks for ±0 and NAN - what about ±INF?

Applies to other functions.

math/exp.c
```diff
@@ -45,6 +49,11 @@ double ldexp(double x, int exp)
 	conv_t *conv = (conv_t *)&x;
 	int exponent = 0;
 
+	if (isnan(x) != 0) {
+		errno = EDOM;
```
Darchiv, line 53: Does any standard/specification dictate that `errno` should be set to `EDOM` in this function?

math/exp.c
```diff
@@ -79,7 +88,7 @@ double log(double x)
 	conv_t *conv = (conv_t *)&tmp;
 	int exp = 0, i;
 
-	if (x < 0.0) {
+	if ((isnan(x) != 0) || (x < 0.0)) {
 		errno = EDOM;
```
Darchiv, line 92: Should `errno` be set when `isnan(x)`?

What about -INF?

math/common.c
adamdebek: Yes, still can be as I said above.

I agree with additional `isinf()`

math/exp.c
adamdebek: Good catch, neither IEC 60559 nor ISO/IEC 9899:1999 mandates it.

math/hyper.c
Darchiv: Probably best to include this information in the comment in code, so as to avoid unnecessary confusion and magic values.

Also applies to `exp()` which will benefit from a comment, too.

Darchiv (approved):
LGTM, but left a nitpick or two

math/exp.c
```diff
@@ -23,9 +23,18 @@ double frexp(double x, int *exp)
...
+	}
+
+	if (isinf(x) != 0) {
+		return x;
+	}
+
 	*exp = 0;
 
 	if ((x == 0.0) || (x == -0.0)) {
+		*exp = 0;
```
Darchiv, line 37: As I have written, `*exp = 0` can be moved before this `if` - now, if `*exp = 0` is also inside `if`, `*exp = 0` is executed twice if the condition holds true. It should be removed from this line.

Darchiv (approved):

PR #409, opened 2025-03-20
arm: adapt to being built as multilib

badochov:
JIRA: RTOS-1028

## Description
Multilib support is very basic and the multilibs are virtually useless as the syscalls are not implemented. To add proper support the ASM code should be chosen for each version based on GCC flags not target.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/arch/arm/Makefile
+++ b/arch/arm/Makefile
@@ -6,17 +6,34 @@
 # %LICENSE%
 #
 
-ifneq (,$(findstring v7a,$(TARGET_SUFF)))
+# Obtain architecture based on ARM ACLE 2.0
+# https://developer.arm.com/documentation/dui0774/l/Other-Compiler-specific-Features/Predefined-macros
+arm_version := $(word 3, $(shell echo | arm-phoenix-gcc $(CFLAGS) -E -dM - | grep -E "[[:space:]]__ARM_ARCH[[:space:]]"))
+arm_profile_num := $(word 3, $(shell echo | arm-phoenix-gcc $(CFLAGS) -E -dM - | grep -E "[[:space:]]__ARM_ARCH_PROFILE[[:space:]]"))
+# ASCII conversion
+arm_profile := $(subst 65,A,$(subst 77,M,$(subst 82,R,$(arm_profile_num))))
+arm_arch := $(arm_version)$(arm_profile)
+
+arm_use_common := y
+ifneq (,$(findstring 7A,$(arm_arch)))
   include arch/arm/v7a/Makefile
-else ifneq (,$(findstring v7m,$(TARGET_SUFF)))
+else ifneq (,$(findstring 7M,$(arm_arch)))
   include arch/arm/v7m/Makefile
-else ifneq (,$(findstring v8m,$(TARGET_SUFF)))
+else ifneq (,$(findstring 8M,$(arm_arch)))
   include arch/arm/v8m/Makefile
-else ifneq (,$(findstring v8r,$(TARGET_SUFF)))
+else ifneq (,$(findstring 8R,$(arm_arch)))
   include arch/arm/v8r/Makefile
+else ifeq ($(LIBPHOENIX_MULTILIB),y)
+  include arch/arm/multilib/Makefile
+  arm_use_common := n
 else
   $(error Unsupported TARGET)
 endif
 
+# FIXME: multilib supports more targets that functions in those files do,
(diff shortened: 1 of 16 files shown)


arch/arm/Makefile
```diff
@@ -14,9 +14,19 @@ else ifneq (,$(findstring v8m,$(TARGET_SUFF)))
   include arch/arm/v8m/Makefile
 else ifneq (,$(findstring v8r,$(TARGET_SUFF)))
   include arch/arm/v8r/Makefile
+else ifneq (,$(findstring v8r,$(TARGET_SUFF)))
+  include arch/arm/v8r/Makefile
+else ifeq ($(TARGET_SUBFAMILY),multilib)
+  # TODO: Try to include correct Makefile if possible in case of a multilib.
```
nalajcie, line 20: I'm a little bit afraid that we will build non-fuctioning subvariants for supported targets while building the toolchain - eg. which implementation of `setjmp` will be used for `armv7m7-imx117x` with FDPIC enabled? (it's not a standard phoenix target, so it will be built only by multilib option?

Maybe we should try to inspect CFLAGS (`march/mcpu` options)? Or try using output of `$(CC) $(CFALGS)  -print-multi-directory`?

arch/arm/multilib/string.c
```diff
@@ -0,0 +1,124 @@
...
+	}
+	dest[i] = '\0';
+	return dest;
+}
+
+
+char *strncpy(char *dest, const char *src, size_t n)
+{
+	size_t i = 0;
+	while (((i < n) && (src[i] != '\0'))) {
```
nalajcie, line 100: this  is not a valid implementation of `strncpy` (zero-fill all of the remaining place, don't add `'\0' if it doesn't fit)

arch/arm/multilib/string.c
```diff
@@ -0,0 +1,124 @@
...
+		int diff = ((int)(((const unsigned char *)s1)[i])) - ((int)(((const unsigned char *)s2)[i]));
+		if (diff != 0) {
+			return diff;
+		}
+	}
+	return 0;
+}
+
+
+int strncmp(const char *s1, const char *s2, size_t count)
```
nalajcie, line 73: invalid - `strncmp("a", "ab", 2)` returns 0 (should return negative value)

arch/arm/multilib/string.c
```diff
@@ -0,0 +1,124 @@
...
+{
+	size_t len = 0;
+	while ((s[len] != '\0') && (len < maxlen)) {
+		len++;
+	}
+	return len;
+}
+
+
+int strcmp(const char *s1, const char *s2)
```
nalajcie, line 61: invalid - `strcmp("a", "ab")` returns 0 (should return negative value)

nalajcie (changes requested):

arch/arm/multilib/string.c
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return 0;
+}
+
+
+char *strcpy(char *dest, const char *src)
+{
+	size_t i = 0;
+	while ((src[i] != '\0')) {
+		((char *)dest)[i] = ((const char *)src)[i];
```
nalajcie, line 89: why do we need explicit casts here?

arch/arm/Makefile
badochov: That would be the go to solution, however, I don't think I have time to fix it

badochov: Note, however, that when we create such a configuration and build simply libphoenix during building the target then correct implementation will be used, as install is make target is called

arch/arm/multilib/syscalls.S
```diff
@@ -0,0 +1,41 @@
...
+#include <phoenix/syscalls.h>
+
+.text
+
+
+#define SYSCALLDEF(sym, sn) \
+.globl sym; \
+.type sym, %function; \
+sym: \
+	bl abort; \
```
badochov, line 26: `bl` as `b` may have to short range

arch/arm/multilib/signal.S
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+#define __ASSEMBLY__
+
+.text
+
+.globl _signal_trampoline
+.type _signal_trampoline, %function
+_signal_trampoline:
+	bl abort
```
badochov, line 23: `bl` as `b` may have to short range

PR #412, opened 2025-05-06
Fixes to allow enabling -Wundef

anglov:
## Description

There are some useful warning flags, which are enabled in internal projects, like `-Wundef` and `-Wshadow`.

This commit allow libphoenix to be build with -Wundef without warnings

## Motivation and Context

To enable `-Wundef` flag in the future

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: run compilation and some build-time checks on armv7m4-stm32l4x6-nucleo armv7m7-imxrt117x-evk sparcv8leon-generic-qemu riscv64-noelv-fpga riscv64-generic-qemu armv8r52-mps3an536-qemu armv8m33-mcxn94x-frdm aarch64a53-zynqmp-qemu armv7a9-zynq7000-qemu armv7a7-imx6ull-evk ia32-generic-qemu. Not run any of target after changes

## Checklist:

## Special treatment

### DIFF
--- a/include/arch/armv7a/arch.h
+++ b/include/arch/armv7a/arch.h
@@ -31,8 +31,8 @@
 #define __MEMMOVE
 
 
-#if __ARM_PCS_VFP || (__VFP_FP__ && !__SOFTFP__)
-#if __ARM_FP & 8
+#if defined(__ARM_PCS_VFP) || (defined(__VFP_FP__) && !defined(__SOFTFP__))
+#if defined(__ARM_FP) && (__ARM_FP & 8) != 0
 #define __IEEE754_SQRT
 #define __ieee754_sqrt(x) ({ double a = (x); __asm__ volatile ("vsqrt.f64 %P0, %P1" : "=w"(a) : "w"(a)); a; })
 #endif
--- a/include/arch/armv7m/arch.h
+++ b/include/arch/armv7m/arch.h
@@ -31,8 +31,8 @@
 #define __MEMMOVE
 
 
-#if __ARM_PCS_VFP || (__VFP_FP__ && !__SOFTFP__)
-#if __ARM_FP & 8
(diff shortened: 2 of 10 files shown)


agkaminski:
> @agkaminski could you verify if soft float sqrt macros is evaluated correctly?

Seems to be working just fine

nalajcie (approved):
looks sane, did not test it :P

include/arch/armv7a/arch.h
```diff
@@ -31,8 +31,8 @@
 #define __MEMMOVE
 
 
-#if __ARM_PCS_VFP || (__VFP_FP__ && !__SOFTFP__)
-#if __ARM_FP & 8
+#if defined(__ARM_PCS_VFP) || (defined(__VFP_FP__) && !defined(__SOFTFP__))
```
agkaminski, line 34: Not really the same as was, what if e.g. `__ARM_PCS_VFP` is set to zero? Not sure if this might be the case, from what I see the use case seen on github is to do it the same way you did it.

agkaminski (approved):
No remarks, just one rant ;)

include/arch/armv7a/arch.h
anglov: Looking in practice
```
$ echo | arm-phoenix-gcc -dM -E  -mcpu=cortex-m4 -mfloat-abi=soft - | grep __ARM_PCS_VFP
$ echo | arm-phoenix-gcc -dM -E -mcpu=cortex-m7 -mfloat-abi=hard -mfpu=fpv5-d16 - | grep __ARM_PCS_VFP
#define __ARM_PCS_VFP 1
```

According to https://developer.arm.com/documentation/dui0774/g/chr1383660321827, I believe that this variable can be set only to 1, when specific condition is met (target is 32-bit and PCS is compatible with floating point)

I believe that it's true for (almost?) all compiler feature defines – it's not defined when not available, defined to 1 when available

PR #420, opened 2025-07-14
!signal: move signal handling to kernel

etiaro:
## Description

* remove default handlers, signal_common and _signals_init from libphoenix
* switch from signalHandle syscall to signalAction
* add handler address argument to signal_trampoline
* remove phx-posix signal numbers translation
* remove signalPost syscall as duplicate to sys_tkill

## Motivation and Context

JIRA: RTOS-736

Depends-On: phoenix-rtos-tests:etiaro/signals

## Types of changes

## How Has This Been Tested?

- [x] New test added: phoenix-rtos/phoenix-rtos-tests#410
- [ ] Tested by hand on: 

## Checklist:

## Special treatment

 - [!syscalls: move signal handling to kernel](https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/670)

### DIFF
--- a/arch/aarch64/signal.S
+++ b/arch/aarch64/signal.S
@@ -20,12 +20,13 @@
 .globl _signal_trampoline
 .type _signal_trampoline, %function
 _signal_trampoline:
-	/* Signal number, oldmask - keep stack aligned to 16 */
-	ldp x0, x19, [sp], #0x10
-	bl _signal_handler
+	/* Signal number, handler address - stack aligned to 16 */
+	ldp x0, x1, [sp], #0x10
+	blr x1
 
-	mov x0, x19             /* move oldmask to x0 */
-	ldp x1, x2, [sp], #0x10 /* cpu_context_t *, pc */
-	ldp x3, x4, [sp], #0x10 /* sp, psr */
+	ldp x0, x1, [sp], #0x10 /* oldmask, cpu_context_t * */
+	ldp x2, x3, [sp], #0x10 /* pc, sp, psr */
+	ldr x4, [sp], #0x10 /* psr, skip padding at the end */
+
 	bl sigreturn
 .size _signal_trampoline, .-_signal_trampoline
--- a/arch/ia32/signal.S
+++ b/arch/ia32/signal.S
@@ -20,8 +20,9 @@
 .globl _signal_trampoline
 .type _signal_trampoline, %function
 _signal_trampoline:
+	pop %ebx /* handler */
 	/* Signal number on stack */
-	call _signal_handler
+	call %ebx
 
 	addl $4, %esp
 	/* oldmask is now on top of the stack */
--- a/include/signal.h
+++ b/include/signal.h
@@ -17,8 +17,7 @@
 #define _LIBPHOENIX_SIGNAL_H_
 
 
-#include <sys/types.h>
-#include <time.h>
+#include <phoenix/signal.h>
 
 
 #ifdef __cplusplus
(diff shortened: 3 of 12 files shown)


arch/ia32/signal.S
```diff
@@ -20,8 +20,9 @@
 .globl _signal_trampoline
 .type _signal_trampoline, %function
 _signal_trampoline:
+	pop %ebx /* handler */
 	/* Signal number on stack */
-	call _signal_handler
+	call %ebx
```
adamgreloch:
```
  /home/runner/work/libphoenix/libphoenix/arch/ia32/signal.S: Assembler messages:
  /home/runner/work/libphoenix/libphoenix/arch/ia32/signal.S:25: Warning: indirect call without `*'
```

arch/aarch64/signal.S
```diff
@@ -20,12 +20,13 @@
...
-	/* Signal number, oldmask - keep stack aligned to 16 */
-	ldp x0, x19, [sp], #0x10
-	bl _signal_handler
+	/* Signal number, handler address - stack aligned to 16 */
+	ldp x0, x1, [sp], #0x10
+	blr x1
+
+	ldp x0, x1, [sp], #0x10 /* oldmask, cpu_context_t * */
+	ldp x2, x3, [sp], #0x10 /* pc, sp, psr */
+	ldr x4, [sp], #0x10 /* psr, skip padding at the end */
```
jmaksymowicz, line 30: It seems that `sigreturn` only takes 2 arguments - `oldmask` and `ctx`, so loading of those extra arguments is no longer necessary. Also, the second comment is now misleading (implies we load 3 distinct values instead of 2).
```suggestion
	ldp x0, x1, [sp], #0x30 /* oldmask, cpu_context_t *, skip other arguments */
```

Perhaps, since we're changing the contract between kernel and `libphoenix` anyway, we can just not pass these extra arguments in `hal_cpuPushSignal`.

etiaro: It seems that `hal_cpuSigreturn` does use these arguments (to revert them inside the copied context, where they are replaced with their versions for signal-handling)

https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/3f86c550b7ef62626358f57d32baef389a7184ee/hal/aarch64/cpu.c#L110-L116

jmaksymowicz: Oof.

It would be nice to have some documentation that `sigreturn` syscall takes a platform-dependent number of additional arguments. Maybe just a comment in `syscalls_sigreturn`?
```c
GETFROMSTACK(ustack, cpu_context_t *, ctx, 1U);
/* Note: `hal_cpuSigreturn` may take additional arguments from ustack */
```

etiaro: Yeah, probably worth adding.
Also, GETFROMSTACK macros appear in some `posix.c` functions, might be useful to note that too.

arch/ia32/signal.S
```diff
@@ -20,8 +20,9 @@
 .globl _signal_trampoline
 .type _signal_trampoline, %function
 _signal_trampoline:
+	pop %ebx /* handler */
 	/* Signal number on stack */
-	call _signal_handler
+	call *%ebx
```
ziemleszcz, line 27: The pop above breaks %esp 16-byte alignment (see  comment in `_start` and STACK_ALIGN in the kernel). I386 ABI requires 4 byte alignment so looks like currently it's not a problem but maybe this alignment was set for some other reason.

etiaro: Yeah, modern compilers probably don't rely on that, but it's still a living requirement and might make a mess in the future.
I'll push the signal number again to align it.

signal/signal.c
```diff
@@ -174,173 +58,60 @@ int killpg(pid_t pgrp, int sig)
...
 
 
 void (*signal(int signum, void (*handler)(int)))(int)
 {
-	sighandler_t t;
-	unsigned int oldmask;
+	struct sigaction act;
+	act.sa_handler = handler;
+	act.sa_flags = 0;
+	sigemptyset(&act.sa_mask);
```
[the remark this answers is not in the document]
etiaro: Not sure if that's really necessary, but can be applied easily and won't break anything so I'll do it.

include/signal.h
```diff
@@ -29,85 +27,6 @@ extern "C" {
 
 typedef void (*sighandler_t)(int);
```
ziemleszcz: Already defined in `phoenix/signal.h`.

ziemleszcz (approved):

PR #433, opened 2025-09-01
posix: implement missing directory functions

R4ken:
Implement missing POSIX functions (fdopendir,seekdir, telldir) working on DIR structs

JIRA: RTOS-1088

## Description
Implemented fdopendir,seekdir and telldir functions. 

## Motivation and Context
Missing POSIX functions useful in porting applications

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic-qemu, armv7a9-zynq7000-qemu)

## Checklist:

## Special treatment

### DIFF
--- a/include/dirent.h
+++ b/include/dirent.h
@@ -48,10 +48,20 @@ struct dirent {
 };
 
 
-extern struct dirent *readdir(DIR *s);
+/* These functions are not thread-safe */
+extern struct dirent *readdir(DIR *dirp);
 
 
 extern DIR *opendir(const char *dirname);
+
+
+extern DIR *fdopendir(int fd);
+
+
+extern void seekdir(DIR *dirp, long loc);
+
+
+extern long telldir(DIR *dirp);
 
 
 extern void rewinddir(DIR *dirp);
--- a/include/stdio.h
+++ b/include/stdio.h
@@ -70,7 +70,8 @@ typedef struct _FILE {
 
 typedef struct _DIR {
 	oid_t oid;
-	size_t pos;
+	int fd;
+	off_t pos;
 	struct dirent *dirent;
 } DIR;
 
--- a/unistd/dir.c
+++ b/unistd/dir.c
@@ -26,6 +26,7 @@
 #include <sys/stat.h>
 #include <sys/file.h>
 #include <posix/utils.h>
+#include <fcntl.h>
 
 
 static struct {
@@ -347,101 +348,191 @@ char *realpath(const char *path, char *resolved_path)
 }
 
 
-struct dirent *readdir(DIR *s)
+struct dirent *readdir(DIR *dirp)
 {
-	if (s->dirent == NULL) {
-		if ((s->dirent = calloc(1, sizeof(struct dirent) + NAME_MAX + 1)) == NULL)
-			return NULL;
+	if (dirp == NULL) {
+		errno = EBADF;
+		return NULL;
+	}
+
+	if (dirp->dirent == NULL) {
+		dirp->dirent = calloc(1, sizeof(struct dirent) + NAME_MAX + 1);
+		if (dirp->dirent == NULL) {
+			return NULL; /* errno set by calloc() */
+		}
 	}
 
 	msg_t msg = {
 		.type = mtReaddir,
-		.oid = s->oid,
-		.i.readdir.offs = s->pos,
-		.o.data = s->dirent,
+		.oid = dirp->oid,
+		.i.readdir.offs = dirp->pos,
+		.o.data = dirp->dirent,
 		.o.size = sizeof(struct dirent) + NAME_MAX + 1
 	};
 
-	if (msgSend(s->oid.port, &msg) < 0) {
-		free(s->dirent);
-		s->dirent = NULL;
-		return NULL; /* EIO */
-	}
-
-	if (msg.o.err < 0) {
-		free(s->dirent);
-		s->dirent = NULL;
+	if (msgSend(dirp->oid.port, &msg) < 0) {
+		free(dirp->dirent);
+		dirp->dirent = NULL;
+		errno = EIO;
 		return NULL;
 	}
 
-	s->pos += s->dirent->d_reclen;
+	if (msg.o.err < 0) {
+		free(dirp->dirent);
+		dirp->dirent = NULL;
+		/* Any other errors should be set, ENOENT signals
+		 * end of a directory stream
+		 */
+		if (msg.o.err != -ENOENT) {
+			errno = -msg.o.err;
+		}
+		return NULL;
+	}
 
-	return s->dirent;
+	dirp->pos += dirp->dirent->d_reclen;
+
+	return dirp->dirent;
 }
 
 
 DIR *opendir(const char *dirname)
 {
 	char *canonical_name = resolve_path(dirname, NULL, 1, 0);
-	DIR *s = calloc(1, sizeof(DIR));
+	DIR *dirp = calloc(1, sizeof(DIR));
 
-	if ((canonical_name == NULL) || (s == NULL)) {
+	if ((canonical_name == NULL) || (dirp == NULL)) {
 		free(canonical_name);
-		free(s);
+		free(dirp);
 		return NULL; /* errno set by resolve_path */
 	}
 
-	if (!dirname[0] || (safe_lookup(canonical_name, NULL, &s->oid) < 0)) {
+	if (!dirname[0] || (safe_lookup(canonical_name, NULL, &dirp->oid) < 0)) {
 		free(canonical_name);
-		free(s);
+		free(dirp);
 		errno = ENOENT;
-		return NULL; /* ENOENT */
+		return NULL;
 	}
 
 	free(canonical_name);
-	s->dirent = NULL;
+	dirp->dirent = NULL;
+	/* Following field is only valid in fdopendir */
+	dirp->fd = -1;
 
 	msg_t msg = {
 		.type = mtGetAttr,
-		.oid = s->oid,
+		.oid = dirp->oid,
 		.i.attr.type = atType,
 	};
 
-	if ((msgSend(s->oid.port, &msg) < 0) || (msg.o.err < 0)) {
-		free(s);
-		errno = EIO;
-		return NULL; /* EIO */
-	}
-
-	if (msg.o.attr.val != otDir) {
-		free(s);
-		errno = ENOTDIR;
-		return NULL; /* ENOTDIR */
-	}
-
-	memset(&msg, 0, sizeof(msg));
-	msg.type = mtOpen;
-	msg.oid = s->oid;
-	msg.i.openclose.flags = 0;
-
-	if (msgSend(s->oid.port, &msg) < 0) {
-		free(s);
-		errno = EIO;
-		return NULL; /* EIO */
-	}
-
-	if (msg.o.err < 0) {
-		free(s);
+	if ((msgSend(dirp->oid.port, &msg) < 0) || (msg.o.err < 0)) {
+		free(dirp);
 		errno = EIO;
 		return NULL;
 	}
 
-	return s;
+	if (msg.o.attr.val != otDir) {
+		free(dirp);
+		errno = ENOTDIR;
+		return NULL;
+	}
+
+	memset(&msg, 0, sizeof(msg));
+	msg.type = mtOpen;
+	msg.oid = dirp->oid;
+	msg.i.openclose.flags = 0;
+
+	if (msgSend(dirp->oid.port, &msg) < 0 || (msg.o.err < 0)) {
+		free(dirp);
+		errno = EIO;
+		return NULL;
+	}
+
+	return dirp;
+}
+
+
+/*
+ * FIXME: fstat() does not return proper OIDs for directories mounted under
+ * a different filesystem.
+ */
+DIR *fdopendir(int fd)
+{
+	DIR *dirp;
+	struct stat statbuf;
+	int flags;
+	off_t pos;
+
+	flags = fcntl(fd, F_GETFL);
+	if (flags < 0) {
+		return NULL; /* errno set by fcntl() */
+	}
+	if ((flags & O_RDONLY) == 0) {
+		errno = EBADF;
+		return NULL;
+	}
+
+	pos = lseek(fd, 0, SEEK_CUR);
+	if (pos == (off_t)-1) {
+		return NULL; /* errno set by lseek() */
+	}
+
+	if (fstat(fd, &statbuf) < 0) {
+		return NULL; /* errno set by fstat() */
+	}
+
+	if (!S_ISDIR(statbuf.st_mode)) {
+		errno = ENOTDIR;
+		return NULL;
+	}
+
+	dirp = calloc(1, sizeof(DIR));
+	if (dirp == NULL) {
+		return NULL; /* errno set by calloc() */
+	}
+	dirp->oid.port = statbuf.st_dev;
+	dirp->oid.id = statbuf.st_ino;
+	dirp->fd = fd;
+	dirp->dirent = NULL;
+	dirp->pos = pos;
+
+	if (fcntl(fd, F_SETFD, FD_CLOEXEC) < 0) {
+		free(dirp);
+		return NULL; /* errno set by fcntl() */
+	}
+
+	return dirp;
+}
+
+
+void seekdir(DIR *dirp, long loc)
+{
+	if (dirp == NULL) {
+		errno = EBADF;
+		return;
+	}
+
+	dirp->pos = (off_t)loc;
+}
+
+
+long telldir(DIR *dirp)
+{
+	if (dirp == NULL) {
+		errno = EBADF;
+		return -1;
+	}
+
+	return (long)dirp->pos;
 }
 
 
 void rewinddir(DIR *dirp)
 {
+	if (dirp == NULL) {
+		errno = EBADF;
+		return;
+	}
+
 	dirp->pos = 0;
 }
 
@@ -450,13 +541,23 @@ int closedir(DIR *dirp)
 {
 	int ret = 0;
 
-	msg_t msg = {
-		.type = mtClose,
-		.oid = dirp->oid
-	};
+	if (dirp == NULL) {
+		return SET_ERRNO(-EBADF);
+	}
 
-	if ((msgSend(dirp->oid.port, &msg) < 0) || (msg.o.err < 0)) {
-		ret = -1;
+	if (dirp->fd >= 0) {
+		ret = close(dirp->fd);
+	}
+	else {
+		msg_t msg = {
+			.type = mtClose,
+			.oid = dirp->oid
+		};
+
+		if ((msgSend(dirp->oid.port, &msg) < 0) || (msg.o.err < 0)) {
+			errno = EIO;
+			ret = -1;
+		}
 	}
 
 	free(dirp->dirent);


unistd/dir.c
```diff
@@ -440,6 +441,55 @@
...
+
+
+void seekdir(DIR *dirp, long loc)
+{
+	dirp->pos = loc;
+}
+
+long telldir(DIR *dirp)
+{
+	return dirp->pos;
```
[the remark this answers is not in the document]
R4ken: No, this behavior exists in Linux, POSIX does not specify it

unistd/dir.c
```diff
@@ -440,6 +441,60 @@ DIR *opendir(const char *dirname)
 }
 
 
+DIR *fdopendir(int filedes)
```
Darchiv, line 444: Why `int filedes` instead of `int fd` like specified by POSIX? Is `filedes` used in existing code?

unistd/dir.c
```diff
@@ -440,6 +441,60 @@ DIR *opendir(const char *dirname)
...
+	struct stat statbuf;
+
+	if (s == NULL) {
+		errno = ENOMEM;
+		return NULL;
+	}
+
+	if (fstat(filedes, &statbuf) < 0) {
+		free(s);
+		return NULL; /* fstat sets EBADF */
```
Darchiv, line 456: Does this check whether or not `filedes` is a "valid file descriptor open for reading."?

> The fdopendir() function shall fail if:
> [EBADF] The fd argument is not a valid file descriptor open for reading.

unistd/dir.c
```diff
@@ -440,6 +441,60 @@ DIR *opendir(const char *dirname)
...
+
+	return s;
+}
+
+
+void seekdir(DIR *dirp, long loc)
+{
+	dirp->pos = loc;
+}
+
```
Darchiv, line 491: nitpick: Two empty lines between function bodies

unistd/dir.c
```diff
@@ -440,6 +441,73 @@ DIR *opendir(const char *dirname)
...
+		free(s);
+		return NULL;
+	}
+
+	fd_flags = fcntl(fd, F_GETFL);
+	if (fd_flags < 0) {
+		free(s);
+		return NULL;
+	}
+	if ((fd_flags & O_RDONLY) == 0) {
```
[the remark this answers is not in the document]
R4ken: `open` from libphoenix sets `EISDIR` on errno while opening directory with either `O_WRONLY` or `O_RDWR` flag

 Fragment of `open` function from libphoenix
```C
if (oflag & (O_WRONLY | O_RDWR)) {
  if ((err = stat(filename, &st)) < 0) {
	  if (errno != ENOENT)
		  return err;
  }
  else if (S_ISDIR(st.st_mode)) {
	  return SET_ERRNO(-EISDIR);
  }
}
```

unistd/dir.c
```diff
@@ -440,6 +441,55 @@
...
+
+	fcntl(filedes, F_SETFD, FD_CLOEXEC);
+
+	return s;
+}
+
+
+void seekdir(DIR *dirp, long loc)
+{
+	dirp->pos = loc;
```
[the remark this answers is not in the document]
oI0ck: Yes, you are right, although I think that this might be a good opportunity to introduce those checks here.
Linux's behaviour in these calls is not unreasonable and helping the programmer by making this deterministic is IMO good

unistd/dir.c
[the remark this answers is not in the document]
oI0ck: This behaviour seems common, I can replicate it on both Linux and one of the BSD's, I think we can keep it like that for the sake of compatibility.

oI0ck (dismissed):
I'd be good to introduce `NULL` checks while we are at it.

unistd/dir.c
```diff
@@ -450,6 +527,10 @@ int closedir(DIR *dirp)
 {
 	int ret = 0;
 
+	if (dirp != NULL) {
+		return -EBADF;
```
Darchiv: So, only `dirp == NULL` is allowed to proceed?

The same case in `telldir()`

include/dirent.h
```diff
@@ -53,6 +53,11 @@ extern struct dirent *readdir(DIR *s);
 
 extern DIR *opendir(const char *dirname);
 
+extern DIR *fdopendir(int fd);
+
+extern void seekdir(DIR *dirp, long loc);
+
+extern long telldir(DIR *dirp);
```
Darchiv, line 64: Keep two empty lines between function prototypes.

include/dirent.h
Darchiv: For some reason, `s` is used as a parameter name instead of `dirp`. Maybe make it consistent now, if you make changes to these functions already?

unistd/dir.c
Darchiv, line 389: As per POSIX:

> Upon successful completion, readdir() shall return a pointer to an object of type struct dirent. When an error is encountered, a null pointer shall be returned and errno shall be set to indicate the error. **When the end of the directory is encountered, a null pointer shall be returned and errno is not changed.**

Discriminating between the end of the directory and errors seems to rely on checking the value of `errno` while `readdir()` returns `NULL`. If so, it makes sense to set `errno` in all cases when an error occurs - i.e. when:
- `calloc()` above fails,
- `msgSend()` above fails - there is a comment about `EIO`, but it is not assigned to `errno`!); `msgSend()` only returns error codes, it does not set `errno`,
- `msg.o.err < 0`, so `errno = -msg.o.err` here? Or `errno = EIO`, but I think returning an actual error code from the same abstraction layer (a filesystem server) is nice, especially for debugging.

If you already make changes to error reporting in this already-existing function, then please do so fully.

unistd/dir.c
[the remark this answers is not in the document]
Darchiv: Please look carefully at the condition.

Also, I agree that POSIX does not specify how to report the error, but: should we deviate from the "common" `errno` setting, as is the case with other directory functions in this file? We don't even have this function documented in p-r-doc, so using an approach other than `errno` seems counter-intuitive. Please either use `errno` or make a PR to p-r-doc (which will eventually be needed, though).

unistd/dir.c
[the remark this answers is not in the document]
Darchiv: Is `fd_flags & O_RDONLY) == 0` idiomatic? WHat about use of `O_ACCMODE`? Should we assume that `O_RDONLY`, `O_WRONLY`, and `O_RDWR` are bitmasks (0x1, 0x2, 0x4)? Or they are an enumeration (0x0, 0x1, 0x2) to be checked with `O_ACCMODE == 0x3`? (I'm not asking about how we or Linux implement it, but what is the most portable way.)

unistd/dir.c
oI0ck: I think we should differentiate between cases when we reach end of directory stream (with `err` set to `-ENOENT`) and cases when we return any other errors; `errno` should be only modified in latter case.

This imposes a requirement on our filesystem implementations to adhere to this behaviour (to return `-ENOENT` on end of dirstream), which is true for `ext2` at least, not sure about other filesystems.

oI0ck:
@Darchiv 
> Is fd_flags & O_RDONLY) == 0 idiomatic? WHat about use of O_ACCMODE? Should we assume that O_RDONLY, O_WRONLY, and O_RDWR are bitmasks (0x1, 0x2, 0x4)? Or they are an enumeration (0x0, 0x1, 0x2) to be checked with O_ACCMODE == 0x3? (I'm not asking about how we or Linux implement it, but what is the most portable way.)

Most POSIX conformant/compilant operating systems seem to treat those values as enums, not bitfields.
I'll change it to enum-like values for the sake or portability.

include/dirent.h
oI0ck: Addressed.

unistd/dir.c
oI0ck: Addressed.

As far as I checked, most of our filesystems return `-ENOENT` on reaching the end of dirstream, with the exception of JFFS2. I'll create and link a new pr in p-r-fs

unistd/dir.c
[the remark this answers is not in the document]
oI0ck: Addressed. Will create a new, linked PR in p-r-k.

unistd/dir.c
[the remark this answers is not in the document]
oI0ck: Addressed

unistd/dir.c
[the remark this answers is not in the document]
oI0ck: Also, this check would be a no-op in enum-like access modes. It seems that it is here in this version to check if ~any~ mode is set (as rdonly is the default), on most systems `O_RDONLY` is defined as 0, so it is implicitly a default and there is no need to check it here.

oI0ck:
I removed the bitfield commit as it required further changes. The dependency graph on the WAMR PR grows too deep and if we make every issue which has come up a hard dependency, we won't merge it ever.

unistd/dir.c
```diff
@@ -440,8 +452,79 @@ DIR *opendir(const char *dirname)
 }
 
 
+DIR *fdopendir(int fd)
```
ziemleszcz, line 458: Where is this file descriptor released?

unistd/dir.c
```diff
@@ -440,8 +452,79 @@ DIR *opendir(const char *dirname)
...
+		return NULL; /* EIO */
+	}
+
+	if (msg.o.attr.val != otDir) {
+		errno = ENOTDIR;
+		return NULL; /* ENOTDIR */
+	}
+
+	if (fcntl(fd, F_SETFD, FD_CLOEXEC) < 0) {
+		return NULL;
```
julianuziemblo: why no errno set here?

unistd/dir.c
```diff
@@ -347,36 +348,47 @@ char *realpath(const char *path, char *resolved_path)
...
-	if (s->dirent == NULL) {
-		if ((s->dirent = calloc(1, sizeof(struct dirent) + NAME_MAX + 1)) == NULL)
+	if (dirp == NULL) {
+		errno = EBADF;
+		return NULL;
+	}
+
+	if (dirp->dirent == NULL) {
+		if ((dirp->dirent = calloc(1, sizeof(struct dirent) + NAME_MAX + 1)) == NULL) {
+			errno = -ENOMEM;
```
julianuziemblo: errno set to negative value

unistd/dir.c
```diff
@@ -440,8 +452,79 @@ DIR *opendir(const char *dirname)
...
+		return NULL;
+	}
+
+	s = calloc(1, sizeof(DIR));
+	if (s == NULL) {
+		errno = ENOMEM;
+		return NULL;
+	}
+	s->oid = msg.oid;
+	s->dirent = NULL;
```
ziemleszcz: According to POSIX: `The file offset associated with the file descriptor at the time of the call determines which entries are returned`. However, `pos` is always set to 0.

unistd/dir.c
```diff
@@ -440,8 +452,79 @@ DIR *opendir(const char *dirname)
...
 
+DIR *fdopendir(int fd)
+{
+	DIR *s;
+	struct stat statbuf;
+	int fd_flags;
+
+	fd_flags = fcntl(fd, F_GETFL);
+	if (fd_flags < 0) {
+		return NULL;
```
julianuziemblo: errno not set?

unistd/dir.c
```diff
@@ -440,8 +452,79 @@ DIR *opendir(const char *dirname)
 }
 
 
+DIR *fdopendir(int fd)
+{
+	DIR *s;
```
julianuziemblo: nit: this could also be named "dirp" for consistency?

unistd/dir.c
```diff
@@ -440,8 +452,79 @@ DIR *opendir(const char *dirname)
...
+	}
+
+	if (fcntl(fd, F_SETFD, FD_CLOEXEC) < 0) {
+		return NULL;
+	}
+
+	s = calloc(1, sizeof(DIR));
+	if (s == NULL) {
+		errno = ENOMEM;
+		return NULL;
```
ziemleszcz: Side effect: FD_CLOEXEC is set.

unistd/dir.c
ziemleszcz: ENOMEM

unistd/dir.c
```diff
@@ -440,8 +452,79 @@ DIR *opendir(const char *dirname)
...
+	s->oid = msg.oid;
+	s->dirent = NULL;
+
+	return s;
+}
+
+
+void seekdir(DIR *dirp, long loc)
+{
+	assert(dirp != NULL);
```
ziemleszcz: Don't assert, check it.

unistd/dir.c
```diff
@@ -440,8 +452,79 @@ DIR *opendir(const char *dirname)
...
+		return SET_ERRNO(-EBADF);
+	}
+
+	return dirp->pos;
+}
+
+
 void rewinddir(DIR *dirp)
 {
+	assert(dirp != NULL);
```
ziemleszcz: Don't assert, check it.

unistd/dir.c
julianuziemblo: probably errno should be set

unistd/dir.c
oI0ck: It seems to me that it does not have to be released according to POSIX ([1] "Upon successful return from fdopendir(), the file descriptor is under the control of the system, and if any attempt is made to close the file descriptor, or to modify the state of the associated description, other than by means of closedir(), readdir(), readdir_r(), rewinddir(), or XSI seekdir(), the behavior is undefined.").

[1] - https://pubs.opengroup.org/onlinepubs/9799919799/functions/fdopendir.html

unistd/dir.c
oI0ck: Checking would work if if was possible to return an error here, but this function returns `void`.
IIRC POSIX does not specify `seekdir()` behaviour when `dirp` is NULL or incorrect in any other way.

`assert()` here seems like a good way to catch bugs without introducing further undefined behaviour.

unistd/dir.c
ziemleszcz: This suggests the system, not the user, should release it (probably in `closedir()`). I don’t see this implemented in `closedir()`.

unistd/dir.c
ziemleszcz: Regardless of the assertion, this pointer should be checked to avoid crashing the user application.

unistd/dir.c
oI0ck: If we hit this branch, `errno` has already been set by `fnctl`

unistd/dir.c
oI0ck: As above

unistd/dir.c
julianuziemblo: Please add comment as other parts of the code have comments in those cases too

unistd/dir.c
oI0ck: Implemented

unistd/dir.c
```diff
@@ -440,8 +452,93 @@ DIR *opendir(const char *dirname)
...
+
+	msg_t msg = {
+		.type = mtGetAttr,
+		.oid = { .port = statbuf.st_dev, .id = statbuf.st_ino },
+		.i.attr.type = atType,
+	};
+
+	if ((msgSend(statbuf.st_dev, &msg) < 0) || (msg.o.err < 0)) {
+		errno = EIO;
+		return NULL; /* EIO */
```
ziemleszcz: Redundant comment.

unistd/dir.c
```diff
@@ -440,8 +452,93 @@ DIR *opendir(const char *dirname)
...
+	}
+
+	if (msg.o.attr.val != otDir) {
+		errno = ENOTDIR;
+		return NULL; /* ENOTDIR */
```
ziemleszcz: Redundant comment.

unistd/dir.c
```diff
@@ -440,8 +452,93 @@ DIR *opendir(const char *dirname)
...
+	fd_flags = fcntl(fd, F_GETFL);
+	if (fd_flags < 0) {
+		return NULL; /* errno set by fcntl() */
+	}
+	if ((fd_flags & O_RDONLY) == 0) {
+		errno = EBADF;
+		return NULL;
+	}
+
+	if ((pos = lseek(fd, 0, SEEK_CUR)) < 0) {
```
ziemleszcz: MISRA style and return value comparision:
pos = lseek(fd, 0, SEEK_CUR);
if (pos == (off_t)-1) {
...
}

unistd/dir.c
```diff
@@ -440,8 +452,93 @@ DIR *opendir(const char *dirname)
 }
 
 
+DIR *fdopendir(int fd)
+{
+	DIR *dirp;
+	struct stat statbuf;
+	int fd_flags;
+	long pos;
```
ziemleszcz: off_t

unistd/dir.c
```diff
@@ -440,8 +452,93 @@ DIR *opendir(const char *dirname)
 }
 
 
+DIR *fdopendir(int fd)
+{
+	DIR *dirp;
+	struct stat statbuf;
+	int fd_flags;
```
ziemleszcz: CamelCase: fdFlags or simply flags.

unistd/dir.c
```diff
@@ -440,8 +452,93 @@ DIR *opendir(const char *dirname)
...
+		return NULL; /* errno set by fstat() */
+	}
+
+	msg_t msg = {
+		.type = mtGetAttr,
+		.oid = { .port = statbuf.st_dev, .id = statbuf.st_ino },
+		.i.attr.type = atType,
+	};
+
+	if ((msgSend(statbuf.st_dev, &msg) < 0) || (msg.o.err < 0)) {
```
ziemleszcz: Is `S_ISDIR(statbuf.st_mode)` not working? I see it used in other places.

include/stdio.h
```diff
@@ -70,6 +70,7 @@ typedef struct _FILE {
 
 typedef struct _DIR {
 	oid_t oid;
+	int fd;
 	size_t pos;
```
ziemleszcz: off_t?

unistd/dir.c
```diff
@@ -347,36 +348,47 @@ char *realpath(const char *path, char *resolved_path)
...
 {
-	if (s->dirent == NULL) {
-		if ((s->dirent = calloc(1, sizeof(struct dirent) + NAME_MAX + 1)) == NULL)
+	if (dirp == NULL) {
+		errno = EBADF;
+		return NULL;
+	}
+
+	if (dirp->dirent == NULL) {
+		if ((dirp->dirent = calloc(1, sizeof(struct dirent) + NAME_MAX + 1)) == NULL) {
```
ziemleszcz: MISRA style:
dirp->dirent = calloc()
if (dirp->dirent == NULL) {
}

unistd/dir.c
```diff
@@ -347,36 +348,47 @@ char *realpath(const char *path, char *resolved_path)
...
 	}
 
 	if (msg.o.err < 0) {
-		free(s->dirent);
-		s->dirent = NULL;
+		free(dirp->dirent);
+		dirp->dirent = NULL;
+		if (msg.o.err != -ENOENT) {
+			errno = -msg.o.err;
+		}
```
ziemleszcz, line 388: Else no `errno` is set?

unistd/dir.c
ziemleszcz: Redundant.

unistd/dir.c
```diff
@@ -440,8 +452,93 @@ DIR *opendir(const char *dirname)
...
+	}
+
+	dirp = calloc(1, sizeof(DIR));
+	if (dirp == NULL) {
+		errno = ENOMEM;
+		return NULL;
+	}
+	dirp->oid = msg.oid;
+	dirp->fd = fd;
+	dirp->dirent = NULL;
```
ziemleszcz, line 495: Redundant.

include/dirent.h
```diff
@@ -48,12 +48,21 @@ struct dirent {
 };
 
 
-extern struct dirent *readdir(DIR *s);
+extern struct dirent *readdir(DIR *dirp);
```
ziemleszcz, line 52: Could you add a comment noting that this implementation is not thread-safe?

unistd/dir.c
```diff
@@ -440,8 +452,93 @@ DIR *opendir(const char *dirname)
...
+	}
+
+	dirp->pos = loc;
+}
+
+
+long telldir(DIR *dirp)
+{
+	if (dirp == NULL) {
+		return SET_ERRNO(-EBADF);
```
ziemleszcz: Why not `errno = EBADF`?

unistd/dir.c
```diff
@@ -450,15 +547,22 @@ int closedir(DIR *dirp)
 {
 	int ret = 0;
 
+	if (dirp == NULL) {
+		return SET_ERRNO(-EBADF);
```
ziemleszcz, line 545: Why not errno = EBADF?

unistd/dir.c
```diff
@@ -450,15 +547,22 @@ int closedir(DIR *dirp)
...
 		.type = mtClose,
 		.oid = dirp->oid
 	};
 
 	if ((msgSend(dirp->oid.port, &msg) < 0) || (msg.o.err < 0)) {
+		errno = EIO;
 		ret = -1;
 	}
 
+	close(dirp->fd);
```
ziemleszcz: `dirp->fd` is not valid for `opendir()`.

unistd/dir.c
oI0ck: This seems to be the way to set `errno` in functions that return integer values. I don't understand what is wrong with this snippet.

unistd/dir.c
oI0ck: Yes, this is a special case, `-ENOENT` is returned by the filesystem server when it reaches and of the 'dirstream'. This is not an error condition in `readdir()` as it should then just return `NULL` and leave `errno` unset.

unistd/dir.c
oI0ck: Corrected

unistd/dir.c
ziemleszcz: Understood. In that case I would revert this check and add a comment as it was not obvious to me:


if (msg.o.err == -ENOENT) {
/* end of the directory reached */
}
else {
errno = -msg.o.err;
}

unistd/dir.c
ziemleszcz: `SET_ERRNO(err)` should be used when `err` could be 0 (success). In this case you know it's an error so you can set `errno` directly.

unistd/dir.c
ziemleszcz: Use `pos == (off_t)-1`.

ziemleszcz:
Please add tests for all DIR functions in phoenix-rtos-test, including both valid use cases and expected failure cases (bad input).

oI0ck:
@ziemleszcz I don't think that `seekdir()`, `telldir()` or `rewinddir()` deserve tests, they are literally 5 lines long, the only thing I could test them for is the returned errors, but someone removing a branch with an error handling code seems improbable.

I already have a [PR](https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/470) for testing the `ENOENT` case in `readdir()`, I think I could extend it by a test for retaining the offset from the FD for `fdopendir` and for `ENOTDIR` behaviour.

ziemleszcz:
All newly implemented POSIX functions should have test coverage, regardless of their size.

For `seekdir()`, you can verify that it correctly restores the position returned by `telldir()` after reading several entries.
For `rewinddir()`, check that `readdir()` starts again from the beginning.

These tests may seem simple, but they help ensure that future changes do not break these functions.

oI0ck:
Also, I don't think that retaining position from a file descriptor in `fdopendir()` works now, as `readdir()` sends a `mtReaddir` message to the filesystem server, and does not use a wrapper syscall that would update the offset in the kernel.

ziemleszcz:
`mtReaddir` already provides an offset (`i.readdir.offs`), so `fdopendir()` can use the current file offset. In any case, we should add tests to verify this behavior.

oI0ck:
Yes, it does but this is not the problem, and it is not related to tests, just a problem I noticed. `readdir()` uses `i.readdir.offs` for sending dir stream offset to the filesystem server. The problem is that, as you've previously noticed, a directory stream returned by `fdopendir()` must retain a position/offset of a file descriptor passed to it.

We don't really need file descriptors here, `readdir()` works by sending messages to the filesystem server, but apart from that, it does not go through the file descriptor modifying 'proxy' syscalls, like `write()` or `read()` do.

My point is, with the way that we currently implement walking directory trees, `lseek(fd, 0, SEEK_CUR)` will ALWAYS  return a 0 when `fd` refers to an open directory, because the offset is accounted for in the userspace, not the kernel.

In order for `lseek()` to work in this case,  and for that matter, in order for position retaining in `fdopendir()` to work, we'd have to implement a new syscall, like `getdents()`, which would take a file descriptor, update its offset in the kernel and forward the `mtReaddir` message to the fs server.

oI0ck:
So my question here is, do we want to add a new syscall here or shall we leave this functionality as-is and correct it later, possibly after the release.

oI0ck:
Actually, this is a bit related to tests as it came out as an issue after I implemented a test for it :D

ziemleszcz:
POSIX states: “The file offset associated with the file descriptor at the time of the call determines which entries are returned.”

So, the initial position of DIR should be set based on the file offset at the time of the `fdopendir()` call. After that, the file descriptor is under system control (“Upon successful return from fdopendir(), the file descriptor is under the control of the system”).

My understanding is that subsequent `readdir()` calls do not need to update the file descriptor’s offset. Only the internal DIR position should be advanced.

ziemleszcz:
You could also check how glibc implements this, but in my opinion updating the file descriptor’s offset is not required by POSIX.

oI0ck:
It is not required by POSIX, my point is that the state of the file descriptor for a directory will always be 0, so there is no need really to do an `lseek()` there.

oI0ck:
We can probably regard this as a quirk of our implementation right now, I'm preparing a p-r-docs PR for this change so I'll mention it there.

ziemleszcz:
If you mean calling `lseek()` in `fdopendir()`, then it is still required, since the current file position does not have to be 0. The user application may have already performed reads on this descriptor. It may seem unusual, but that is what POSIX specifies.

Please check how glibc handles this case (passing descriptor with a non-zero offset to `fdopendir()`).

oI0ck:
I checked OpenBSD sources, they do set the `lseek()` at all times.

> The user application may have already performed reads on this descriptor`\

Is performing a `read()` or `write()` on a directory file descriptor a valid operation? Do we, or the filesystem servers, ensure that, if it is possible to perform such odd operations, the offset stays valid? Do we even have a filesystem implementation that supports operations like these so we can test that behaviour?

We also have an inconsistency here, because `fdopendir()` does not make an assumption about the initial position value, but `opendir()` does, and it's zero. Moreover, because it does not interact with file descriptors at all, there is no way of determining the initial offset.

oI0ck:
In this case, I think we might want to make `opendir()` also use file descriptors, so that the userspace won't make assumptions about the position, but gets it from `lseek()`.

ziemleszcz:
Directories are files, so reading them is fine. I assume the difference between how `opendir()` and `fdopendir()` set the initial position stems from this. Not sure how our file systems handle this (reading directories).
I propose to test it on Linux and RTOS:

 - open directory,
 - move offset past first entry (with known size),
 - call fdopendir() + readdir(),
 - expect second entry

ziemleszcz:
For `opendir()` POSIX clearly states: “The directory stream is positioned at the first entry.”

oI0ck:
Yep, I do have such [test](https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/477) and [it fails on RTOS](https://github.com/phoenix-rtos/phoenix-rtos-tests/actions/runs/25444606889/job/74645756983?pr=477), because we lack a way to report reading directories back to kernel.

I'm almost sure it would work on Linux though, we'd have to move it by using `getdents()` as most filesystems implement `read()` for directories as [just returining `-EISDIR`](https://elixir.bootlin.com/linux/v7.0.1/source/fs/libfs.c#L225).
`getdents()` would advance the internal offset of the file descriptor with a filesystem specific callback, so after `fdopendir()` and `readdir()` it should return the second entry, which does not happen on RTOS, again, because we lack a primitive, `getdents()`, to do so.

I've looked through our filesystems source code and you are in fact right, `ext2` for some wicked reason implements both `mtRead` and `mtReaddir` on directories, so `read()` will work on a directory, advancing offset in the kernel (though, it's very hackish, me no like). I think FAT allows for it too.

I just think this behavour is confusing, semantically `read()`'ing from a directory FD is a bit odd (a function for reading contents of a file to a byte buffer? Unices don't treat directories as files or byte arrays. Directories are structured, files don't have to be), although there might be some uses.

In the end, we can leave it as-is for now, maybe implementing `getdents()` in the future would just make sense.
I'll correct these tests to use `read()`'s and launch only on targets with filesystems that do support `read()` on directories.

oI0ck:
@ziemleszcz I see that while I had compiled some tests of my own, there is already a larger PR for `dirent` tests, https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/475. I think I can build up on their work by adding tests for new functions after it gets merged.

In the mean time, is this PR by itself clear to go in your opinion?

unistd/dir.c
```diff
@@ -450,15 +550,24 @@ int closedir(DIR *dirp)
...
 		.oid = dirp->oid
 	};
 
 	if ((msgSend(dirp->oid.port, &msg) < 0) || (msg.o.err < 0)) {
+		errno = EIO;
 		ret = -1;
 	}
 
+	if (dirp->fd >= 0) {
+		close(dirp->fd);
```
ziemleszcz: Remove `mtClose` message in this path. `close()` already handles it. Maybe restructure this into `if-else` block.

unistd/dir.c
```diff
@@ -347,101 +348,200 @@ char *realpath(const char *path, char *resolved_path)
...
+	pos = lseek(fd, 0, SEEK_CUR);
+	if (pos == (off_t)-1) {
+		return NULL; /* errno set by lseek() */
+	}
+
+	if (fstat(fd, &statbuf) < 0) {
+		return NULL; /* errno set by fstat() */
+	}
+
+	msg_t msg = {
```
ziemleszcz: We don’t need this struct since no message is sent here.

unistd/dir.c
```diff
@@ -347,101 +348,200 @@ char *realpath(const char *path, char *resolved_path)
...
+		errno = ENOTDIR;
+		return NULL;
+	}
+
+	dirp = calloc(1, sizeof(DIR));
+	if (dirp == NULL) {
+		errno = ENOMEM;
+		return NULL;
+	}
+	dirp->oid = msg.oid;
```
ziemleszcz: Use `oid` from `statbuf` (verify that it is correct `oid`).

unistd/dir.c
ziemleszcz: Assign result of `close()` to `ret`.

unistd/dir.c
```diff
@@ -347,101 +348,195 @@ char *realpath(const char *path, char *resolved_path)
...
+	}
+
+	if (!S_ISDIR(statbuf.st_mode)) {
+		errno = ENOTDIR;
+		return NULL;
+	}
+
+	dirp = calloc(1, sizeof(DIR));
+	if (dirp == NULL) {
+		errno = ENOMEM;
```
ziemleszcz: calloc() sets ENOMEM.

unistd/dir.c
```diff
@@ -347,101 +348,195 @@ char *realpath(const char *path, char *resolved_path)
...
-	if (msgSend(s->oid.port, &msg) < 0) {
-		free(s);
+	if (msgSend(dirp->oid.port, &msg) < 0) {
+		free(dirp);
 		errno = EIO;
-		return NULL; /* EIO */
+		return NULL;
 	}
 
 	if (msg.o.err < 0) {
```
ziemleszcz: Could be combined with `if` above.

oI0ck: As in, making an `if` and `if else`?

ziemleszcz: As in L561.

unistd/dir.c
ziemleszcz: Please verify that this is correct OID by running some tests. AI review reports it's not.

oI0ck: I changed it to use `oid` from `statbuf`, which should be correct. I don't think I can verify file descriptors `oid` in any other way. My tests pass, so I'm not sure what can be wrong here.

ziemleszcz (approved):

PR #437, opened 2025-09-17
Add possibility to disable float in stdio functions (*printf, *scanf)

anglov:
## Description

stdio functions, especially *printf and *scanf involves usage of code
for handling floating point. In resource-constrained targets, especially
without hardware floating point support, it causes that each binary that
uses function from printf or scanf family, have to be compiled with
large code logic for handling floating points and enable software
floating point from compiler.
    
With this solution it is possible to compile libphoenix without support
of formats %{A, a, E, e, F, f, G, g} for *printf and *scanf
    
As this solution require building with additional environment variable,
it is temporary. Splitting libphoenix to smaller parts and mechanisms
for chosing desired version should be involved in the future

## Types of changes

## How Has This Been Tested?

## Checklist:

- [ x My changes generate no new compilation warnings for any of the targets.

## Special treatment


nalajcie:
not sure: as we're preventing to get `double` from va args, the `printf` might break in undesired ways if somebody tries to do it by accident. Maybe we should `assert` if somebody passes any of the not-supported val? Or try to work anyway (harder, we would need to get integer with sizeof(double) from varargs?)

anglov:
> not sure: as we're preventing to get `double` from va args, the `printf` might break in undesired ways if somebody tries to do it by accident. Maybe we should `assert` if somebody passes any of the not-supported val? Or try to work anyway (harder, we would need to get integer with sizeof(double) from varargs?)

Well, initially I was skeptical about that but after talk with @xvuko and looking at other libc I decided with following approach:
In case of printf:
- take floating type argument from stack, but not assign it.
- print format specifier
eg. when LIBPHOENIX_IO_NO_FLOAT is enabled, and we have
```C
float f1 = 42.0;
int i1 = 84;
printf("%f %d", f1, i1);
```
should return
```
%f 84
```
In case of scanf:
- return number of parsed arguments before floating point format was read
In fact is hard to do anything better in that case.

stdio/format.c
```diff
@@ -76,6 +76,7 @@
 	} while (0)
 
 
+#ifndef IO_NO_FLOAT
```
xvuko, line 79: #ifndef is not needed here

stdio/scanf.c
```diff
@@ -684,7 +685,9 @@ static int scanf_parse(char *ccltab, const char *inp, int *inr, char const *fmt0
 
 					nassigned++;
 				}
-
+#else
+				return (nconversions != 0 ? nassigned : -1);
+#endif
 				break;
```
xvuko, line 691: `/* never reached */`
lol

xvuko (approved):

PR #454, opened 2025-12-05
sys/threads: modify threadsinfo declaration

oI0ck:
JIRA: RTOS-1187

Depends-On: phoenix-rtos-utils#251
Depends-On: phoenix-rtos-kernel#720

## Description
This pull request adjusts the function declaration to respective changes in the kernel

## Types of changes

## How Has This Been Tested?

## Checklist:
- [X] I have updated the documentation accordingly.

## Special treatment
- https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/720
- https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/251
- https://github.com/phoenix-rtos/phoenix-rtos-doc/pull/247

### DIFF
--- a/include/sys/threads.h
+++ b/include/sys/threads.h
@@ -74,7 +74,19 @@ static inline int beginthread(void (*start)(void *), unsigned int priority, void
 }
 
 
-extern int threadsinfo(int n, threadinfo_t *info);
+extern int sys_threadsinfo(int tid, unsigned int flags, int n, threadinfo_t *info);
+
+
+static inline int threadsinfo(int n, unsigned int flags, threadinfo_t *info)
+{
+	return sys_threadsinfo(PH_THREADINFO_THREADS_ALL, flags, n, info);
+}
+
+
+static inline int threadinfo(int tid, unsigned int flags, threadinfo_t *info)
+{
+	return sys_threadsinfo(tid, flags, 1, info);
+}
 
 
 extern int priority(int priority);


ziemleszcz:
Breaking commits should be marked by `!`.

ziemleszcz:
Add kernel PR to the list list of additional PRs.

include/sys/threads.h
```diff
@@ -74,7 +74,7 @@ static inline int beginthread(void (*start)(void *), unsigned int priority, void
 }
 
 
-extern int threadsinfo(int n, threadinfo_t *info);
+extern int threadsinfo(int tid, unsigned int flags, int n, threadinfo_t *info);
```
ziemleszcz: I don't like this API change. The function has become too complex. I'd rather add additional syscall wrappers to preserve backward compatibility:

/* The same as old API. Returns thread count if info==NULL? */
int threadsinfo(int n, threadinfo_t *info)

/* Adds `flags` argument. */
int threadsinfo3(int n, threadinfo_t *info, unsigned int flags);

/* Single thread info, not sure if we need `flags` here. Returns 0 on success? */
int threadinfo(int tid, threadinfo_t *info, unsigned int flags)

oI0ck: There would be a symbol collision if I were to change `threadsinfo()` declaration here.

oI0ck: To achieve what you are describing here, I'd have to rename the syscall in the kernel.
IMHO the function (or it's declaration) is not overly complex.

adamgreloch: The ABI breaks anyway, I'm not sure there's a need to preserve the compat when the `threadsinfo` is used in psh applets only.

However, the complexity concern is valid IMHO as tid == -1 and  n != 1 are mutually exclusive. I'm in for having `threadinfo` and `threadsinfo` as seperate wrappers for `sys_threadsinfo` syscall. Renaming the syscall is pretty much free, as the PR is still not merged.

Flags are important for both cases.

adamgreloch (dismissed):

include/sys/threads.h
```diff
@@ -74,7 +74,19 @@ static inline int beginthread(void (*start)(void *), unsigned int priority, void
 }
 
 
-extern int threadsinfo(int n, threadinfo_t *info);
+extern int sys_threadsinfo(int tid, unsigned int flags, int n, threadinfo_t *info);
```
ziemleszcz, line 77: Remove.

include/sys/threads.h
```diff
@@ -74,7 +74,19 @@ static inline int beginthread(void (*start)(void *), unsigned int priority, void
...
+
+
+static inline int threadsinfo(int n, unsigned int flags, threadinfo_t *info)
```
ziemleszcz, line 80: Use `inline` only. See https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1722.

include/sys/threads.h
```diff
@@ -74,7 +74,19 @@ static inline int beginthread(void (*start)(void *), unsigned int priority, void
...
+{
+	return sys_threadsinfo(PH_THREADINFO_THREADS_ALL, flags, n, info);
```
ziemleszcz, line 82: I'm not sure if `-1` (with a comment) would be a better choice, since it's clearly not a valid `tid`. Alternatively, we could keep `PH_THREADINFO_THREADS_ALL` and pass it via `flags`.

include/sys/threads.h
```diff
@@ -74,7 +74,19 @@ static inline int beginthread(void (*start)(void *), unsigned int priority, void
...
+}
+
+
+static inline int threadinfo(int tid, unsigned int flags, threadinfo_t *info)
+{
+	return sys_threadsinfo(tid, flags, 1, info);
```
ziemleszcz, line 88: Return error if `tid < 0`.

include/sys/threads.h
```diff
@@ -74,7 +74,19 @@ static inline int beginthread(void (*start)(void *), unsigned int priority, void
...
-extern int threadsinfo(int n, threadinfo_t *info);
+extern int sys_threadsinfo(int tid, unsigned int flags, int n, threadinfo_t *info);
+
+
+static inline int threadsinfo(int n, unsigned int flags, threadinfo_t *info)
+{
+	return sys_threadsinfo(PH_THREADINFO_THREADS_ALL, flags, n, info);
+}
+
+
```
ziemleszcz, line 85: I insist on adding `threadscount()/threadcount()` wrapper. Using
```
 threadsinfo(0, PH_THREADINFO_OPT_THREADCOUNT, NULL)
```
to get the thread count is unnecessarily convoluted.

include/sys/threads.h
julianuziemblo: Can it still stay in the header file then?

include/sys/threads.h
oI0ck: Why? Other syscalls have declarations in libphoenix and there should a possibility to call raw (not the wrapper) syscall (with it's signature) if the developer desires it.

ziemleszcz: I misread it as `threadsinfo()`.

include/sys/threads.h
oI0ck: Disagreed, using magic values is bad. That's what we have constants and preprocessor defines for.

adamgreloch:
CI fails due to removal of `lib_idtreeof`, please rebase the PR

include/sys/threads.h
```diff
@@ -73,7 +74,29 @@ static inline int beginthread(void (*start)(void *), int priority, void *stack,
 }
 
 
-int threadsinfo(int n, threadinfo_t *info);
+int sys_threadsinfo(int tid, unsigned int flags, int n, threadinfo_t *info);
+
+
+inline int threadsinfo(int n, unsigned int flags, threadinfo_t *info)
```
ziemleszcz, line 80: Add `extern inline` copies of these function to `threads.c` (see `accept()` as an example of this pattern).

include/sys/threads.h
```diff
@@ -73,7 +74,29 @@ static inline int beginthread(void (*start)(void *), int priority, void *stack,
...
+		return -EINVAL;
+	}
+
+	return sys_threadsinfo(tid, flags, 1, info);
+}
+
+
+inline int threadcount(void)
+{
+	return sys_threadsinfo(PH_THREADINFO_THREADS_ALL, PH_THREADINFO_OPT_THREADCOUNT, 0, NULL);
```
ziemleszcz, line 98: Can it fail? If no, return `unsigned int`.

oI0ck: Collecting thread count should not error out, the `sys_threadinfo()` syscall though, can, so I'd leave it as it is.

sys/threads.c
```diff
@@ -102,3 +102,12 @@ int getPriority(void)
 	(void)sys_priority(PH_GET_PRIO, &prio);
 	return prio;
 }
+
+
+extern inline int threadsinfo(int n, unsigned int flags, threadinfo_t *info);
```
ziemleszcz: Move these 3 declarations before functions definitions.

oI0ck: done

ziemleszcz (approved):

PR #457, opened 2026-01-29
posix: implement POSIX semaphore interfaces

oI0ck:
TASK: RTOS-1407

## Description
This PR introduces implementation of POSIX standard semaphore interfaces.

## Types of changes

## How Has This Been Tested?
- [X] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/448
- [X] Tested by hand on: `ia32-generic-qemu`

## Checklist:
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
- [X] All new and existing linter checks and tests passed.
- [X] My changes generate no new compilation warnings for any of the targets.

## Special treatment

 - https://github.com/phoenix-rtos/phoenix-rtos-posixsrv/pull/27
 - https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/251
- [X] I will merge this PR by myself when appropriate.

### DIFF
--- a/include/semaphore.h
+++ b/include/semaphore.h
@@ -6,7 +6,7 @@
  * POSIX implementation - semaphores
  *
  * Copyright 2026 Phoenix Systems
- * Author: Michał Lach
+ * Author: Michal Lach
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
--- a/include/semaphore.h
+++ b/include/semaphore.h
@@ -1,0 +1,76 @@
+/*
+ * Phoenix-RTOS
+ *
+ * libphoenix
+ *
+ * POSIX implementation - semaphores
+ *
+ * Copyright 2026 Phoenix Systems
+ * Author: Michał Lach
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef _LIBPHOENIX_SEMAPHORE_H_
+#define _LIBPHOENIX_SEMAPHORE_H_
+
+#include <time.h>
+#include <limits.h>
+#include <sys/threads.h>
+#include <sys/semaphore.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define SEM_FAILED ((sem_t *)0)
+
+
+typedef struct _sem_t {
+	/* clang-format off */
+	enum { smNamed, smUnnamed } type;
+	/* clang-format on */
+
+	union {
+		semaphore_t unnamed;
+		oid_t named;
+	};
+} sem_t;
+
+
+extern int sem_wait(sem_t *sem);
+
+
+extern int sem_trywait(sem_t *sem);
+
+
+extern int sem_timedwait(sem_t *restrict sem, const struct timespec *restrict abs_timeout);
+
+
+extern int sem_getvalue(sem_t *restrict sem, int *restrict value);
+
+
+extern int sem_post(sem_t *sem);
+
+
+extern int sem_close(sem_t *sem);
+
+
+extern sem_t *sem_open(const char *name, int oflag, ... /* mode_t mode, unsigned int value */);
+
+
+extern int sem_unlink(const char *name);
+
+
+extern int sem_destroy(sem_t *sem);
+
+
+extern int sem_init(sem_t *sem, int pshared, unsigned int value);
+
+
+#ifdef __cplusplus
+}
+#endif
+
+
+#endif /* _LIBPHOENIX_SEMAPHORE_H_ */
--- a/include/sys/semaphore.h
+++ b/include/sys/semaphore.h
@@ -6,7 +6,7 @@
  * POSIX implementation - semaphores
  *
  * Copyright 2026 Phoenix Systems
- * Author: Michał Lach
+ * Author: Michal Lach
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
@@ -14,6 +14,7 @@
 #ifndef _LIBPHOENIX_SYS_SEMAPHORE_H_
 #define _LIBPHOENIX_SYS_SEMAPHORE_H_
 
+#include <assert.h>
 #include <limits.h>
 #include <sys/ioctl.h>
 
@@ -26,6 +27,11 @@ extern "C" {
 
 STATIC_ASSERT(SEM_VALUE_MAX >= _POSIX_SEM_VALUE_MAX, "SEM_VALUE_MAX shall be greater or equal to _POSIX_SEM_VALUE_MAX");
 STATIC_ASSERT(SEM_NSEMS_MAX >= _POSIX_SEM_NSEMS_MAX, "SEM_NSEMS_MAX shall be greater or equal to _POSIX_SEM_NSEMS_MAX");
+
+/* not required by POSIX, but ensures that
+ * sem_getvalue() does not return an overflowed value.
+ */
+STATIC_ASSERT(SEM_VALUE_MAX <= INT_MAX, "SEM_VALUE_MAX shall be smaller or equal to INT_MAX");
 
 #define SEM_UP           _IO('s', 0x1)
 #define SEM_DOWN         _IO('s', 0x2)
--- a/include/sys/semaphore.h
+++ b/include/sys/semaphore.h
@@ -1,0 +1,40 @@
+/*
+ * Phoenix-RTOS
+ *
+ * libphoenix
+ *
+ * POSIX implementation - semaphores
+ *
+ * Copyright 2026 Phoenix Systems
+ * Author: Michał Lach
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef _LIBPHOENIX_SYS_SEMAPHORE_H_
+#define _LIBPHOENIX_SYS_SEMAPHORE_H_
+
+#include <limits.h>
+#include <sys/ioctl.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define SEMAPHORE_PATH "/dev/posix/sem/"
+#define SEMCTL_PATH    "/dev/posix/semctl"
+
+STATIC_ASSERT(SEM_VALUE_MAX >= _POSIX_SEM_VALUE_MAX, "SEM_VALUE_MAX shall be greater or equal to _POSIX_SEM_VALUE_MAX");
+STATIC_ASSERT(SEM_NSEMS_MAX >= _POSIX_SEM_NSEMS_MAX, "SEM_NSEMS_MAX shall be greater or equal to _POSIX_SEM_NSEMS_MAX");
+
+#define SEM_UP           _IO('s', 0x1)
+#define SEM_DOWN         _IO('s', 0x2)
+#define SEM_DOWN_TRY     _IO('s', 0x3)
+#define SEM_DOWN_TIMEOUT _IOW('s', 0x4, time_t)
+#define SEM_GETVALUE     _IOR('s', 0x5, unsigned int)
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* _LIBPHOENIX_SYS_SEMAPHORE_H_ */
--- a/posix/sem.c
+++ b/posix/sem.c
@@ -1,0 +1,471 @@
+/*
+ * Phoenix-RTOS
+ *
+ * libphoenix
+ *
+ * POSIX implementation - semaphores
+ *
+ * Copyright 2026 Phoenix Systems
+ * Author: Michal Lach
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <stdbool.h>
+#include <fcntl.h>
+#include <string.h>
+#include <assert.h>
+#include <errno.h>
+#include <semaphore.h>
+#include <stdarg.h>
+#include <dirent.h>
+#include <limits.h>
+#include <sys/time.h>
+#include <sys/msg.h>
+#include <sys/types.h>
+#include <sys/minmax.h>
+#include <sys/semaphore.h>
+#include <phoenix/ioctl.h>
+
+#include "../common/util.h"
+
+
+static int sem_msgOpen(sem_t *sem)
+{
+	int ret = EOK;
+	msg_t msg = {
+		.type = mtOpen,
+		.oid = sem->named,
+	};
+
+	ret = msgSend(sem->named.port, &msg);
+	if (ret != EOK) {
+		return ret;
+	}
+
+	return msg.o.err;
+}
+
+
+static int sem_msgDown(sem_t *sem, bool try, time_t timeout)
+{
+	int ret = EOK;
+	ioctl_in_t *data;
+	msg_t msg = {
+		.type = mtDevCtl,
+		.oid = sem->named,
+	};
+
+	data = (ioctl_in_t *)&msg.i.raw;
+	if (try) {
+		data->request = SEM_DOWN_TRY;
+	}
+	else if (timeout > 0) {
+		data->request = SEM_DOWN_TIMEOUT;
+		memcpy(data->data, &timeout, sizeof(timeout));
+	}
+	else {
+		data->request = SEM_DOWN;
+	}
+
+	ret = msgSend(sem->named.port, &msg);
+	if (ret == EOK) {
+		ret = msg.o.err;
+	}
+
+	return ret;
+}
+
+
+static int sem_msgCreate(const char *name, int value, oid_t *oid)
+{
+	int ret = EOK;
+	oid_t semctl;
+	msg_t msg = { 0 };
+
+	ret = lookup(SEMCTL_PATH, NULL, &semctl);
+	if (ret != EOK) {
+		return ret;
+	}
+
+	msg.type = mtCreate;
+	msg.oid = semctl;
+	msg.i.create.type = value;
+	msg.i.create.mode = 0;
+	msg.i.data = name;
+	msg.i.size = strlen(name) + 1;
+
+	ret = msgSend(semctl.port, &msg);
+	if (ret == EOK) {
+		ret = msg.o.err;
+	}
+
+	if (oid != NULL) {
+		*oid = msg.o.create.oid;
+	}
+
+	return ret;
+}
+
+
+static int sem_lookup(const char *name, oid_t *sem)
+{
+	int ret = EOK;
+	DIR *dirp;
+	struct dirent *dent;
+	oid_t oid;
+	char *path;
+
+	dirp = opendir(SEMAPHORE_PATH);
+	if (dirp == NULL) {
+		/* posixsrv has not yet initialized the semaphore subsystem */
+		return -ENODEV;
+	}
+
+	path = malloc(PATH_MAX + 1);
+	if (path == NULL) {
+		return -ENOMEM;
+	}
+
+	if (*name == '/') {
+		name++;
+	}
+
+	memset(path, 0, PATH_MAX + 1);
+	for (;;) {
+		errno = 0;
+		dent = readdir(dirp);
+		if (dent == NULL) {
+			closedir(dirp);
+			ret = -ENOENT;
+			break;
+		}
+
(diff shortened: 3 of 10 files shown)


posix/sem.c
```diff
@@ -0,0 +1,453 @@
...
+}
+
+
+static int _sem_create(const char *name, int value, oid_t *oid)
+{
+	int ret = EOK;
+	oid_t semctl;
+	msg_t msg;
+
+	ret = lookup(SEMCTL_PATH, NULL, &semctl);
```
julianuziemblo: is it created by the kernel and thst's why we assume it's always there? maybe add such comment stating this here?

posix/sem.c
```diff
@@ -0,0 +1,453 @@
...
+}
+
+
+static int _sem_lookup(const char *name, oid_t *sem)
+{
+	int ret = EOK;
+	DIR *dirp;
+	struct dirent *dent;
+	oid_t oid;
+	char path[PATH_MAX];
```
julianuziemblo: seems like a big thing for the stack, especially on smaller targets. Is backing this up by a static field possible (calling this from 2 different threads could potentially break things) or should memory be allocated?

include/sys/semaphore.h
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+#include <limits.h>
+#include <sys/ioctl.h>
+
+#define SEMAPHORE_PATH      ("/dev/posix/sem/")
+#define SEMCTL_PATH         ("/dev/posix/semctl")
+#define SEMAPHORE_MAX_COUNT (128)
+#define SEMAPHORE_NAME_MAX  (NAME_MAX - sizeof(SEMAPHORE_PATH) - 1)
+#define SEM_VALUE_MAX       INT_MAX
```
Darchiv: 1. I think this should be defined per-architecture in libphoenix/include/arch/*/limits.h (like `NAME_MAX`). It can be just `#define SEM_VALUE_MAX INT_MAX`, but the important part is that the definition should be available via limits.h.

2. Is `SEMAPHORE_MAX_COUNT` used anywhere? Is should be `SEM_NSEMS_MAX` and defined in limits.h like in the point above. Please actually enforce the limit. Note that `_POSIX_SEM_NSEMS_MAX == 256` which is the minimum value for this limit.

2. See `sysconf()`. You could implement `_SC_SEM_VALUE_MAX ` and `_SC_SEM_NSEMS_MAX`.

3. Also, this is a good place to static-assert `SEM_VALUE_MAX` with `_POSIX_SEM_VALUE_MAX` and `SEM_NSEMS_MAX` with `_POSIX_SEM_NSEMS_MAX` in case someone changes limits.h incorrectly. Maybe also check whether the limit does not exceed the semaphore value type size (`unsigned int`).

posix/sem.c
```diff
@@ -0,0 +1,453 @@
...
+
+	if (name != NULL) {
+		if (strlen(name) > SEMAPHORE_NAME_MAX) {
+			SET_ERRNO(-EINVAL);
+			return SEM_FAILED;
+		}
+		else {
+		}
+
+		sem = (sem_t *)malloc(sizeof(*sem));
```
Darchiv: Type cast redundant

include/sys/semaphore.h
oI0ck: Applied

include/sys/semaphore.h
```diff
@@ -0,0 +1,28 @@
...
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <limits.h>
+#include <sys/ioctl.h>
+
+#define SEMAPHORE_PATH      ("/dev/posix/sem/")
+#define SEMCTL_PATH         ("/dev/posix/semctl")
```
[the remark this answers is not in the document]
oI0ck: Applied

oI0ck:
This is currently blocked by:
 - https://github.com/phoenix-rtos/libphoenix/pull/486
 - https://github.com/phoenix-rtos/phoenix-rtos-filesystems/pull/166

oI0ck:
CI shows problems with this change, I'm converting it to a draft until I am sure that it is in mergable state.

posix/sem.c
```diff
@@ -0,0 +1,422 @@
...
+		if (dent == NULL) {
+			closedir(dirp);
+			ret = -ENOENT;
+			break;
+		}
+
+		if (strcmp(name, dent->d_name) == 0) {
+			strcpy(path, SEMAPHORE_PATH);
+			strncpy(path + strlen(SEMAPHORE_PATH), dent->d_name, PATH_MAX - strlen(SEMAPHORE_PATH));
+			lookup(path, NULL, &oid);
```
julianuziemblo: lookup could have failed here, leading to uninitialized dev oid

posix/sem.c
```diff
@@ -0,0 +1,435 @@
...
+	if (dirp != NULL) {
+		for (;;) {
+			errno = 0;
+			dent = readdir(dirp);
+			if (dent != NULL) {
+				sz = min(strlen(name), dent->d_namlen);
+				if (strncmp(name, dent->d_name, sz) == 0) {
+					memset(path, 0, sizeof(path));
+					strcpy(path, SEMAPHORE_PATH);
+					strcpy(path + strlen(SEMAPHORE_PATH), dent->d_name);
```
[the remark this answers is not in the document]
oI0ck: done

include/semaphore.h
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <time.h>
+#include <limits.h>
+#include <sys/threads.h>
+#include <sys/semaphore.h>
+
+#define SEM_FAILED ((sem_t *)0xDAAB0000)
```
julianuziemblo: why this value in particular?

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+{
+	int ret = EOK;
+	DIR *dirp;
+	struct dirent *dent;
+	oid_t oid;
+	char *path;
+
+	dirp = opendir(SEMAPHORE_PATH);
+	if (dirp == NULL) {
+		/* posixsrv has yet not initialized the semaphore subsystem */
```
julianuziemblo: nitpick: `not yet`

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+	sem_t *sem;
+
+	(void)mode;
+
+	if (name == NULL) {
+		SET_ERRNO(-EINVAL);
+		return SEM_FAILED;
+	}
+
+	sem = (sem_t *)malloc(sizeof(*sem));
```
julianuziemblo: nitpick: unnecessary cast

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+
+sem_t *sem_open(const char *name, int oflag, ...)
+{
+	va_list list;
+	int err = EOK;
+	mode_t mode = 0;
+	unsigned int value = 0;
+	sem_t *sem;
+
+	(void)mode;
```
julianuziemblo, line 178: does POSIX not enforce passing mode along? or we just don't want to handle it for now? in the second case, add a TODO here.

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+
+	if (name == NULL) {
+		SET_ERRNO(-EINVAL);
```
julianuziemblo: `errno = EINVAL` for cases like this is more idiomatic in libphoenix

julianuziemblo: POSIX specifies EINVAL only when `The sem_open() operation is not supported for the given name, or O_CREAT was specified in oflag and value was greater than {SEM_VALUE_MAX}.`. `NULL + 1` would still crash, so I wouldn't handle this case, as this is still userland.

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+
+	sem = (sem_t *)malloc(sizeof(*sem));
+	if (sem == NULL) {
+		SET_ERRNO(-ENOMEM);
+		return SEM_FAILED;
+	}
+
+	sem->type = smNamed;
+
+	if ((oflag & O_CREAT) > 0) {
```
julianuziemblo: we checks flags with `(val & flag) != 0` (or `(val & flag) == 0`)

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+			err = sem_msg_create(name, value, &sem->named);
+		}
+	}
+
+	if (err != EOK) {
+		SET_ERRNO(err);
+		free(sem);
+		return SEM_FAILED;
+	}
+	else {
```
julianuziemblo: nitpick: not needed as we `return` in the `if` branch

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+		ret = -EINVAL;
+	}
+
+	return SET_ERRNO(ret);
+}
+
+
+int sem_timedwait(sem_t *restrict sem, const struct timespec *restrict abs_timeout)
```
julianuziemblo: nitpick: `abs_timeout` -> `abstimeout` to conform with our requirements (as the name does not matter to POSIX I suppose)

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+#include <limits.h>
+#include <sys/time.h>
+#include <sys/msg.h>
+#include <sys/types.h>
+#include <sys/minmax.h>
+#include <sys/semaphore.h>
+#include <phoenix/ioctl.h>
+
+
+static int sem_msg_open(sem_t *sem)
```
julianuziemblo: nitpick: names should be `sem_msgOpen`, `sem_msgDown`, etc.

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+		return SET_ERRNO(-EINVAL);
+	}
+
+	if (sem->type == smNamed) {
+		msg.type = mtGetAttr;
+		msg.oid = sem->named;
+		ret = msgSend(sem->named.port, &msg);
+		if (ret == EOK && value != NULL) {
+			*value = msg.o.attr.val;
+		}
```
julianuziemblo: I think this should be an ioctl (mtDevctl), as we use mtGetAttr for `proc_size` in kernel, and this would report the value as the size of the file

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+	msg_t msg = { 0 };
+
+	if (sem == NULL) {
+		return SET_ERRNO(-EINVAL);
+	}
+
+	if (sem->type == smNamed) {
+		msg.type = mtClose;
+		msg.oid = sem->named;
+		ret = msgSend(sem->named.port, &msg);
```
julianuziemblo, line 403: `if (ret == 0) ret = msg.o.err` probably

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+		return SET_ERRNO(-ENOSYS);
+	}
+
+	if (sem == NULL) {
+		return SET_ERRNO(-EINVAL);
+	}
+
+	new.type = smUnnamed;
+	ret = semaphoreCreate(&new.unnamed, value);
+	*sem = new;
```
julianuziemblo, line 436: nitpick: this is set unconditionally. What is `semaphoreCreate` failed?

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+{
+	int ret = EOK;
+	oid_t oid;
+	msg_t msg = { 0 };
+
+	ret = sem_lookup(name, &oid);
+	if (ret == EOK) {
+		msg.type = mtDestroy;
+		msg.oid = oid;
+		ret = msgSend(oid.port, &msg);
```
julianuziemblo: `if (ret == 0) ret = msg.o.err` probably

posix/utils.c
```diff
@@ -5,33 +5,55 @@
...
 	}
 	else {
-		*dir = path;
-		*base = slash + 1;
+		if (dir != NULL) {
+			*dir = path;
+		}
+
+		if (base != NULL) {
+			*base = slash + 1;
```
julianuziemblo: but.. why?

posix/utils.c
```diff
@@ -5,33 +5,55 @@
...
 
 void splitname(char *path, char **base, char **dir)
 {
 	char *slash;
 
+	if (path == NULL) {
+		errno = EINVAL;
+		return;
+	}
+
```
julianuziemblo: this seems especially wrong as now we could silently change errno but return 0 (from the calling function).

include/semaphore.h
oI0ck: Why not?

julianuziemblo: are you sure this cannot be an allocated address on any of our targets?

posix/sem.c
oI0ck: This is technically not wrong, we should not support passing `NULL` as `name`. Even if this gets rejected by the server later on, we can catch that before making a round trip for an obviously erroneous parameter.

posix/sem.c
oI0ck: Right, I got TOCTOU'd, nice catch.

posix/sem.c
oI0ck: Will change.

On a side note, I don't think that POSIX enforces variable names on the implementation. FreeBSD uses `abs_timeout` and is compliant.

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+	for (;;) {
+		errno = 0;
+		dent = readdir(dirp);
+		if (dent == NULL) {
+			closedir(dirp);
+			ret = -ENOENT;
+			break;
+		}
+
+		if (strcmp(name, dent->d_name) == 0) {
```
ziemleszcz: Leading `/` is not stripped from `name` as it in `semaphore_create()` from phoenix-rtos-posixsrv.

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+
+	if (sem == NULL) {
+		return SET_ERRNO(-EINVAL);
+	}
+
+	gettime(&now, &offs);
+	now += offs;
+
+	timeout += abs_timeout->tv_sec * 1000000 + abs_timeout->tv_nsec / 1000;
+	timeout -= now;
```
ziemleszcz, line 293: Note that with absolute timeout this could be negative value (timeout could be in the past) which is not handled correctly by `sem_msg_down` (results in unbounded down).

posix/sem.c
```diff
@@ -0,0 +1,421 @@
...
+
+	if (sem->type == smNamed) {
+		msg.type = mtClose;
+		msg.oid = sem->named;
+		ret = msgSend(sem->named.port, &msg);
+	}
+	else {
+		ret = -EINVAL;
+	}
+
```
ziemleszcz, line 412: `sem` memory is not freed.

include/semaphore.h
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * POSIX implementation - semaphores
+ *
+ * Copyright 2026 Phoenix Systems
+ * Author: Michal Lach
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <time.h>
```
ziemleszcz, line 17: Missing C++ support (`extern "C`).

include/sys/semaphore.h
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * POSIX implementation - semaphores
+ *
+ * Copyright 2026 Phoenix Systems
+ * Author: Michal Lach
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <assert.h>
```
ziemleszcz, line 17: Missing C++ support (`extern "C`).

posix/sem.c
oI0ck: `create_dev()` does not take a `mode` parameter, there is a `TODO` in there to add it. Devices are created by default with `0666` mode.

I can leave a comment and revisit this after this in another PR.

posix/sem.c
oI0ck: Will change this to returning `-ENOSPC` if `semaphoreCreate()` fails. Seems like the most appropriate `errno` that is compilant with POSIX.

include/semaphore.h
oI0ck: Will change to `NULL`

posix/utils.c
oI0ck: I wanted to make `dir` and `base` optional if you only want one of these. Since now I use `destroy_dev()`, I can drop this commit

julianuziemblo: I'd drop it since it somewhat obfuscates the function, but that's not a bad idea IMO

posix/sem.c
```diff
@@ -0,0 +1,458 @@
...
+	gettime(&now, &offs);
+	now += offs;
+
+	timeout += abstime->tv_sec * 1000000 + abstime->tv_nsec / 1000;
+	timeout -= now;
+
+	if (timeout < 0) {
+		return SET_ERRNO(-EINVAL);
```
ziemleszcz: Try to acquire the semaphore, return 0 on success and `ETIMEDOUT` on failure.

posix/sem.c
```diff
@@ -0,0 +1,458 @@
...
+	}
+
+	gettime(&now, &offs);
+	now += offs;
+
+	timeout += abstime->tv_sec * 1000000 + abstime->tv_nsec / 1000;
+	timeout -= now;
+
+	if (timeout < 0) {
```
ziemleszcz: `timeout <= 0` (currently it blocks indefinitely on 0).

posix/sem.c
```diff
@@ -0,0 +1,458 @@
...
+	time_t timeout = 0, now = 0, offs = 0;
+
+	if (sem == NULL) {
+		return SET_ERRNO(-EINVAL);
+	}
+
+	gettime(&now, &offs);
+	now += offs;
+
+	timeout += abstime->tv_sec * 1000000 + abstime->tv_nsec / 1000;
```
ziemleszcz, line 292: `abstime == NULL` check and `tv_*` fields validation is missing.

posix/sem.c
```diff
@@ -0,0 +1,458 @@
...
+			if (value != NULL) {
+				*value = v;
+			}
+		}
+	}
+	else if (sem->type == smUnnamed) {
+		ret = semaphoreCount(&sem->unnamed);
+		if (ret > 0 && value != NULL) {
+			ret = EOK;
+			if (value != NULL) {
```
ziemleszcz: Redundant check.

posix/sem.c
```diff
@@ -0,0 +1,458 @@
...
+				*value = ret;
```
ziemleszcz: `value` is always  set to 0 (see L364).

posix/sem.c
oI0ck: I don't understand what do you mean by that. Isn't this what happens here?

ziemleszcz: A negative timeout should have `sem_trywait()` semantics, except that it should return `ETIMEDOUT` instead of `EAGAIN` when the semaphore cannot be acquired. Your code simply returns `EINVAL`.

oI0ck: Huh, right, I guess I could add a `sem_trywait()` call in here.

ziemleszcz:
Please rebase thsi PR and resolve conficts.

posix/sem.c
```diff
@@ -0,0 +1,471 @@
...
+	if (sem == NULL) {
+		return SET_ERRNO(-EINVAL);
+	}
+
+	if (sem->type == smNamed) {
+		msg.type = mtDevCtl;
+		msg.oid = sem->named;
+		data = (ioctl_in_t *)msg.i.raw;
+		data->request = SEM_GETVALUE;
+		ret = msgSend(sem->named.port, &msg);
```
ziemleszcz: This hand-crafted semi-ioctl must set `data->size` correctly to `IOCPARM_LEN(SEM_GETVALUE)`, as posix-srv uses `ioctl_setResponse()`. Apply the same fix to the other locations in this file.

PR #466, opened 2026-03-16
Add destroy_dev() function as a counterpart to create_dev()

julianuziemblo:
## Motivation and Context
Depends-On: phoenix-rtos-kernel:julianuziemblo/port-unregister
Depends-On: phoenix-rtos-tests:julianuziemblo/CI-651

## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/471
- [x] Tested by hand on: `armv7m7-imxrt117x-evk`, `ia32-generic-qemu`, `armv7a7-imx6ull-evk`

## Checklist:

## Special treatment

    - https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/750

### DIFF
--- a/sys/msg.c
+++ b/sys/msg.c
@@ -1,0 +1,31 @@
+/*
+ * Phoenix-RTOS
+ *
+ * libphoenix
+ *
+ * sys/msg
+ *
+ * Copyright 2026 Phoenix Systems
+ * Author: Julian Uziembło
+ *
+ * %LICENSE%
+ */
+
+#include "../common/util.h"
+
+
+extern int sys_portRegister(uint32_t port, const char *name, size_t len, oid_t *oid);
+
+extern int sys_portUnregister(const char *name, size_t len);
+
+
+int portRegister(uint32_t port, const char *name, oid_t *oid)
+{
+	return sys_portRegister(port, name, __strSizeNull(name), oid);
+}
+
+
+int portUnregister(const char *name)
+{
+	return sys_portUnregister(name, __strSizeNull(name));
+}
--- a/unistd/file.c
+++ b/unistd/file.c
@@ -634,25 +634,42 @@ int destroy_dev(const char *path)
 
 	oid_t oid, odev;
 	int retry = 0, err;
-	char *canonical_path, *dir, *sep;
+	char *canonical_path, *dir, *sep, *tpathalloc = NULL;
+	const char *tpath = path;
 
 	if (path == NULL) {
 		return -EINVAL;
 	}
 
 	while (lookup("devfs", NULL, &odev) < 0) {
-		if (++retry < 3 || nodevfs != 0) {
+		/* if destroy_dev() is called by anyone started from syspage devfs
+		 * may be not registered yet so we try 3 times until we give up */
+		if (++retry > 3 || nodevfs != 0) {
 			nodevfs = 1;
 
 			if (lookup("/dev", NULL, &odev) < 0) {
-				return -ENOSYS;
-			}
-			else {
-				break;
-			}
-		}
+				/* Looks like we don't have a filesystem.
+				 * Fall back to portUnregister. */
+				if (*path != '/') {
+					/* Move point to /dev */
+					tpathalloc = malloc(strlen(path) + 6);
+					if (tpathalloc == NULL) {
+						return -ENOMEM;
+					}
+					strcpy(tpathalloc, "/dev/");
+					strcpy(tpathalloc + 5, path);
+					tpath = tpathalloc;
+				}
 
-		usleep(100000);
+				err = portUnregister(tpath);
+				free(tpathalloc);
+				return err;
+			}
+			break;
+		}
+		else {
+			usleep(100000);
+		}
 	}
 
 	if (strncmp("/dev/", path, 5) == 0) {
--- a/unistd/file.c
+++ b/unistd/file.c
@@ -628,6 +628,111 @@ int create_dev(oid_t *oid, const char *path)
 }
 
 
+int destroy_dev(const char *path)
+{
+	static int nodevfs = 0;
+
+	oid_t oid, odev;
+	int retry = 0, err;
+	char *canonical_path, *dir, *sep;
+
+	if (path == NULL) {
+		return -EINVAL;
+	}
+
+	while (lookup("devfs", NULL, &odev) < 0) {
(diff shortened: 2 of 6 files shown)


nalajcie:
I would rather see `create_dev` rewritten first. IMHO this function should not try to create `/dev` dir, also should either accept `/dev/devname` or `devname` param, not try to handle all cases. Probably some other thing should also be straightened out here.

julianuziemblo:
> I would rather see `create_dev` rewritten first. IMHO this function should not try to create `/dev` dir, also should either accept `/dev/devname` or `devname` param, not try to handle all cases. Probably some other thing should also be straightened out here.

I agree with your suggestions, but it seems like they should he handled in separate PR as they are big and very breaking. This PR only adds missing functionality as a basis for drivers that need it now to work without creating excessive overhead (e.g. spawning cleanup threads just to take care of deleting their own devs).

unistd/file.c
```diff
@@ -628,6 +628,80 @@ int create_dev(oid_t *oid, const char *path)
...
+				/* Looks like we don't have a filesystem.
+				 * Fall back to portUnregister. */
+				if (*path != '/') {
+					/* Move point to /dev */
+					tpathalloc = malloc(strlen(path) + 6);
+					if (tpathalloc == NULL) {
+						return -ENOMEM;
+					}
+					strcpy(tpathalloc, "/dev/");
+					strcpy(tpathalloc + 5, path);
```
agkaminski: `strcat`?

agkaminski (commented):
One nitpick, otherwise LGTM

unistd/file.c
julianuziemblo: yeah, probably - this was just straight up copied (excluding creating `/dev`) from `create_dev()` - I'll look into it

sys/msg.c
```diff
@@ -0,0 +1,33 @@
...
+#include <string.h>
+#include <sys/msg.h>
+
+
+extern int sys_portRegister(uint32_t port, const char *name, size_t len, oid_t *oid);
+
+
+int portRegister(uint32_t port, const char *name, oid_t *oid)
+{
+	return sys_portRegister(port, name, (name == NULL) ? 0 : strlen(name), oid);
```
agkaminski: If we want to do these checks its better to create helper `strlen` variant that handles NULL

julianuziemblo: something like this?

```c
static inline size_t msg_strlenCheckNull(const char *str) {
    return (str == NULL) ? 0 : strlen(str);
}
```

agkaminski: Yes, but with name along the lines of `__strlenNull` or something like this. It will be needed many syscalls, so no need to reduce it to be part of `msg_` only.

julianuziemblo: Okay, so maybe place it in some "global" header then? Do we have some "common" header in libphoenix?

agkaminski: Don't think so. E.g. musl has an entire mini lib for such stuff

julianuziemblo: Added, lmk what you think

agkaminski: LGTM, but perhaps move it to the place it'll be available? The goal is that we won't have to do that for every syscall separately

julianuziemblo: transferred it to new `common/util.h` file

agkaminski (dismissed):

unistd/file.c
```diff
@@ -628,6 +628,120 @@ int create_dev(oid_t *oid, const char *path)
...
+	int retry = 0, err;
+	char *canonical_path, *dir, *sep, *tpathalloc = NULL;
+	const char *tpath = path;
+
+	if (path == NULL) {
+		return -EINVAL;
+	}
+
+	while (lookup("devfs", NULL, &odev) < 0) {
+		/* if remove_dev() is called by anyone started from syspage devfs
```
ziemleszcz: /* if destroy_dev() is called

unistd/file.c
```diff
@@ -628,6 +628,120 @@ int create_dev(oid_t *oid, const char *path)
...
+				/* Looks like we don't have a filesystem.
+				 * Fall back to portUnregister. */
+				if (*path != '/') {
+					/* Move point to /dev */
+					tpathalloc = malloc(strlen(path) + 6);
+					if (tpathalloc == NULL) {
+						return -ENOMEM;
+					}
+					strcpy(tpathalloc, "/dev/");
+					strcat(tpathalloc, path);
```
ziemleszcz: strcpy(tpathalloc + 5, path)?

unistd/file.c
```diff
@@ -628,6 +628,120 @@ int create_dev(oid_t *oid, const char *path)
...
+		canonical_path = strdup(path);
+	}
+	else {
+		size_t len = 5 + strlen(path) + 1;
+		canonical_path = malloc(len);
+		snprintf(canonical_path, len, "/dev/%s", path);
+	}
+
+	if (lookup(canonical_path, NULL, &oid) < 0) {
+		return -ENODEV;
```
ziemleszcz, line 692: Memory leak (canonical_path).

unistd/file.c
```diff
@@ -628,6 +628,120 @@ int create_dev(oid_t *oid, const char *path)
...
+		.type = mtUnlink,
+		.oid = odev,
+		.i.ln.oid = oid,
+		.i.data = path,
+		.i.size = strlen(path) + 1,
+	};
+
+	err = msgSend(odev.port, &msg);
+	if (err < 0) {
+		return err;
```
ziemleszcz, line 746: Memory leak (canonical_path).

unistd/file.c
```diff
@@ -628,6 +628,120 @@ int create_dev(oid_t *oid, const char *path)
...
+			usleep(100000);
+		}
+	}
+
+	if (strncmp("/dev/", path, 5) == 0) {
+		canonical_path = strdup(path);
+	}
+	else {
+		size_t len = 5 + strlen(path) + 1;
+		canonical_path = malloc(len);
```
ziemleszcz, line 683: Missing NULL check.

unistd/file.c
julianuziemblo: @agkaminski suggested strcat here - why do you prefer strcpy, because we know the length?

ziemleszcz: Yes, it's faster.

ziemleszcz (approved):

PR #469, opened 2026-04-16
socket: implement `accept()` with `syscalls_sys_accept4`

xvuko:
This will allow to deprecate/remove `accept` syscall.

TASK: RTOS-1289

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/include/sys/socket.h
+++ b/include/sys/socket.h
@@ -71,7 +71,6 @@ int connect(int socket, const struct sockaddr *address, socklen_t address_len);
 int bind(int socket, const struct sockaddr *address, socklen_t address_len);
 int listen(int socket, int backlog);
 int accept4(int socket, struct sockaddr *address, socklen_t *address_len, int flags);
-int accept(int socket, struct sockaddr *address, socklen_t *address_len);
 ssize_t sendto(int socket, const void *message, size_t length, int flags, const struct sockaddr *dest_addr, socklen_t dest_len);
 ssize_t sendmsg(int socket, const struct msghdr *msg, int flags) __attribute__((warning("sendmsg() is not fully supported")));
 ssize_t recvfrom(int socket, void *message, size_t length, int flags, struct sockaddr *src_addr, socklen_t *src_len);
@@ -82,6 +81,12 @@ int getsockopt(int socket, int level, int optname, void *optval, socklen_t *optl
 int setsockopt(int socket, int level, int optname, const void *optval, socklen_t optlen);
 int shutdown(int socket, int how);
 int socketpair(int domain, int type, int protocol, int socket_vector[2]);
+
+
+inline int accept(int socket, struct sockaddr *address, socklen_t *address_len)
+{
+	return accept4(socket, address, address_len, 0);
+}
 
 
 inline ssize_t send(int socket, const void *message, size_t length, int flags)
--- a/include/sys/socket.h
+++ b/include/sys/socket.h
@@ -72,10 +72,8 @@ int bind(int socket, const struct sockaddr *address, socklen_t address_len);
 int listen(int socket, int backlog);
 int accept4(int socket, struct sockaddr *address, socklen_t *address_len, int flags);
 int accept(int socket, struct sockaddr *address, socklen_t *address_len);
-ssize_t send(int socket, const void *message, size_t length, int flags);
 ssize_t sendto(int socket, const void *message, size_t length, int flags, const struct sockaddr *dest_addr, socklen_t dest_len);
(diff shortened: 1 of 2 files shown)


include/sys/socket.h
```diff
@@ -85,6 +84,10 @@ int setsockopt(int socket, int level, int optname, const void *optval, socklen_t
 int shutdown(int socket, int how);
 int socketpair(int domain, int type, int protocol, int socket_vector[2]);
 
+inline int accept(int socket, struct sockaddr *address, socklen_t *address_len)
+{
+	return accept4(socket, address, address_len, 0);
+}
```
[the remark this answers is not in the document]
xvuko: Added commit that changes send/recv to inline functions too.

sys/socket.c
```diff
@@ -53,6 +53,9 @@ extern ssize_t sys_sendmsg(int socket, const struct msghdr *msg, int flags);
 
 int h_errno;
 
+/* defined in sys/socket.h */
+extern inline int accept(int socket, struct sockaddr *address, socklen_t *address_len);
```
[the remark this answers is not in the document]
xvuko: We are not targeting C versions before C99

include/sys/socket.h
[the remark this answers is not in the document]
xvuko: For small targets code size (including relocations) and stack usage need to be taken into consideration.

include/sys/socket.h
```diff
@@ -86,6 +83,23 @@ int shutdown(int socket, int how);
...
+
+inline ssize_t send(int socket, const void *message, size_t length, int flags)
+{
+	return sendto(socket, message, length, flags, NULL, 0);
+}
+
+
+inline ssize_t recv(int socket, void *message, size_t length, int flags)
+{
+	return recvfrom(socket, message, length, flags, NULL, 0);
```
julianuziemblo: nit from previous implementation: last argument is a pointer, so NULL could be more readable

julianuziemblo (commented):
while at it, we could potentially just remove `syscalls_sys_accept`, as we're still before release. @Darchiv what do you think?

include/sys/socket.h
xvuko: Fixed

include/sys/socket.h
```diff
@@ -86,6 +83,23 @@ int shutdown(int socket, int how);
 int socketpair(int domain, int type, int protocol, int socket_vector[2]);
 
 
+inline int accept(int socket, struct sockaddr *address, socklen_t *address_len)
```
ziemleszcz, line 86: With `static inline`, the `extern inline` declarations in sys/socket.c are probably no longer necessary.

Darchiv:
> while at it, we could potentially just remove `syscalls_sys_accept`, as we're still before release. @Darchiv what do you think?

Seems sane. Remove

include/sys/socket.h
xvuko: Those are intentionally non-static inlines

ziemleszcz: Could you clarify the intention here?

xvuko: `accept` is a part of official POSIX API. It would be good to ensure that compiled library will provide symbol for this function. One of the cases where this would be needed is linking with assembly code. Using non-static inline guarantees that global symbol is created, but compiler can choose to inline.

ziemleszcz: Understood. Note that several other POSIX functions are already using `static inline`. Could you update those as well?

xvuko: Added separate issue https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1722

ziemleszcz (approved):

julianuziemblo:
> > while at it, we could potentially just remove `syscalls_sys_accept`, as we're still before release. @Darchiv what do you think?
> 
> Seems sane. Remove

@xvuko are you going to do this one too?

nalajcie:
Idea: maybe we could write `extern inline` in cases like those - to make the intent evident? That way nobody would try to "fix it" making it static in the future?

The default linkage is extern, so the keyword has no other side effects.

xvuko:
> Idea: maybe we could write `extern inline` in cases like those - to make the intent evident?

We can not do that. In C99 adding `extern inline` to header will create global symbol in all compilation units leading to linking error. This is why `extern inline` declarations are added to `sys/socket.c` and only there.

From  ISO/IEC 9899:1999 (E),  §6.7.4.6:
> If all of the file scope declarations for a function in a translation unit include the **inline function specifier without `extern`**, then the definition in that translation unit is an inline definition. An inline definition does not provide an external definition for the function, and does not forbid an external definition in another translation unit.

PR #470, opened 2026-04-21
unistd/sys.c: fix execle to populate envp from independent va_list

### LINKED ISSUE #1457: `execle()` can trigger page fault in kernel
Image

Image

Reproduction (ia32-generic-qemu):
```c
#include <stdio.h>
#include <unistd.h>
#include <errno.h>

int main(void)
{
        int ret;

        ret = execle("/bin/ls", "ls", "-l", NULL, NULL);
        printf("execle ret: %d errno: %d\n", ret, errno);

        return 0;
}
```

**NOTE:** When I used `execle("/bin/ls", "ls", NULL, NULL);`, it worked fine

### LINKED ISSUE #1465: `execle()` doesn't pass environment
Reproduction (ia32-generic-qemu):

First program:
```c
#include <stdio.h>

extern char **environ;

int main(void)
{
        int i = 1;
        char *var;

        puts("Printing environment...");

        while ((var = *environ++)) {
                printf("%d: %s\n", i++, var);
        }

        return 0;
}
```

Second program:
```c
#include <stdio.h>
#include <errno.h>
#include <unistd.h>

int main(void)
{
        int ret;
        char *env[] = {
                "USER=adam",
                "SHELL=bash",
                "HOME=/home/adam",
                "PATH=/usr/bin:/bin",
                "EDITOR=nano",
                NULL
        };

        errno = 0;
        ret = execle("/usr/bin/print_env", "print_env", NULL, env);
        printf("execv ret: %d errno: %d\n", ret, errno);

        return 1;
}
```

Image

adamgreloch:
Previous implementation seemed to make non-portable assumptions about va_list reuse/consumption in the callee. As a result

```
execle("/bin/ls", "ls", "-l", NULL, NULL);
```

would cause `envp` to point to garbage on riscv64.

This change uses fresh va_list to ensure safe access to both handles.

Fixes: phoenix-rtos/phoenix-rtos-project#1457
Fixes: phoenix-rtos/phoenix-rtos-project#1465

JIRA: RTOS-1295

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [X] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/481

## Checklist:

## Special treatment

### DIFF
--- a/unistd/sys.c
+++ b/unistd/sys.c
@@ -226,10 +226,15 @@ static char **argv_gather(const char *arg, va_list args)
 		return NULL;
 
 	argv[argc++] = (char *)arg;
+	if (arg == NULL) {
+		return argv;
+	}
 
 	while ((argv[argc++] = va_arg(args, char *)) != NULL) {
 		if (argc == max) {
-			if ((res = realloc(argv, max *= 2)) == NULL) {
+			max *= 2;
+			res = realloc(argv, max * sizeof(char *));
+			if (res == NULL) {
 				free(argv);
 				return NULL;
 			}
--- a/unistd/sys.c
+++ b/unistd/sys.c
@@ -287,6 +287,14 @@ int execle(const char *path, const char *arg, ...)
 
 	va_start(args, arg);
 	argv = argv_gather(arg, args);
+	va_end(args);
+
+	/* reinitialize args as it could have been invalidated by (partial) consumption by argv_gather */
+	va_start(args, arg);
+	const char *p = arg;
+	while (p != NULL) {
+		p = va_arg(args, const char *);
+	}
 	envp = va_arg(args, char **);
 	va_end(args);
 


unistd/sys.c
```diff
@@ -281,12 +286,22 @@ int execlp(const char *file, const char *arg, ...)
...
 {
-	va_list args;
+	va_list args, argsCopy;
 	char **argv, **envp;
 	int err;
 
 	va_start(args, arg);
-	argv = argv_gather(arg, args);
+
+	/* copy args to argsCopy and gather from it so that args is left untouched for envp */
```
etiaro: I would rather see in this comment that va_list gets invalidated when (even partially) consumed  inside callee, to follow the convention of "why" instead of "what" comments.

etiaro (commented):
Generally fine, just a little comment suggestion.

Also, you could consider just using va_start/va_end twice, I guess that's a little cleaner..

unistd/sys.c
adamgreloch: good suggestion, thanks

etiaro (approved):
LGTM

PR #473, opened 2026-05-08
stdio: fix fgets(), add gets()

### LINKED ISSUE #1489: `fgets()` behavior and implementation diverges from ISO C standard
According to ISO C standard:
https://pubs.opengroup.org/onlinepubs/009696699/functions/fgets.html
The fgets() function shall read bytes from stream into the array pointed to by s until n-1 bytes are read, or a <newline> is read and transferred to s, or an end-of-file condition is encountered. A null byte shall be written immediately after the last byte read into the array. If the end-of-file condition is encountered before any bytes are read, the contents of the array pointed to by s shall not be changed.

Meanwhile, in libphoenix `fgets()` behavior is divergent for n=1. The function reads single byte, which is outcome of implementation https://github.com/phoenix-rtos/libphoenix/blob/3f0d10611013ed992fb351d0e039a8e6655aa647/stdio/file.c#L736

Functionality was tested with the following code:
```c
#include <stdio.h>

int main(void)
{
	char buff[2];

	fgets(buff, 1, stdin);
	printf("output: %u\t%u\n", buff[0], buff[1]);

	return 0;
}
```

julianuziemblo:
Fixes: phoenix-rtos/phoenix-rtos-project#1489

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `ia32-generic-qemu`

## Checklist:

- [x] I have added tests to cover my changes: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/479

## Special treatment

### DIFF
--- a/stdio/file.c
+++ b/stdio/file.c
@@ -734,20 +734,35 @@ int fputc(int c, FILE *stream)
 
 char *fgets_unlocked(char *str, int n, FILE *stream)
 {
+	if (n < 1) {
+		return NULL;
+	}
+
+	if (n == 1) {
+		str[0] = '\0';
+		return str;
+	}
+
 	int c, i = 0;
-	while ((c = fgetc_unlocked(stream)) != EOF) {
-		str[i++] = c;
-		if (c == '\n' || i == n - 1) {
+
+	while (i < n - 1) {
+		c = getc_unlocked(stream);
+		if (c == EOF) {
+			if (i == 0) {
+				return NULL;
+			}
+			break;
+		}
+
+		str[i] = c;
+		i++;
+
+		if (c == '\n') {
 			break;
 		}
 	}
 
-	if (i) {
-		str[i] = 0;
-	}
-	else {
-		return NULL;
-	}
+	str[i] = '\0';
 
 	return str;
 }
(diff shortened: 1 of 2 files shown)


stdio/file.c
```diff
@@ -732,18 +732,26 @@ int fputc(int c, FILE *stream)
 }
 
 
-char *fgets_unlocked(char *str, int n, FILE *stream)
+/* if n < 1 -> no limit */
+static char *_fgets(char *str, int n, FILE *stream)
 {
-	int c, i = 0;
+	int c, i = -1;
```
nalajcie: I think we don't need to start from `-1` here.

Also: for n == 1 this function behaves incorrectly - consumes byte from the string when it shouldn't (it was previously also probably incorrect, still - you're meddling with `n` and `i` so you should spot and fix it).

Edit: you've explicitly stated that this PR fixes the `n=1` case which is not the case.

I'm against merging this until tests for corner cases are added. Start by adding tests to verify the problem exists, change PR to draft until the issues are fixed.

nalajcie (changes requested):
Adding multiple people before even the test were run (and failed)? Please be more mindful to the potential reviewers,

stdio/file.c
```diff
@@ -734,21 +734,36 @@ int fputc(int c, FILE *stream)
 
 char *fgets_unlocked(char *str, int n, FILE *stream)
 {
+	if (n < 1) {
+		return NULL;
+	}
+
+	if (n == 1) {
```
etiaro: I don't think this corner case requires separate handling here

etiaro (commented):
Should be split into 2 separate, atomic commits. Other than that, lgtm.

etiaro (approved):

PR #478, opened 2026-05-22
stdio: implement tmpnam() and tempnam() functions

julianuziemblo:
## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/490

## Checklist:

## Special treatment


stdlib/mktemp.c
```diff
@@ -119,3 +121,61 @@ char *mkdtemp(char *templt)
...
+		}
+	}
+
+	dirlen = strlen(dir);
+	pfxlen = (pfx == NULL) ? 0 : strlen(pfx);
+	if (dir[dirlen - 1] != '/') {
+		sep = 1;
+	}
+
+	char *nam = malloc(dirlen + sep + pfxlen + 6 + 1);
```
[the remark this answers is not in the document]
julianuziemblo: wanted to stay consistant with function name

stdlib/mktemp.c
```diff
@@ -119,3 +121,61 @@ char *mkdtemp(char *templt)
...
+	int err, dirlen, pfxlen, sep = 0;
+
+	if (dir == NULL || stat(dir, &statbuf) < 0 || !S_ISDIR(statbuf.st_mode)) {
+		if (access(P_tmpdir, W_OK | X_OK) < 0) {
+			dir = "/";
+		}
+		else {
+			dir = P_tmpdir;
+		}
+	}
```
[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist link me the appropriate documentation on POSIX requiring checking TMPDIR please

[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist there is no mention of `TMPDIR` on the site you link. There's not even a mention about environment. Get your facts straight.

julianuziemblo:
Drafting for now as added tests depend on tmpfile() tests, which depend on tmpfile() fixes. Will undraft after https://github.com/phoenix-rtos/phoenix-rtos-posixsrv/pull/32 is merged.

nalajcie:
if you would use the same branch name for related changes across all submodules, these changes would compile and test succesfully. 

PLEASE always use the same branch name on related changes OR use `Depends-On:` in PR description (as described in https://youtrack.phoenix-rtos.com/articles/CI-A-23070284/Github)

julianuziemblo:
> if you would use the same branch name for related changes across all submodules, these changes would compile and test succesfully.

it's not this here: as I wrote, I'm fixing tmpfile() in another branch and wanted to base tests for this PR on that branch for now as it adds "posixsrv" directory to tests, which I didn't want to duplicate across PRs. I'll undraft, rerun tests and add reviewers once that PR is merged.

stdlib/mktemp.c
```diff
@@ -119,3 +121,61 @@ char *mkdtemp(char *templt)
...
```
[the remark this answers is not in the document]
julianuziemblo: there is no mention of `TMPDIR` checking in POSIX specification that I found. It seems to be a glibc thing. Do you think its worth the hustle @ziemleszcz ?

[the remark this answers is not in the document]
julianuziemblo: (keep in mind this function is described as "obsolescent" in POSIX documentation anyway)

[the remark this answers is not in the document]
ziemleszcz: Ignore it.

include/stdio.h
```diff
@@ -47,6 +47,7 @@ extern "C" {
 #define FOPEN_MAX 16
 #define L_ctermid 11 /* Fit at least /dev/pts/XX */
 #define L_tmpnam  16
+#define P_tmpdir  "/tmp"
```
ziemleszcz: `/var/tmp` is used by tmpfile(). Also on DATARO `/var/tmp` is running on dummyfs.

include/stdio.h
```diff
@@ -194,6 +195,10 @@ extern FILE *tmpfile(void);
 extern char *tmpnam(char *str);
 
 
+/* Generates a pathname that may be used for a temporary file. */
+extern char *tempnam(const char *dir, const char *pfx);
```
ziemleszcz: `extern` is not recommended.

stdlib/mktemp.c
```diff
@@ -119,3 +121,64 @@ char *mkdtemp(char *templt)
...
+	int err;
+
+	if (str == NULL) {
+		str = template;
+	}
+
+	do {
+		(void)strcpy(str, "/tmp/fileXXXXXX");
+		err = __mktemp(str, 0);
+	} while ((err < 0) && errno == EEXIST);
```
ziemleszcz: Unbounded loop, maybe limit to `TMP_MAX` iterations. This applies to `tempnam()` as well.

stdlib/mktemp.c
ziemleszcz: Could you fix `errno == EINTR`?

stdlib/mktemp.c
```diff
@@ -119,3 +121,64 @@ char *mkdtemp(char *templt)
...
+}
+
+
+char *tempnam(const char *dir, const char *pfx)
+{
+	struct stat statbuf;
+	int err, dirlen, pfxlen, sep = 0;
+
+	if (dir == NULL || stat(dir, &statbuf) < 0 || !S_ISDIR(statbuf.st_mode)) {
+		if (access(P_tmpdir, W_OK | X_OK) < 0) {
```
ziemleszcz: Why `X_OK`? Anyway it's better to run the same check (`stat()`) for both `dir` and `P_tmpdir`. Maybe reset `dir` to to `P_tmpdir` on NULL and do the check in next step?

stdlib/mktemp.c
```diff
@@ -119,3 +121,64 @@ char *mkdtemp(char *templt)
...
+{
+	static char template[L_tmpnam];
+	int err;
+
+	if (str == NULL) {
+		str = template;
+	}
+
+	do {
+		(void)strcpy(str, "/tmp/fileXXXXXX");
```
ziemleszcz: `/tmp/fileXXXXXX` shoud be defined and it should use `P_tmpdir`. Also add static assert checking that `L_tmpnam` is large enought to hold it. Is 'fileXXXXXX' standard?

stdlib/mktemp.c
```diff
@@ -16,7 +16,9 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <stdlib.h>
+#include <stdio.h>
 #include <string.h>
+#include <sys/stat.h>
```
ziemleszcz: Duplicate.

stdlib/mktemp.c
```diff
@@ -119,3 +121,64 @@ char *mkdtemp(char *templt)
...
+		if (access(P_tmpdir, W_OK | X_OK) < 0) {
+			dir = "/";
+		}
+		else {
+			dir = P_tmpdir;
+		}
+	}
+
+	dirlen = strlen(dir);
+	pfxlen = (pfx == NULL) ? 4 /* strlen("file") */ : strlen(pfx);
```
ziemleszcz: Magic number 4. I see the comment but assume that template can change.

stdlib/mktemp.c
```diff
@@ -119,3 +121,64 @@ char *mkdtemp(char *templt)
...
+	} while ((err < 0) && errno == EEXIST);
+
+	return (err < 0) ? NULL : str;
+}
+
+
+char *tempnam(const char *dir, const char *pfx)
+{
+	struct stat statbuf;
+	int err, dirlen, pfxlen, sep = 0;
```
ziemleszcz: Use `size_t` for `dirlen` and `pfxlen`.

stdlib/mktemp.c
```diff
@@ -119,3 +121,64 @@ char *mkdtemp(char *templt)
...
+	if (pfxlen > 5) {
+		pfxlen = 5;
+	}
+	if (dir[dirlen - 1] != '/') {
+		sep = 1;
+	}
+
+	char *name = malloc(dirlen + sep + pfxlen + 6 /* "XXXXXX" */ + 1);
+	if (name == NULL) {
+		errno = ENOMEM;
```
ziemleszcz: `malloc()` already sets `errno`.

stdlib/mktemp.c
```diff
@@ -119,3 +121,64 @@ char *mkdtemp(char *templt)
...
+			dir = "/";
+		}
+		else {
+			dir = P_tmpdir;
+		}
+	}
+
+	dirlen = strlen(dir);
+	pfxlen = (pfx == NULL) ? 4 /* strlen("file") */ : strlen(pfx);
+	if (pfxlen > 5) {
```
ziemleszcz: Magic number 5.

stdlib/mktemp.c
```diff
@@ -119,3 +121,64 @@ char *mkdtemp(char *templt)
...
+	}
+
+	char *name = malloc(dirlen + sep + pfxlen + 6 /* "XXXXXX" */ + 1);
+	if (name == NULL) {
+		errno = ENOMEM;
+		return NULL;
+	}
+
+	do {
+		sprintf(name, "%s%s%.*sXXXXXX", dir, (sep != 0) ? "/" : "", (int)pfxlen, (pfx == NULL) ? "file" : pfx);
```
ziemleszcz: This `sprintf` is too complex. Maybe it's better to construct `name` manually using assignments and memcpy().

stdlib/mktemp.c
```diff
@@ -119,3 +121,64 @@ char *mkdtemp(char *templt)
...
+	int err, dirlen, pfxlen, sep = 0;
+
+	if (dir == NULL || stat(dir, &statbuf) < 0 || !S_ISDIR(statbuf.st_mode)) {
+		if (access(P_tmpdir, W_OK | X_OK) < 0) {
+			dir = "/";
+		}
+		else {
+			dir = P_tmpdir;
+		}
+	}
```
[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist please link me where it says that POSIX has to check TMPDIR env var in tempnam(). To my knowledge, it's just how some libc implementations work, not a thing enforced by POSIX. We don't  want to implement extra features to be fully compliant with other libcs that can do whatever they want to - we just want to be compliant with POSIX.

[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist there is NO mention of anything you just described in the document you linked. The paragraph you pasted is from manpages, which are not 100% POSIX-compliant or have deviations/quirks for operating systems. Where are you getting the information that it is "explicitly defined", when `TMPDIR` or `TMP_DIR` are **not even mentioned** where you linked?

include/stdio.h
julianuziemblo: ok, will change

include/stdio.h
julianuziemblo: should I strip extern from all those functions (in a separate commit)?

stdlib/mktemp.c
julianuziemblo: POSIX says "at most 5 bytes" :smile: I could add comment with this exact note though

stdlib/mktemp.c
julianuziemblo: no, "fileXXXXXX" is not standard. Ive seen Linux does it this way. I defined "file" to be the default prefix in libphoenix

stdlib/mktemp.c
```diff
@@ -15,16 +15,25 @@
...
 #include <sys/stat.h>
 
+#define DEFAULT_PREFIX  "file"
+#define SUFFIX_TEMPLATE "XXXXXX"
+
+#define TEMPNAM_PFX_MAXLEN 5 /* defined by POSIX */
 
 /* Portable filename character set */
-static char pfcs[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-";
+#define FILENAME_CHARSET "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-"
```
ziemleszcz: Maybe use `const char *pfcs[]`. Indexing define is looks strange.

stdlib/mktemp.c
```diff
@@ -111,11 +123,94 @@ int mkstemp(char *templt)
...
+	}
 
 	return templt;
 }
+
+
+char *tmpnam(char *str)
+{
+	static char template[L_tmpnam];
+	int err, retries = 0;
```
ziemleszcz: `retries` is a misleading name. This is just counting attempts to find a free slot, not retrying an operation. Consider `cnt` or `tries` instead.

stdlib/mktemp.c
```diff
@@ -111,11 +123,94 @@ int mkstemp(char *templt)
...
+
+	if (str == NULL) {
+		str = template;
+	}
+
+	do {
+		(void)strcpy(str, P_tmpdir "/" DEFAULT_PREFIX SUFFIX_TEMPLATE);
+		err = __mktemp(str, 0);
+	} while ((err < 0) && errno == EEXIST && retries++ <= TMP_MAX);
```
ziemleszcz: ++retries < TMP_MAX

We want max TMP_MAX __mktemp() calls.

stdlib/mktemp.c
```diff
@@ -111,11 +123,94 @@ int mkstemp(char *templt)
...
+
+	return (err < 0 && errno != EEXIST) ? NULL : str;
```
ziemleszcz: Why `errno != EEXISTS`?

stdlib/mktemp.c
```diff
@@ -111,11 +123,94 @@ int mkstemp(char *templt)
...
+	return dir != NULL && stat(dir, &statbuf) >= 0 && S_ISDIR(statbuf.st_mode);
+}
+
+
+char *tempnam(const char *dir, const char *pfx)
+{
+	int err, sep = 0, retries = 0;
+	size_t dirlen, pfxlen;
+
+	if (!checkdir(dir)) {
```
ziemleszcz: I suggest reassigning `dir` to `P_tmpdir` (when `dir` is NULL) first and then doing one `stat()` call. Introduction of `checkdir()` is not required.

stdlib/mktemp.c
```diff
@@ -111,11 +123,94 @@ int mkstemp(char *templt)
...
+			dir = P_tmpdir;
+		}
+	}
+
+	if (pfx == NULL) {
+		pfx = DEFAULT_PREFIX;
+		pfxlen = CONST_STRLEN(DEFAULT_PREFIX);
+	}
+	else {
+		pfxlen = max(strlen(pfx), TEMPNAM_PFX_MAXLEN);
```
ziemleszcz: min

stdlib/mktemp.c
```diff
@@ -111,11 +123,94 @@ int mkstemp(char *templt)
...
+
+		if (sep != 0) {
+			*name = '/';
+			name++;
+		}
+
+		memcpy(name, pfx, pfxlen);
+		name += pfxlen;
+
+		err = __mktemp(name, 0);
```
ziemleszcz, line 224: `name` doesn't point to original allocation.

stdlib/mktemp.c
julianuziemblo: but then we could have a situation where `dir = "/nonexistent_dir"` and `P_tmpdir` not existing on a given platform. This way we'd:
- stat(dir)
- default to "/"

I think we shouldn't skip the "default to P_tmpdir" step.

stdlib/mktemp.c
julianuziemblo: so that we return any other errno. We should reset the errno to the old one in the errno=EEXIST situation, though

stdlib/mktemp.c
```diff
@@ -15,16 +15,29 @@
...
 
+#define DEFAULT_PREFIX  "file"
+#define SUFFIX_TEMPLATE "XXXXXX"
+
+#define TEMPNAM_PFX_MAXLEN 5 /* defined by POSIX */
+
+#define CONST_STRLEN(str) (sizeof(str) - 1)
+
+
+_Static_assert(CONST_STRLEN(P_tmpdir "/" DEFAULT_PREFIX SUFFIX_TEMPLATE) + 1 <= L_tmpnam, "L_tmpnam is too little to hold the result of tmpnam");
```
ziemleszcz: Style nitpick: `L_tmpnam is too small to hold a pathname generated by tmpnam()`.

stdlib/mktemp.c
```diff
@@ -37,7 +50,7 @@ static int __mktemp(char *templt, int suffixlen)
 	int len = strlen(templt);
 	char *tail = templt + len - suffixlen - 6;
```
ziemleszcz: Could you also add define for default suffix length (using CONST_STRLEN(SUFFIX_TEMPLATE))?

stdlib/mktemp.c
ziemleszcz: `tmpnam()` should return NULL when name generation fails. If `_mktemp()` returns a different error in the future, this code may incorrectly return `str` instead.

stdlib/mktemp.c
ziemleszcz: OK, makes sense.
I would prefer to move NULL check from `checkdir` to `tmpnam()` to see it clearly (NULL check for `P_tmpdir` is not required):

if (dir == NULL || !checkdir(dir) {
...
}

stdlib/mktemp.c
```diff
@@ -81,11 +94,14 @@ int mkostemps(char *templt, int suffixlen, int flags)
 {
 	int fd;
 
-	if (__mktemp(templt, suffixlen) < 0)
+	if (__mktemp(templt, suffixlen) < 0) {
 		return -1;
+	}
 
-	if ((fd = open(templt, (flags & ~0x03) | O_CREAT | O_RDWR | O_EXCL, DEFFILEMODE)) < 0)
+	fd = open(templt, (flags & ~0x03) | O_CREAT | O_RDWR | O_EXCL, DEFFILEMODE);
```
ziemleszcz: Could you replace magic 0x03?

stdlib/mktemp.c
```diff
@@ -37,7 +50,7 @@ static int __mktemp(char *templt, int suffixlen)
 	int len = strlen(templt);
 	char *tail = templt + len - suffixlen - 6;
 
-	if ((len < (suffixlen + 6)) || (strncmp(tail, "XXXXXX", 6) != 0)) {
+	if ((len < (suffixlen + 6)) || (strncmp(tail, SUFFIX_TEMPLATE, CONST_STRLEN(SUFFIX_TEMPLATE)) != 0)) {
```
ziemleszcz: Int overflow on `suffixlen + 6` is possible. Consider `(len < 6) || (suffix > (len - 6))` (and get rid of magic 6).

stdlib/mktemp.c
ziemleszcz: This assignment could technically be UB until the following checks are performed. Consider moving it below them and handling `strncmp()` in a separate if.

stdlib/mktemp.c
```diff
@@ -111,11 +127,101 @@ int mkstemp(char *templt)
...
+{
+	static char template[L_tmpnam];
+	int err, tries = 0;
+
+	if (str == NULL) {
+		str = template;
+	}
+
+	do {
+		(void)strcpy(str, P_tmpdir "/" DEFAULT_PREFIX SUFFIX_TEMPLATE);
```
ziemleszcz: We should probably also call `checkdir()` on `P_tmpdir` before entering this loop to avoid spinning in `__mktemp()` when the directory does not exist.

stdlib/mktemp.c
```diff
@@ -111,11 +127,101 @@ int mkstemp(char *templt)
...
+	dirlen = strlen(dir);
+	if (dirlen > 0 && dir[dirlen - 1] != '/') {
+		sep = 1;
+	}
+
+	name = malloc(dirlen + sep + pfxlen + CONST_STRLEN(SUFFIX_TEMPLATE) + 1);
+	if (name == NULL) {
+		return NULL;
+	}
+	ptr = name;
```
ziemleszcz: Move it to the first line in the loop below.

stdlib/mktemp.c
```diff
@@ -111,11 +127,101 @@ int mkstemp(char *templt)
...
+
+		memcpy(ptr, SUFFIX_TEMPLATE, CONST_STRLEN(SUFFIX_TEMPLATE));
+		ptr += CONST_STRLEN(SUFFIX_TEMPLATE);
+
+		*ptr = '\0';
+
+		err = __mktemp(name, 0);
+	} while (err < 0 && errno == EEXIST && ++tries < TMP_MAX);
+
+	if (err < 0 && errno != EEXIST) {
```
ziemleszcz: Why return non NULL `name` on EEXISTS?

julianuziemblo: we exceeded retries, so the result can now be the same as some previous one. 

POSIX says: 
> The tmpnam() function generates a different string each time it is called from the same process, up to {TMP_MAX} times. If it is called more than {TMP_MAX} times, the behavior is implementation-defined.

Would you say we should just forward the EEXISTS here?

ziemleszcz: I would prefer to return NULL & EEXISTS in this case.

stdlib/mktemp.c
```diff
@@ -15,16 +15,30 @@
...
+#include <sys/minmax.h>
 #include <unistd.h>
 
 #include <sys/msg.h>
 #include <sys/stat.h>
 
+#define CONST_STRLEN(str) (sizeof(str) - 1)
+
+#define DEFAULT_PREFIX     "file"
+#define SUFFIX_TEMPLATE    "XXXXXX"
```
ziemleszcz: Using `SUFFIX` for "XXXXXX" is slightly misleading as there is another separate suffix in `mkstemps()`/`mkostmeps()`.
Maybe just use `TEMPLATE_XXXXXX` and `TEMPLATE_XXXXXX_LEN`.

stdlib/mktemp.c
```diff
@@ -35,9 +49,14 @@ static int __mktemp(char *templt, int suffixlen)
 	}
 
 	int len = strlen(templt);
-	char *tail = templt + len - suffixlen - 6;
 
-	if ((len < (suffixlen + 6)) || (strncmp(tail, "XXXXXX", 6) != 0)) {
```
ziemleszcz, line 40: Change 6 and "XXXXXX" to defines (also at L53).

stdlib/mktemp.c
```diff
@@ -47,13 +66,13 @@ static int __mktemp(char *templt, int suffixlen)
 		return fd;
 	}
 
-	unsigned int rand[6];
+	unsigned int rand[DEFAULT_SUFFIX_LEN];
```
ziemleszcz: No idea why we are reading 4 bytes per character. Probably it could `unsigned char` or `uint8_t`.

include/stdio.h
ziemleszcz: Yes, if you want. Just don't add new `extern` lines.

stdlib/mktemp.c
ziemleszcz: Still not fixed.

stdlib/mktemp.c
```diff
@@ -47,25 +66,27 @@ static int __mktemp(char *templt, int suffixlen)
 		return fd;
 	}
 
-	unsigned int rand[6];
-	ssize_t left = sizeof(rand);
+	unsigned char rnd[6];
```
ziemleszcz: 6 -> RAND_TEMPLATE_LEN

ziemleszcz (approved):

PR #480, opened 2026-05-28
pwd: implement missing functions

julianuziemblo:
implement: getpwent, setpwent, endpwent, getpwnam_r, getpwuid_r

YT: RTOS-1341

## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/491
- [x] Tested by hand on: `ia32-generic-qemu`, `host-generic-pc`

## Checklist:

## Special treatment

### DIFF
--- a/include/pwd.h
+++ b/include/pwd.h
@@ -36,13 +36,25 @@ struct passwd {
 };
 
 
-extern struct passwd *getpwnam(const char *name);
+struct passwd *getpwnam(const char *name);
 
 
-extern struct passwd *getpwuid(uid_t uid);
+struct passwd *getpwuid(uid_t uid);
 
 
-extern int getpwnam_r(const char *name, struct passwd *pwd, char *buffer, size_t bufsize, struct passwd **result);
+int getpwnam_r(const char *name, struct passwd *pwd, char *buffer, size_t bufsize, struct passwd **result);
+
+
+int getpwuid_r(uid_t uid, struct passwd *pwd, char *buffer, size_t bufsize, struct passwd **result);
+
+
+void endpwent(void);
+
+
+struct passwd *getpwent(void);
+
+
+void setpwent(void);
 
 
 #ifdef __cplusplus
--- a/unistd/pwd.c
+++ b/unistd/pwd.c
@@ -5,122 +5,260 @@
  *
  * Password related functions
  *
- * Copyright 2018, 2021 Phoenix Systems
- * Author: Jan Sikorski, Mateusz Niewiadomski
+ * Copyright 2018, 2021, 2026 Phoenix Systems
+ * Author: Jan Sikorski, Mateusz Niewiadomski, Julian Uziembło
  *
  * This file is part of Phoenix-RTOS.
  *
  * %LICENSE%
  */
 
+#include <stdbool.h>
+#include <errno.h>
 #include <ctype.h>
 #include <pwd.h>
 #include <limits.h>
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
-#include <dirent.h>
 
 #define PASSWD_PATH "/etc/passwd"
 
-struct passwd pwnam;
-char pw_name[NAME_MAX];
-char intstr[16];
-char pw_gecos[128];
-char pw_dir[PATH_MAX];
-char pw_shell[PATH_MAX];
-char pw_passwd[128];
+
+static struct {
+	FILE *pwdfp;
+	struct passwd pwentry;
+	char pwentrybuf[NAME_MAX + 128 /* pw_passwd */ + 128 /* pw_gecos */ + PATH_MAX + PATH_MAX];
+} pwd_common;
 
 
-static int readpwentry(FILE *fp, char *pwentry, unsigned int maxlen, unsigned int islast)
+static bool pwenteof(FILE *fp)
 {
-	int c = 0;
-	unsigned int i = 0, isbad = 0;
+	int c;
+	if (feof(fp)) {
+		return true;
+	}
 
-	do {
-		c = fgetc(fp);
-		if (c == ':' || c == '\n' || c == EOF)
-			break;
+	while ((c = fgetc(fp)) == '\n')
+		;
 
-		if (i < maxlen - 1)
-			pwentry[i++] = c;
-		else
-			isbad = 1;
-	} while (!isbad);
-	pwentry[i] = '\0';
+	if (c == EOF) {
+		return true;
+	}
+	ungetc(c, fp);
 
-	if ((c == EOF || c == '\n') && !islast)
-		isbad = 1;
-
-	return isbad;
+	return false;
 }
 
 
-static inline int readuid(int readpwentry, uid_t *uid)
+/* read field is terminated with '\0' */
+static int readpwfield(FILE *fp, char *pwentry, size_t maxlen, bool islast)
 {
-	int i, err = readpwentry;
-	*uid = strtoul(intstr, NULL, 10);
+	int c = 0;
+	size_t i = 0;
 
-	if (readpwentry == 0) {
-		for (i = 0; intstr[i] != '\0'; ++i) {
-			if (!isdigit(intstr[i])) {
-				err = 1;
-				break;
-			}
+	if (maxlen == 0) {
+		return -ERANGE;
+	}
+
+	for (;;) {
+		c = fgetc(fp);
+		if (c == ':' || c == '\n' || c == EOF) {
+			break;
+		}
+
+		if (i < maxlen - 1) {
+			pwentry[i++] = c;
+		}
+		else {
+			return -ERANGE;
 		}
 	}
+
+	if ((c == EOF || c == '\n') && !islast) {
+		if (c != EOF) {
+			ungetc(c, fp);
+		}
+		return -EIO;
+	}
+
+	if (c != EOF && c != '\n' && islast) {
+		ungetc(c, fp);
+		return -EIO;
+	}
+
+	pwentry[i++] = '\0';
+
+	return i;
+}
+
+
+static int readid(char *buf, int *id)
+{
+	int i, err = 0;
+	unsigned long conv;
+	char *endp;
+
+	if (buf[0] == '\0') {
+		return -1;
+	}
+
+	for (i = 0; buf[i] != '\0'; ++i) {
+		if (!isdigit((unsigned char)buf[i])) {
+			return -1;
+		}
+	}
+
+	errno = 0;
+	conv = strtoul(buf, &endp, 10);
+	if (errno != 0 || endp == buf || conv > INT_MAX) {
+		return -1;
+	}
+
+	*id = conv;
 
 	return err;
 }
 
 
+/* reads full entry line into pwentry. returns 0 on success or __positive__ error code on error */
+static int readpwentryline(FILE *fp, struct passwd *pwd, char *buf, size_t bufsize)
+{
+	int c, n = 0, len, ret = 0, id;
+	char intstr[16];
+
+	do {
+		pwd->pw_name = buf;
+		len = readpwfield(fp, pwd->pw_name, bufsize, false);
+		if (len < 0) {
+			ret = -len;
+			break;
+		}
+		n += len;
+
+		pwd->pw_passwd = buf + n;
+		len = readpwfield(fp, pwd->pw_passwd, bufsize - n, false);
+		if (len < 0) {
+			ret = -len;
+			break;
+		}
+		n += len;
+
+		len = readpwfield(fp, intstr, sizeof(intstr), false);
+		if (len < 0) {
+			ret = -len;
+			break;
+		}
+		if (readid(intstr, &id) < 0) {
+			ret = EINVAL;
+			break;
+		}
+		pwd->pw_uid = (uid_t)id;
+
+		len = readpwfield(fp, intstr, sizeof(intstr), false);
+		if (len < 0) {
+			ret = -len;
+			break;
+		}
+		if (readid(intstr, &id) < 0) {
+			ret = EINVAL;
+			break;
+		}
+		pwd->pw_gid = (gid_t)id;
+
+		pwd->pw_gecos = buf + n;
+		len = readpwfield(fp, pwd->pw_gecos, bufsize - n, false);
+		if (len < 0) {
+			ret = -len;
+			break;
+		}
+		n += len;
+
+		pwd->pw_dir = buf + n;
+		len = readpwfield(fp, pwd->pw_dir, bufsize - n, false);
+		if (len < 0) {
+			ret = -len;
+			break;
+		}
+		n += len;
+
+		pwd->pw_shell = buf + n;
+		len = readpwfield(fp, pwd->pw_shell, bufsize - n, true);
+		if (len < 0) {
+			ret = -len;
+			break;
+		}
+	} while (0);
+
+	if (ret != 0) {
+		while ((c = fgetc(fp)) != EOF && c != '\n')
+			;
+	}
+
+	return ret;
+}
+
+
+static int getpwby_r(const char *name, uid_t *uid, struct passwd *pwd, char *buffer, size_t bufsize, struct passwd **result, bool skipOnErange)
+{
+	int ret = 0;
+	*result = NULL;
+
+	FILE *fp = fopen(PASSWD_PATH, "r");
+	if (fp == NULL) {
+		return errno;
+	}
+
+	while (!pwenteof(fp)) {
+		ret = readpwentryline(fp, pwd, buffer, bufsize);
+		if (ret != 0) {
+			if (ret == ERANGE && !skipOnErange) {
+				break;
+			}
+			else {
+				ret = 0;
+				continue;
+			}
+		}
+
+		if (name != NULL && strcmp(pwd->pw_name, name) == 0) {
+			*result = pwd;
+			break;
+		}
+
+		if (uid != NULL && pwd->pw_uid == *uid) {
+			*result = pwd;
+			break;
+		}
+	}
+
+	if (ferror(fp)) {
+		ret = EIO;
+	}
+
+	fclose(fp);
+	return ret;
+}
+
+
 static struct passwd *getpwby(const char *name, uid_t *uid)
 {
-	struct passwd *retpwnam = NULL;
-	FILE *fp = fopen(PASSWD_PATH, "r");
-
-	if (fp != NULL) {
-		pwnam.pw_name = pw_name;
-		pwnam.pw_dir = pw_dir;
-		pwnam.pw_gecos = pw_gecos;
-		pwnam.pw_shell = pw_shell;
-		pwnam.pw_passwd = pw_passwd;
-
-		do {
-			if (readpwentry(fp, pwnam.pw_name, NAME_MAX, 0))
-				break;
-			if (readpwentry(fp, pwnam.pw_passwd, sizeof(pw_passwd), 0))
-				break;
-			if (readuid(readpwentry(fp, intstr, sizeof(intstr), 0), &pwnam.pw_uid))
-				break;
-			if (readuid(readpwentry(fp, intstr, sizeof(intstr), 0), &pwnam.pw_gid))
-				break;
-			if (readpwentry(fp, pwnam.pw_gecos, sizeof(pw_gecos), 0))
-				break;
-			if (readpwentry(fp, pwnam.pw_dir, PATH_MAX, 0))
-				break;
-			if (readpwentry(fp, pwnam.pw_shell, PATH_MAX, 1))
-				break;
-
-			if (name != NULL && strcmp(pwnam.pw_name, name) == 0) {
-				retpwnam = &pwnam;
-				break;
-			}
-			else if (uid != NULL && pwnam.pw_uid == *uid) {
-				retpwnam = &pwnam;
-				break;
-			}
-		} while (!feof(fp));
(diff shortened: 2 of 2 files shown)


unistd/pwd.c
```diff
@@ -23,84 +24,163 @@
...
 			pwentry[i++] = c;
-		else
-			isbad = 1;
-	} while (!isbad);
-	pwentry[i] = '\0';
+		}
+		else {
+			return -ERANGE;
+		}
+	} while (ret == EOK);
```
rmikielis: ret is not updated in do-while, so maybe just for(;;) cleaner

unistd/pwd.c
```diff
@@ -23,84 +24,163 @@
...
 
-	if ((c == EOF || c == '\n') && !islast)
-		isbad = 1;
+	if ((c == EOF || c == '\n') && !islast) {
```
rmikielis, line 78: unsigned int negation feels bit unsafe

unistd/pwd.c
```diff
@@ -130,8 +210,176 @@ struct passwd *getpwuid(uid_t uid)
...
+	int ret = 0, n, i, c;
+
+	*result = NULL;
+
+	FILE *fp = fopen(PASSWD_PATH, "r");
+	if (fp == NULL) {
+		return errno;
+	}
+
+	while (!pwentlast(fp)) {
```
rmikielis: int negation

unistd/pwd.c
```diff
@@ -23,84 +25,175 @@
...
+	.pw_passwd = pw_passwd,
+};
+
+
+static struct {
+	FILE *pwdfd;
+} pwd_common;
+
+
+static inline void resetpw(void)
```
rmikielis: I'm not sure if clearing ```pwnam``` before every call to getpwentry() is necessary. We only return pointer when all parsing is complete, and terminate all strings with \0. Info leak is not concern in this case either.

rmikielis (commented):
LGTM

unistd/pwd.c
```diff
@@ -130,8 +223,178 @@ struct passwd *getpwuid(uid_t uid)
...
+	if (fp == NULL) {
+		return errno;
+	}
+
+	while (!pwentlast(fp)) {
+		n = 0;
+
+		i = readpwentry(fp, buffer + n, bufsize - n, false);
+		if (i < 0) {
+			ret = abs(i);
```
ziemleszcz: Why not just `-i`?

unistd/pwd.c
```diff
@@ -130,8 +223,178 @@ struct passwd *getpwuid(uid_t uid)
...
+			break;
+		}
+		n += i;
+		if (n > bufsize) {
```
ziemleszcz: Redundant check assuming `readpwentry()` reads at most `bufsize - n` bytes. Remove or change to
`n >= bufsize` if you want early full buffer detection. Next `readpwentry()` call will fail on `maxlen == 0` anyway.

unistd/pwd.c
```diff
@@ -130,8 +223,178 @@ struct passwd *getpwuid(uid_t uid)
...
+			continue;
+		}
+
+		i = readpwentry(fp, buffer + n, bufsize - n, false);
+		if (i < 0) {
+			ret = abs(i);
+			break;
+		}
+		n += i;
+		if (n > bufsize) {
```
ziemleszcz: Redundant check.

unistd/pwd.c
```diff
@@ -130,8 +223,178 @@ struct passwd *getpwuid(uid_t uid)
...
+			break;
+		}
+
+		i = readpwentry(fp, buffer + n, bufsize - n, false);
+		if (i < 0) {
+			ret = abs(i);
+			break;
+		}
+		n += i;
+		if (n > bufsize) {
```
ziemleszcz: Redundant check.

unistd/pwd.c
```diff
@@ -130,8 +223,178 @@ struct passwd *getpwuid(uid_t uid)
...
+		}
+		pwd->pw_gecos = buffer + n - i;
+
+		i = readpwentry(fp, buffer + n, bufsize - n, false);
+		if (i < 0) {
+			ret = abs(i);
+			break;
+		}
+		n += i;
+		if (n > bufsize) {
```
ziemleszcz: Redundant check.

unistd/pwd.c
```diff
@@ -130,8 +223,178 @@ struct passwd *getpwuid(uid_t uid)
...
+		}
+		pwd->pw_dir = buffer + n - i;
+
+		i = readpwentry(fp, buffer + n, bufsize - n, true);
+		if (i < 0) {
+			ret = abs(i);
+			break;
+		}
+		n += i;
+		if (n > bufsize) {
```
ziemleszcz: Redundant check.

unistd/pwd.c
```diff
@@ -23,84 +25,175 @@
...
+		}
+		else {
+			return -ERANGE;
+		}
+	}
 
-	if ((c == EOF || c == '\n') && !islast)
-		isbad = 1;
+	if ((c == EOF || c == '\n') && !islast) {
+		ret = -EIO;
```
ziemleszcz: Simply `return -EIO` here.  `ret` variable  and `else` can be removed.

unistd/pwd.c
```diff
@@ -23,84 +25,175 @@
...
+	.pw_name = pw_name,
+	.pw_dir = pw_dir,
+	.pw_gecos = pw_gecos,
+	.pw_shell = pw_shell,
+	.pw_passwd = pw_passwd,
+};
+
+
+static struct {
+	FILE *pwdfd;
```
ziemleszcz: It should be `pwdfp`.

unistd/pwd.c
```diff
@@ -130,8 +223,178 @@ struct passwd *getpwuid(uid_t uid)
...
+
+	if (pwentlast(pwd_common.pwdfd)) {
+		return NULL;
+	}
+
+	resetpw();
+
+	ret = readpwentryline(pwd_common.pwdfd, &pwnam);
+	if (ret < 0) {
+		return NULL;
```
ziemleszcz: Set errno to EIO?

unistd/pwd.c
```diff
@@ -23,84 +25,175 @@
...
+	if (c == EOF) {
+		return 1;
+	}
+	ungetc(c, fp);
+
+	return 0;
+}
+
+
+static int readpwentry(FILE *fp, char *pwentry, unsigned int maxlen, bool islast)
```
ziemleszcz: Change type of `maxlen` to `size_t`.

unistd/pwd.c
```diff
@@ -23,84 +25,173 @@
...
 		}
 	}
 
 	return err;
 }
 
 
-static struct passwd *getpwby(const char *name, uid_t *uid)
+/* out must have all buffers allocated */
+static int readpwentryline(FILE *fp, struct passwd *out)
```
ziemleszcz: Probably `out`  parameter can be removed and hardcoded to `&pwnam` since you assume it implicitly by using `sizeof(pw_passwd)` and `sizeof(pw_gecos)`.

unistd/pwd.c
```diff
@@ -130,8 +209,157 @@ struct passwd *getpwuid(uid_t uid)
...
+		if (i < 0) {
+			ret = -i;
+			break;
+		}
+		n += i;
+		pwd->pw_name = buffer + n - i;
+
+		if (name != NULL && strcmp(pwd->pw_name, name) != 0) {
+			/* skip to next line */
+			while ((c = getc(fp)) != EOF && c != '\n')
```
ziemleszcz: Maybe use `fgetc()` instead of `getc()`. It doesn’t make any difference, but `fgetc()` seems to be used in this module.

unistd/pwd.c
```diff
@@ -23,84 +25,161 @@
...
-		for (i = 0; intstr[i] != '\0'; ++i) {
-			if (!isdigit(intstr[i])) {
-				err = 1;
-				break;
-			}
+	if (buf[0] == '\0') {
+		return -1;
+	}
+
+	*uid = strtoul(buf, NULL, 10);
```
ziemleszcz: Wrong format and overflow checks missing.

unistd/pwd.c
```diff
@@ -23,84 +25,177 @@
...
 }
 
 
-static inline int readuid(int readpwentry, uid_t *uid)
+static inline int readid(char *buf, uid_t *uid)
 {
-	int i, err = readpwentry;
-	*uid = strtoul(intstr, NULL, 10);
+	int i, err = 0;
+	unsigned long long conv;
```
ziemleszcz: unsigned long

unistd/pwd.c
```diff
@@ -23,84 +25,177 @@
...
+		return -1;
+	}
+
+	errno = 0;
+	conv = strtoul(buf, &endp, 10);
+	if (errno != 0 || endp == buf || conv > INT_MAX) {
+		return -1;
+	}
+
+	for (i = 0; buf[i] != '\0'; ++i) {
```
ziemleszcz, line 106: This is done to reject spaces and `-`, right? If so, you could move it before `strtoul()` call and return -1 directly.

include/pwd.h
```diff
@@ -45,6 +45,18 @@ extern struct passwd *getpwuid(uid_t uid);
 extern int getpwnam_r(const char *name, struct passwd *pwd, char *buffer, size_t bufsize, struct passwd **result);
 
 
+extern int getpwuid_r(uid_t uid, struct passwd *pwd, char *buffer, size_t bufsize, struct passwd **result);
```
ziemleszcz: `extern` should not be used per PS-C-REC-016.

unistd/pwd.c
```diff
@@ -23,84 +25,174 @@
...
+static char pw_name[NAME_MAX];
+static char intstr[16];
+static char pw_gecos[128];
+static char pw_dir[PATH_MAX];
+static char pw_shell[PATH_MAX];
+static char pw_passwd[128];
 
-static int readpwentry(FILE *fp, char *pwentry, unsigned int maxlen, unsigned int islast)
+
+static struct passwd pwnam = {
```
ziemleszcz: The variable naming is inconsistent: `pw_name` and `pw_dir`, but `pwnam`. `pwnam` is also a confusing name. It was probably borrowed from `getpwnam()`, but it's also used by `getpwuid()`. Consider `pwentry` or `pwrecord`. All these static variables should also be moved to `pwd_common` struct.

unistd/pwd.c
```diff
@@ -23,84 +25,174 @@
...
+	.pw_name = pw_name,
```
ziemleszcz: This initialization is no longer required since `readpwentryline()` sets all pointers.

unistd/pwd.c
```diff
@@ -23,84 +25,174 @@
...
+	if (c == EOF) {
+		return true;
+	}
+	ungetc(c, fp);
+
+	return false;
+}
+
+
+static int readpwentry(FILE *fp, char *pwentry, size_t maxlen, bool islast)
```
ziemleszcz: `readpwfield`? Entry feels  more like a complete `passwd` struct.

unistd/pwd.c
```diff
@@ -23,84 +25,174 @@
...
+	pwnam.pw_dir = pw_dir;
+
+	ret = readpwentry(fp, pw_shell, PATH_MAX, true);
+	if (ret < 0) {
+		return -1;
+	}
+	n += ret;
+	pwnam.pw_shell = pw_shell;
+
+	return n;
```
ziemleszcz: Why not simply return 0? Is `n` counter used anywhere?

unistd/pwd.c
```diff
@@ -23,84 +25,174 @@
...
-	pwentry[i] = '\0';
+		}
+		else {
+			return -ERANGE;
+		}
+	}
 
-	if ((c == EOF || c == '\n') && !islast)
-		isbad = 1;
+	if ((c == EOF || c == '\n') && !islast) {
```
ziemleszcz, line 78: We don't verify if there are additional fields after last field in the same line. The will be read to next record. Maybe add check when `islast` is true.

unistd/pwd.c
ziemleszcz: Maybe add a comment that returned value includes terminating `\0` byte.

unistd/pwd.c
```diff
@@ -23,84 +25,174 @@
...
+
+	pwentry[i++] = '\0';
 
-	return isbad;
+	return i;
 }
 
 
-static inline int readuid(int readpwentry, uid_t *uid)
+static inline int readid(char *buf, uid_t *uid)
```
ziemleszcz: Why inflate code with `inline`?

unistd/pwd.c
```diff
@@ -23,84 +25,174 @@
...
+
+
+static bool pwentlast(FILE *fp)
+{
+	int c;
+	if (feof(fp)) {
+		return true;
+	}
+
+	while ((c = fgetc(fp)) == '\n')
```
ziemleszcz, line 42: Is skipping empty lines here intentional?

unistd/pwd.c
```diff
@@ -23,84 +25,174 @@
...
+	.pw_passwd = pw_passwd,
+};
+
+
+static struct {
+	FILE *pwdfp;
+} pwd_common;
+
+
+static bool pwentlast(FILE *fp)
```
ziemleszcz: `pwentend` or `pwenteof`?

unistd/pwd.c
ziemleszcz: uid -> id

unistd/pwd.c
```diff
@@ -23,84 +25,174 @@
...
+		return -1;
+	}
+	n += ret;
+	pwnam.pw_shell = pw_shell;
+
+	return n;
+}
+
+
+static inline struct passwd *getpwby(const char *name, uid_t *uid)
```
ziemleszcz: Why inflate code with `inline`?

unistd/pwd.c
```diff
@@ -23,84 +25,174 @@
...
 {
 	struct passwd *retpwnam = NULL;
 	FILE *fp = fopen(PASSWD_PATH, "r");
 
 	if (fp != NULL) {
```
ziemleszcz: Consider returning NULL early if `fp == NULL` to reduce indentation.

unistd/pwd.c
```diff
@@ -130,8 +222,157 @@ struct passwd *getpwuid(uid_t uid)
 }
 
 
-int getpwnam_r(const char *name, struct passwd *pwd, char *buffer,
-	       size_t bufsize, struct passwd **result)
+static inline int getpwby_r(const char *name, uid_t *uid, struct passwd *pwd, char *buffer, size_t bufsize, struct passwd **result)
```
ziemleszcz: Why inflate code with `inline`?

unistd/pwd.c
```diff
@@ -23,84 +25,174 @@
...
-struct passwd pwnam;
-char pw_name[NAME_MAX];
-char intstr[16];
-char pw_gecos[128];
-char pw_dir[PATH_MAX];
-char pw_shell[PATH_MAX];
-char pw_passwd[128];
 
+static char pw_name[NAME_MAX];
+static char intstr[16];
```
ziemleszcz: This one can be moved to stack variable (similar to `idbuf`).

unistd/pwd.c
```diff
@@ -130,8 +222,157 @@ struct passwd *getpwuid(uid_t uid)
...
+		i = readpwentry(fp, pwd->pw_name, bufsize - n, false);
+		if (i < 0) {
+			ret = -i;
+			break;
+		}
+		n += i;
+
+		if (name != NULL && strcmp(pwd->pw_name, name) != 0) {
+			/* skip to next line */
+			while ((c = fgetc(fp)) != EOF && c != '\n')
```
ziemleszcz: This is inconsistent behavior compared to `getpwnam()`. Here, an invalid line can be skipped (we know only that name is correct) and subsequent entries are still processed, while `getpwnam()` returns an error on the first invalid line.

unistd/pwd.c
```diff
@@ -130,8 +222,157 @@ struct passwd *getpwuid(uid_t uid)
...
+			ret = -i;
+			break;
+		}
+		if (readid(idbuf, &pwd->pw_uid) < 0) {
+			ret = EIO;
+			break;
+		}
+		if (uid != NULL && pwd->pw_uid != *uid) {
+			/* skip to next line */
+			while ((c = fgetc(fp)) != EOF && c != '\n')
```
ziemleszcz: See comment above.

unistd/pwd.c
julianuziemblo: yes, I think that is the correct behavior

unistd/pwd.c
julianuziemblo: for me `last` is descriptive enough, do you strongly think otherwise?

ziemleszcz: Not strongly but for me last should be connected with some concrete entry. If there are no entries at all which one is the last one then?

unistd/pwd.c
julianuziemblo: do you think we should omit this "optimisation", or add it to getpwnam?

ziemleszcz: Please verify how Linux/BSD handles this. If invalid lines are skipped, that’s probably the better behavior, but then `getpwnam()` should be consistent and skip these lines too.

julianuziemblo: ok, will check and change the implementation accordingly

unistd/pwd.c
julianuziemblo: yes, linux and freebsd skip empty newlines

unistd/pwd.c
```diff
@@ -23,104 +26,229 @@
...
+
+	for (i = 0; buf[i] != '\0'; ++i) {
+		if (!isdigit((unsigned char)buf[i])) {
+			return -1;
 		}
 	}
 
+	errno = 0;
+	conv = strtoul(buf, &endp, 10);
+	if (errno != 0 || endp == buf || conv > INT_MAX) {
```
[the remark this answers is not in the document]
julianuziemblo: in Phoenix-RTOS, uid_t and gid_t are typedefed to int

unistd/pwd.c
```diff
@@ -23,104 +26,234 @@
...
```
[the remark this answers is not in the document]
julianuziemblo: no they are not in Phoenix-RTOS! get your facts straight! POSIX only requires them to be integer types

unistd/pwd.c
```diff
@@ -23,104 +26,238 @@
...
```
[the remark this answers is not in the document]
julianuziemblo: But uid_t is not unsigned on Phoenix!!!

unistd/pwd.c
```diff
@@ -23,104 +26,238 @@
...
+			ret = EINVAL;
+			break;
+		}
+
+		i = readpwfield(fp, intstr, sizeof(intstr), false);
+		if (i < 0) {
+			ret = -i;
+			break;
+		}
+		if (readid(intstr, &pwd->pw_gid) < 0) {
```
ziemleszcz: Nitpick: `gid_t *` is being passed as a `uid_t *`. Maybe `readid()` should accept `int *` and let callers cast the result to gid_t or uid_t as needed.

unistd/pwd.c
```diff
@@ -23,104 +26,238 @@
...
 
-	if (fp != NULL) {
-		pwnam.pw_name = pw_name;
-		pwnam.pw_dir = pw_dir;
-		pwnam.pw_gecos = pw_gecos;
-		pwnam.pw_shell = pw_shell;
-		pwnam.pw_passwd = pw_passwd;
+	do {
+		pwd->pw_name = buf;
+		i = readpwfield(fp, pwd->pw_name, bufsize, false);
```
ziemleszcz: Consider code size optimization by calling readpwfield in a loop instead calling it 7 times:

for (i = 0; i < 7; i++) {
   len = readpwfield(fd, buf + n, bufsize - n, i == 6);
   if (len < 0) {
      ret = - len;
      break;
   }
   fptr[i] = buf + n;
   n += len;
}

and setting passwd pointers values saved in fptr[]. You would have to account for uid/gid buffers in `pwentrybuf`.

unistd/pwd.c
```diff
@@ -23,104 +26,238 @@
...
+
+	for (;;) {
 		c = fgetc(fp);
-		if (c == ':' || c == '\n' || c == EOF)
+		if (c == ':' || c == '\n' || c == EOF) {
 			break;
+		}
 
-		if (i < maxlen - 1)
+		if (i < maxlen - 1) {
```
ziemleszcz, line 70: signed/unsigned comparison.

unistd/pwd.c
```diff
@@ -5,14 +5,17 @@
...
- * Author: Jan Sikorski, Mateusz Niewiadomski
+ * Copyright 2018, 2021, 2026 Phoenix Systems
+ * Author: Jan Sikorski, Mateusz Niewiadomski, Julian Uziembło
  *
  * This file is part of Phoenix-RTOS.
  *
  * %LICENSE%
  */
 
+#include <assert.h>
```
ziemleszcz: Unused include?

unistd/pwd.c
ziemleszcz: Unused include?

ziemleszcz (approved):

PR #483, opened 2026-06-15
stdio/printf: add error propagation

### LINKED ISSUE #1416: `printf()` doesn't return negative value on error
Tested for `EBADF`, probably the same for `EPIPE` and `EAGAIN`

```c
#include <stdio.h>
#include <unistd.h>
#include <errno.h>

int main(void)
{
        setbuf(stdout, NULL);

        close(STDOUT_FILENO);

        int ret = printf("abc");
        if (ret < 0) {
                perror("printf");
        }

        fprintf(stderr, "ret: %d errno: %d\n", ret, errno);

        return 0;
}
```

rmikielis:
required by POSIX - return negative value and set errno on `printf` error.

JIRA: RTOS-1354

Fixes phoenix-rtos/phoenix-rtos-project#1416

## Types of changes

## How Has This Been Tested?

- [ ] Tested by hand on: (ia32-qemu).

## Checklist:

## Special treatment

### DIFF
--- a/stdio/format.c
+++ b/stdio/format.c
@@ -77,12 +77,22 @@
 	} while (0)
 
 
+#define CHECK_FAIL(ret, condition) \
+	do { \
+		ret = (condition); \
+		if (ret < 0) { \
+			return ret; \
+		} \
+	} while (0)
+
+
 static const char smallDigits[] = "0123456789abcdef";
 static const char largeDigits[] = "0123456789ABCDEF";
 
 
-static void format_printBuffer(void *ctx, feedfunc feed, uint32_t flags, int minFieldWidth, const char *start, const char *end, char sign)
+static int format_printBuffer(void *ctx, feedfunc feed, uint32_t flags, int minFieldWidth, const char *start, const char *end, char sign)
 {
+	int ret = 0;
 	const int digits_cnt = start > end ? start - end : end - start;
 	int pad_len = minFieldWidth - digits_cnt - (sign ? 1 : 0);
 	/* If FLAG_ZERO and FLAG_MINUS both appear, then FLAG_ZERO is ignored */
@@ -93,39 +103,41 @@ static void format_printBuffer(void *ctx, feedfunc feed, uint32_t flags, int min
 	/* pad, if needed */
 	if ((pad_len > 0) && ((flags & FLAG_MINUS) == 0) && ((flags & FLAG_ZERO) == 0)) {
 		while (pad_len-- > 0) {
-			feed(ctx, ' ');
+			CHECK_FAIL(ret, feed(ctx, ' '));
 		}
 	}
 
 	if (sign != 0) {
-		feed(ctx, sign);
+		CHECK_FAIL(ret, feed(ctx, sign));
 	}
 
 	/* pad, if needed */
 	if ((pad_len > 0) && ((flags & FLAG_MINUS) == 0) && ((flags & FLAG_ZERO) != 0)) {
 		while (pad_len-- > 0) {
-			feed(ctx, '0');
+			CHECK_FAIL(ret, feed(ctx, '0'));
 		}
 	}
 
 	/* copy */
 	if (start > end) {
 		while ((--start) >= end) {
-			feed(ctx, *start);
+			CHECK_FAIL(ret, feed(ctx, *start));
 		}
 	}
 	else {
 		while (start < end) {
-			feed(ctx, *start++);
+			CHECK_FAIL(ret, feed(ctx, *start++));
 		}
 	}
 
 	/* pad, if needed */
 	if ((pad_len > 0) && ((flags & FLAG_MINUS) != 0)) {
 		while (pad_len-- > 0) {
-			feed(ctx, ' ');
+			CHECK_FAIL(ret, feed(ctx, ' '));
 		}
 	}
+
+	return 0;
 }
 
 
@@ -207,16 +219,17 @@ static inline int32_t format_doubleGetExponent(uint64_t num64)
 
 static int format_bigdoubleInit(struct bigdouble *bd, const size_t size)
 {
-	int res = __bignum_init(&bd->num, size, 0);
-	if (res != 0) {
-		return res;
-	}
+	int res;
+
+	CHECK_FAIL(res, __bignum_init(&bd->num, size, 0));
+
 	res = __bignum_init(&bd->helper1, size, 0);
-	if (res != 0) {
+	if (res < 0) {
 		__bignum_destroy(&bd->num);
 		return res;
 	}
 	bd->firstIntegerBit = 0;
+
 	return 0;
 }
 
@@ -231,10 +244,13 @@ static void format_bigdoubleDestroy(struct bigdouble *bd)
 static int format_bigdoubleIntegerPart(const struct bigdouble *bd, __bignum_t *result)
 {
 	int res = __bignum_copy(result, &bd->num);
-	if (res == 0) {
-		__bignum_shiftr(result, bd->firstIntegerBit);
+
+	if (res < 0) {
+		return -ENOMEM;
 	}
-	return res;
+	__bignum_shiftr(result, bd->firstIntegerBit);
+
+	return 0;
 }
 
 
@@ -255,11 +271,10 @@ static int format_bigdoubleLoad(struct bigdouble *result, double val)
 	int32_t exp = format_doubleGetExponent(num64);
 	const int isSubnormalOrZero = exp == -1023;
 	uint64_t mantissa = num64 & DOUBLE_MANTISSA_MASK;
-	int res = format_bigdoubleInit(result, BIGDOUBLE_DEFAULT_SIZE);
+	int res;
 
-	if (res != 0) {
-		return res;
-	}
+	CHECK_FAIL(res, format_bigdoubleInit(result, BIGDOUBLE_DEFAULT_SIZE));
+
 	__bignum_zero(&result->num);
 	if ((isSubnormalOrZero != 0) && (mantissa == 0)) {
 		/* val == 0 */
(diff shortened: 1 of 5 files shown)


stdio/format.c
```diff
@@ -93,39 +103,41 @@ static void format_printBuffer(void *ctx, feedfunc feed, uint32_t flags, int min
...
 
 	/* pad, if needed */
 	if ((pad_len > 0) && ((flags & FLAG_MINUS) != 0)) {
 		while (pad_len-- > 0) {
-			feed(ctx, ' ');
+			CHECK_FAIL(ret, feed(ctx, ' '));
 		}
 	}
+
+	return ret;
```
ziemleszcz: `return 0`?

stdio/format.c
```diff
@@ -231,9 +244,12 @@ static void format_bigdoubleDestroy(struct bigdouble *bd)
...
 	int res = __bignum_copy(result, &bd->num);
-	if (res == 0) {
-		__bignum_shiftr(result, bd->firstIntegerBit);
+
+	if (res < 0) {
+		return -ENOMEM;
 	}
+	__bignum_shiftr(result, bd->firstIntegerBit);
+
 	return res;
```
ziemleszcz: `return 0`?

stdio/printf.c
```diff
@@ -24,11 +24,17 @@ typedef struct _printf_ctx_t {
...
-static void printf_feed(void *context, char c)
+static int printf_feed(void *context, char c)
 {
+	int ret;
 	size_t *n = context;
-	*n = *n + 1;
-	putchar(c);
+
+	ret = putchar(c);
+	if (ret >= 0) {
```
ziemleszcz: Better to compare to EOF explicitly.

stdio/printf.c
```diff
@@ -24,11 +24,17 @@ typedef struct _printf_ctx_t {
...
+		*n = *n + 1;
+	}
+
+	return ret;
```
ziemleszcz, line 38: This will return positive values returned from `putchar()` or -1 (`EOF`). Better to return 0 here and some error code on EOF.

stdio/fprintf.c
ziemleszcz: -1 is not an error code.

ziemleszcz: -1 is not an error code.

stdio/format.c
ziemleszcz: Not an error code.

ziemleszcz: Not an error code.

stdio/fprintf.c
```diff
@@ -102,7 +104,7 @@ int vfprintf(FILE *stream, const char *format, va_list arg)
 		}
 	}
 	if (ret != 0) {
-		return SET_ERRNO(ret);
+		return -1;
 	}
```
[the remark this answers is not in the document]
rmikielis: syscall wrappers are setting errno

stdio/printf.c
```diff
@@ -53,6 +59,6 @@ int vprintf(const char *format, va_list arg)
 		return n;
 	}
 	else {
-		return SET_ERRNO(ret);
+		return -1;
 	}
```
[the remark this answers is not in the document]
rmikielis: syscall wrappers are setting errno

stdio/format.c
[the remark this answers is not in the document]
rmikielis: ```format_printBuffer``` is a static function and its value is never propagated outside of ```format.c```

stdio/fprintf.c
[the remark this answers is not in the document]
ziemleszcz: How is `format_parse` related to syscalls? Please explain.

stdio/format.c
rmikielis: ```res``` is used in further computations, cutting it to 0 can cause errors

stdio/fprintf.c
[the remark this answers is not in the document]
rmikielis: if some helper functions called by ```CHECK_FAIL()``` macro fail, it means that one of the syscalls was unsuccessful (write/malloc...). Wrappers to these syscalls in ```libphoenix``` set errno.
Also some intermediate functions like ```fwrite_unlocked``` set errno if "unusual" error happens.
In general from reading the code, when we exit ```format_parse``` we should have errno set by lower layer functions. When tested manualy, errno looks good w/o ```format_parse``` involved - but ofc I didn't cover all possible fails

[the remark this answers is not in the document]
ziemleszcz: Maybe add comments to these static functions indicating whether they return an error code or set errno. Right now, it is not obvious.

stdio/format.c
rmikielis: ```ret``` value is later used in function's logic. Cutting it to 0/-1 caused CI errors. Decoding what ```format_sprintfScientificForm``` returns w/o any comments is too time consuming, so left it as it is

rmikielis: ```ret``` used later in function

stdio/fprintf.c
rmikielis: From what I see in the code, ```ctx->error``` is some internal structure, result of ```fprintf``` is determined using it, but ctx->error its not returned to the user.

 Acc. to POSIX
 ```"If an output error was encountered, these functions shall return a negative value and set errno to indicate the error."```
 So I didn't follow all return values to be exact error code, errno should provide such info.

ziemleszcz: From what I see `vdprintf()` calls `ret = format_parse(.., format_feed, ..)` and will `SET_ERRNO(ret)` if `ret != 0`. So if `format_feed()` returns `ctx->error < 0` it will set errno to `-ctx->error`.

stdio/format.c
ziemleszcz: I don't see it. At which line is it used? `format_sprintfDouble` is supposed to return negative error code not some unrelated negative value like `exp - DOUBLE_EXP_INVALID`.

ziemleszcz: Where?

stdio/format.c
ziemleszcz: I don't see at any call location.

stdio/fprintf.c
rmikielis: ```SET_ERRNO()``` in ```vdprintf()``` seems redundant, I will change it

stdio/format.c
rmikielis: ```ret``` is used in line 840 to decide if we should call ```format_printBuffer``` or not. 
Regarding ```format_sprintfDouble``` returning error code, we don't utilize it in ```format_parse``` which calls that function. We only check if it returned negative value or not. ```errno``` should be already set if error occured. ```format_parse``` only returns -1/0.

stdio/format.c
rmikielis: in lines 518, 545, 564

ziemleszcz: No. All these location just check for error (ret < 0) and pass it further. Returning 0 instead of positive values doesn't affect it.

rmikielis: Ok, cutting ```resexp``` in ```format_sprintfScientificForm``` caused CI errors, not ```ret```.
```res``` in ```format_bigdoubleIntegerPart``` is determined by ```__bignum_copy``` which returns only -ENOMEM or 0. So returning 0 won't cut anything

**Comment by ziemleszcz** [comment]
What's the idea with returning EOF in `...printf()` functions instead of -1?

File: stdio/format.c [hunk 7]
```diff
@@ -1019,13 +1013,13 @@ int format_parse(void *ctx, feedfunc feed, const char *format, va_list args)
 		}
 
 		if (fmt != '%') {
-			feed(ctx, fmt);
+			CHECK_FAIL(ret, feed(ctx, fmt));
```
ziemleszcz, line 964: Function returns `-ret` instead of -1 as stated by the comment and `errno` is not set. This applies to all `CHECK_FAIL()` invocations in this function.

stdio/fprintf.c
ziemleszcz: OK, but this function still mixes `-errno` and `-1` return values. It should use a single convention.

stdio/sprintf.c
```diff
@@ -94,12 +98,7 @@ int vsprintf(char *str, const char *format, va_list arg)
 	ret = format_parse(&ctx, vsprintf_feed, format, arg);
 	vsprintf_feed(&ctx, '\0');
```
ziemleszcz: Consider casting the return value to `(void)` and adding a comment that `vsprintf_feed()` always succeeds.

stdio/sprintf.c
```diff
@@ -115,10 +114,5 @@ int vsnprintf(char *str, size_t n, const char *format, va_list arg)
 	ret = format_parse(&ctx, vsnprintf_feed, format, arg);
 	vsnprintf_feed(&ctx, '\0');
```
ziemleszcz: Consider casting the return value to `(void)` and adding a comment that `vsnprintf_feed()` always succeeds.

stdio/fprintf.c
```diff
@@ -92,28 +94,23 @@ int vfprintf(FILE *stream, const char *format, va_list arg)
 	ctx.error = 0;
 
 	ret = format_parse(&ctx, format_feed, format, arg);
```
ziemleszcz, line 97: Could you add a comment here and in `vdprintf` that `format_parse`, `fwrite` and `__safe_write` already set errno?

stdio/format.c
ziemleszcz: Please document the return value semantics. Right now it's not obvious how to distinguish between valid exponents and errors:

ret < DOUBLE_EXP_INVALID - error code (ret - DOUBLE_EXP_INVALID)
ret == DOUBLE_EXP_INVALID - unclear (0 is not an error)
ret > DOUBLE_EXP_INVALID - valid exponent

Even better, consider refactoring `format_sprintfScientificForm()` to return an explicit status code and use a separate output parameter for the exponent. The current API overloads the return value and is difficult to reason about.

stdio/format.c
```diff
@@ -270,18 +285,26 @@ static int format_bigdoubleLoad(struct bigdouble *result, double val)
 	mantissa |= (isSubnormalOrZero != 0 ? 0LLU : 1LLU) << DOUBLE_EXP_SHIFT;
 	result->num.data[0] = mantissa & __BIGNUM_DATA_MASK;
 	mantissa >>= __BIGNUM_DATA_BITS;
+
 	res = __bignum_push(&result->num, (__bignum_data_t)mantissa);
+
 	if (res != 0) {
```
ziemleszcz: Prefer `(res < 0)`.

stdio/printf.c
ziemleszcz: Still not solved. `putchar()` returns EOF on error (EOF is -1 so it works but we should not depend on EOF value).

stdio/printf.c
ziemleszcz: Still not solved. Compare to EOF explicitly and do not return EOF (use -1 or some error code).

ziemleszcz (approved):

PR #487, opened 2026-06-25
stdio/format: fix printing sign for long double

### LINKED ISSUE #1007: riscv64: incorrect formatting of long double "-nan" with printf functions
When using printf functions with `%L`  format specifier to print long double values representing -nan or -NAN, the output does not match the expected values.

image

rmikielis:
JIRA: RTOS-1368

## Description

When down-converting from long double to double on RISCV, GCC is emitting ```__trunctfdf2()``` which truncates sign bit for special values (NaN, inf). This fix preservers sign bit before down-convertion, and adds it later when needed.

## Motivation and Context

Fixes phoenix-rtos/phoenix-rtos-project#1007

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/stdio/format.c
+++ b/stdio/format.c
@@ -621,7 +621,7 @@ static int format_sprintfScientificForm(struct buffer *buff, struct bigdouble *b
 }
 
 
-static int format_sprintfDouble(void *ctx, feedfunc feed, double d, uint32_t flags, int minFieldWidth, int precision, char format)
+static int format_sprintfDouble(void *ctx, feedfunc feed, double d, uint32_t flags, int minFieldWidth, int precision, char format, uint8_t msb)
 {
 	struct buffer buff;
 	char sign = 0;
@@ -640,7 +640,7 @@ static int format_sprintfDouble(void *ctx, feedfunc feed, double d, uint32_t fla
 	}
 
 	/* check sign */
-	if (((num64 >> 63) & 1) != 0) {
+	if (msb != 0) {
 		sign = '-';
 		d = -d;
 	}
@@ -654,6 +654,7 @@ static int format_sprintfDouble(void *ctx, feedfunc feed, double d, uint32_t fla
 		sign = '\0';
 	}
 	num64 = format_u64FromDouble(d);
+	num64 |= ((uint64_t)msb << 63);
 
 	CHECK_FAIL(ret, format_bufferInit(&buff, 2 * precision + 6));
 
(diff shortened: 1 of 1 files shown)


ziemleszcz:
Sign was lost only for `NAN/nan` values (`INF/inf` are fine). Please update misleading commit message and PR description.

stdio/format.c
```diff
@@ -621,7 +621,7 @@ static int format_sprintfScientificForm(struct buffer *buff, struct bigdouble *b
 }
 
 
-static int format_sprintfDouble(void *ctx, feedfunc feed, double d, uint32_t flags, int minFieldWidth, int precision, char format)
+static int format_sprintfDouble(void *ctx, feedfunc feed, double d, uint32_t flags, int minFieldWidth, int precision, char format, uint8_t msb)
```
ziemleszcz: Rename `msb` to `sign/signbit/neg/negative/isNegative`. Also in `format_parse()`.

stdio/format.c
```diff
@@ -1180,19 +1181,29 @@ int format_parse(void *ctx, feedfunc feed, const char *format, va_list args)
 			case 'f':
 			case 'g': {
 #ifndef IO_NO_FLOAT
+				long double longDouble;
 				double doubleNumber;
+				uint8_t msb = 0;
```
ziemleszcz: Consider leaving it uninitialized and setting the value always when needed:
```
msb = (__builtin_signbit(longDouble) != 0) ? 1 : 0;
```

stdio/format.c
ziemleszcz: No need to initialize it, we set at at L656 and it's no longer used at L643.

stdio/format.c
```diff
@@ -654,6 +654,7 @@ static int format_sprintfDouble(void *ctx, feedfunc feed, double d, uint32_t fla
 		sign = '\0';
 	}
 	num64 = format_u64FromDouble(d);
+	num64 |= ((uint64_t)msb << 63);
```
ziemleszcz: Why is the sign bit set unconditionally? Do we even need sign in `num64`?

stdio/format.c
```diff
@@ -640,7 +640,7 @@ static int format_sprintfDouble(void *ctx, feedfunc feed, double d, uint32_t fla
 	}
 
 	/* check sign */
-	if (((num64 >> 63) & 1) != 0) {
+	if (msb != 0) {
 		sign = '-';
 		d = -d;
```
ziemleszcz: Is there possibility that `d` and `msb` sign will disagree after `long double -> double` conversion in `format_parse`?

I think it's safer to clear sign bit explicitly. Remove `d = -d` and replace num64 assignment at L657 with:
```
num64 = format_u64FromDouble(d) & ~((uint64_t)1 << 63);
d = format_doubleFromU64(num64);
```
and add:
```
static inline double format_doubleFromU64(uint64_t num64)
{
	union double_u64 u;

	u.u = num64;
	return u.d;
}
```

rmikielis: ```num64 |= ((uint64_t)msb << 63)``` seems to be sparse, and I think ```d = -d``` is also unnecessary.
It looks like only ```sign``` symbol is used to assess double sign, no its ```msb``` in printing function (```format_printBuffer()``` in L795). It sounds sane since diff. between positive and negative floating point number is only 1 bit, not almost all bits like for integers.
Testing locally with mentioned lines deleted looks good so far.

PR #488, opened 2026-06-25
math: improve performance of sin and cos

jakubsmolaga:
TASK: PP-294

> [!NOTE]
> This change was only tested on `armv8m55-stm32n6` since this is the only platform I have access to. It would be great if people in @phoenix-rtos/phoenix-team could verify that this change doesn't introduce regressions on other relevant platforms

This change achieves a speedup of ~3x on `armv8m55-stm32n6` for `sin` and `cos` without sacrificing precision

## Description
Instead of iteratively calculating the factorials and dividing by them, I just pre-computed the inverses of those factorials, thus replacing divides with multiplies which are much cheaper (>14x on our platform). The methodology for benchmarking and testing precision is described in the next section. When reviewing the code please note that hexadecimal floating point literals were used to maintain maximum precision.

Average time of calls to `cos` before the change:
image
Average time of calls to `cos` after the change:
image
Absolute errors before the change:
image
Absolute errors after the change:
image

## Motivation and Context

On our platform (Arm Cortex-M55), floating point division is very expensive. For double precision it's 28 cycles per divide and only one division can be in the pipeline at a time. The previous implementation was doing 10 divisions, which added up quickly across our entire application. Below is the code that was used for generating constants, benchmarking and verifying the implementation:

<details>
<summary>Code used for benchmarking</summary>

```c
#include <stdio.h>
#include <math.h>
#include <sys/time.h>


#define ITERATIONS 1000000


int main(void)
{
	time_t start, end;

	register double x = 1.0;
	register double y = 0.0;
	volatile double antiOptimize = 0.0;

	gettime(&start, NULL);
	for (int i = 0; i < ITERATIONS; i++) {
		y = cos(x);
		x = y + antiOptimize;
	}
	gettime(&end, NULL);
	__asm__ volatile("" : : "r"(x));

	time_t total = end - start;

	x = 1.0;
	gettime(&start, NULL);
	for (int i = 0; i < ITERATIONS; i++) {
		y = x;
		x = y + antiOptimize;
	}
	gettime(&end, NULL);
	__asm__ volatile("" : : "r"(x));

	time_t overhead = end - start;

	time_t real = total - overhead;
	double avg = (double)real / ITERATIONS;

	printf("avg: %fus/iter\n", avg);
}
```
</details>

<details>
<summary>Code used to test precision</summary>

```c
#include <stdio.h>
#include <math.h>


typedef struct {
	double x, sin, cos;
} testCase_t;


int main(void)
{
	testCase_t tests[] = {
		{ 0x0.0p+0, 0x0.0p+0, 0x1.0000000000000p+0 },
		{ 0x1.0c152382d7366p-1, 0x1.0000000000000p-1, 0x1.bb67ae8584caap-1 },
		{ 0x1.921fb54442d18p-1, 0x1.6a09e667f3bcdp-1, 0x1.6a09e667f3bcdp-1 },
		{ 0x1.0c152382d7366p+0, 0x1.bb67ae8584caap-1, 0x1.0000000000000p-1 },
		{ 0x1.921fb54442d18p+0, 0x1.0000000000000p+0, 0x1.9b3cd3a431b30p-339 },
		{ 0x1.921fb54442d18p+1, 0x1.9b3cd3a431b30p-338, -0x1.0000000000000p+0 },
		{ 0x1.921fb54442d18p+2, -0x1.9b3cd3a431b30p-337, 0x1.0000000000000p+0 },
		{ -0x1.3c083126e978dp+0, -0x1.e351c8409f41dp-1, 0x1.51e9b9f0886adp-2 },
		{ 0x1.4f8b588e368f1p-17, 0x1.4f8b588e1e8a2p-17, 0x1.ffffffff920c8p-1 },
		{ 0x1.34a3d70a3d70ap+10, 0x1.5f77622e0912cp-4, -0x1.fe1c933027b42p-1 },
	};

	int n = sizeof(tests) / sizeof(tests[0]);

	double sinMin = INFINITY;
	double sinMax = 0.0;
	double sinSum = 0.0;
	double cosMin = INFINITY;
	double cosMax = 0.0;
	double cosSum = 0.0;

	for (int i = 0; i < n; i++) {
		double actual_sin = sin(tests[i].x);
		double actual_cos = cos(tests[i].x);
		double err_sin = fabs(actual_sin - tests[i].sin);
		double err_cos = fabs(actual_cos - tests[i].cos);
		if (err_sin < sinMin)
			sinMin = err_sin;
		if (err_sin > sinMax)
			sinMax = err_sin;
		sinSum += err_sin;
		if (err_cos < cosMin)
			cosMin = err_cos;
		if (err_cos > cosMax)
			cosMax = err_cos;
		cosSum += err_cos;
	}

	double sinAvg = sinSum / n;
	double cosAvg = cosSum / n;

	printf("sin: min=%e, max=%e, avg=%e, total=%e\n", sinMin, sinMax, sinAvg, sinSum);
	printf("cos: min=%e, max=%e, avg=%e, total=%e\n", cosMin, cosMax, cosAvg, cosSum);

	return 0;
}
```
</details>

<details>
<summary>Code used to generate test data</summary>

```python
import mpmath

mpmath.mp.dps = 100 

angles = [
    mpmath.mpf('0.0'),
    mpmath.pi / 6,
    mpmath.pi / 4,
    mpmath.pi / 3,
    mpmath.pi / 2,
    mpmath.pi,
    2 * mpmath.pi,
    mpmath.mpf('-1.2345'),
    mpmath.mpf('1e-5'),
    mpmath.mpf('1234.56')
]

for a in angles:
    print(f"{float.hex(float(a))} {float.hex(float(mpmath.sin(a)))} {float.hex(float(mpmath.cos(a)))}")
```
</details>

<details>
<summary>Code used to generate constants (manually edited for the two cases)</summary>

```c
#include <stdio.h>

int main()
{
	long double factorial = 1.0L;
	long double sign = -1.0L;
	for (int i = 3; i <= 22; i += 2) {
		factorial *= i * 1.0L;
		factorial *= (i - 1) * 1.0L;
		long double inv = 1.0L / factorial;
		long double val = sign * inv;
		printf("%a\n", (double)val);
		sign = -sign;
	}
	return 0;
}
```
</details>

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv8m55-stm32n6.

## Checklist:

## Special treatment

### DIFF
--- a/math/trig.c
+++ b/math/trig.c
@@ -5,8 +5,8 @@
  *
  * cos, sin, tan, acos, asin, atan
  *
- * Copyright 2017, 2018 Phoenix Systems
- * Author: Aleksander Kaminski
+ * Copyright 2017, 2018, 2026 Phoenix Systems
+ * Author: Aleksander Kaminski, Jakub Smolaga
  *
  * This file is part of Phoenix-RTOS.
  *
@@ -17,11 +17,10 @@
 #include <errno.h>
 
 
-/* Calculates value of cosine using Maclaurin series. */
+/* Calculates value of cosine using Maclaurin series via Horner's method. */
 double cos(double x)
 {
-	int i;
-	double res, xn, xpow, factorial;
+	double res, xpow;
 
 	if (isnan(x) != 0) {
 		return NAN;
@@ -50,32 +49,28 @@ double cos(double x)
 		return sin(x + M_PI_2);
 	}
 
-	res = 1.0;
 	xpow = x * x;
-	xn = xpow;
-	factorial = 2.0;
 
-	for (i = 0; i < 10; ++i) {
-		if (i & 1) {
-			res += xn / factorial;
-		}
-		else {
-			res -= xn / factorial;
-		}
-
-		xn *= xpow;
-		factorial = factorial * ((2 * i) + 3) * ((2 * i) + 4);
-	}
+	res = 0x1.e542ba4020225p-62;              /* 1/20! */
+	res = res * xpow - 0x1.6827863b97d97p-53; /* 1/18! */
+	res = res * xpow + 0x1.ae7f3e733b81fp-45; /* 1/16! */
+	res = res * xpow - 0x1.93974a8c07c9dp-37; /* 1/14! */
+	res = res * xpow + 0x1.1eed8eff8d898p-29; /* 1/12! */
+	res = res * xpow - 0x1.27e4fb7789f5cp-22; /* 1/10! */
+	res = res * xpow + 0x1.a01a01a01a01ap-16; /* 1/8!  */
+	res = res * xpow - 0x1.6c16c16c16c17p-10; /* 1/6!  */
+	res = res * xpow + 0x1.5555555555555p-5;  /* 1/4!  */
+	res = res * xpow - 0x1p-1;                /* 1/2!  */
+	res = res * xpow + 0x1p0;                 /* 1/0!  */
 
 	return res;
 }
 
 
-/* Calculates value of sine using Maclaurin series. */
+/* Calculates value of sine using Maclaurin series via Horner's method. */
 double sin(double x)
 {
-	int i;
-	double res, xn, xpow, factorial;
+	double res, xpow;
 
 	if (isnan(x) != 0) {
 		return NAN;
@@ -108,24 +103,21 @@ double sin(double x)
 		return -cos(x + M_PI_2);
 	}
 
-	res = x;
 	xpow = x * x;
-	xn = xpow * x;
-	factorial = 6.0;
 
-	for (i = 0; i < 10; ++i) {
-		if (i & 1) {
-			res += xn / factorial;
-		}
-		else {
-			res -= xn / factorial;
-		}
+	res = 0x1.71b8ef6dcf572p-66;              /* 1/21! */
+	res = res * xpow - 0x1.2f49b46814157p-57; /* 1/19! */
+	res = res * xpow + 0x1.952c77030ad4ap-49; /* 1/17! */
+	res = res * xpow - 0x1.ae7f3e733b81fp-41; /* 1/15! */
+	res = res * xpow + 0x1.6124613a86d09p-33; /* 1/13! */
+	res = res * xpow - 0x1.ae64567f544e4p-26; /* 1/11! */
+	res = res * xpow + 0x1.71de3a556c734p-19; /* 1/9!  */
+	res = res * xpow - 0x1.a01a01a01a01ap-13; /* 1/7!  */
+	res = res * xpow + 0x1.1111111111111p-7;  /* 1/5!  */
+	res = res * xpow - 0x1.5555555555555p-3;  /* 1/3!  */
+	res = res * xpow + 0x1.0p0;               /* 1/1!  */
 
-		xn *= xpow;
-		factorial = factorial * ((2 * i) + 4) * ((2 * i) + 5);
-	}
-
-	return res;
+	return res * x;
 }
 
 


julianuziemblo:
> This change was only tested on armv8m55-stm32n6 since this is the only platform I have access to. It would be great if people in @phoenix-rtos/phoenix-team could verify that this change doesn't introduce regressions on other relevant platforms

Couldn't the provided tests just be added to https://github.com/phoenix-rtos/phoenix-rtos-tests if the goal is to verify regressions? We don't have all the board at hand either, so I suppose this is the strategy we'd use also. This way, the PR adding the tests (with the same branch name so that it checks out this PR) will run the test for all the targets in CI.

If you're talking about comparing the 2 solutions (with and without this PR) on other targets, this could also be done with the help of the added tests, albeit there would be some manual work involved.

jakubsmolaga:
@julianuziemblo 
> > This change was only tested on armv8m55-stm32n6 since this is the only platform I have access to. It would be great if people in @phoenix-rtos/phoenix-team could verify that this change doesn't introduce regressions on other relevant platforms
> 
> Couldn't the provided tests just be added to https://github.com/phoenix-rtos/phoenix-rtos-tests if the goal is to verify regressions? We don't have all the board at hand either, so I suppose this is the strategy we'd use also. This way, the PR adding the tests (with the same branch name so that it checks out this PR) will run the test for all the targets in CI.
> 
> If you're talking about comparing the 2 solutions (with and without this PR) on other targets, this could also be done with the help of the added tests, albeit there would be some manual work involved.

If my understanding is correct then the correctness is already tested in `phoenix-rtos-tests/libc/math/trig.c`. What I meant is that I wouldn't want to accidentally introduce a performance regression which could impact other projects. I can't really imagine this implementation being slower on any platform but I just wanted to be safe.

math/trig.c
ziemleszcz: Please add yourself to the list of authors so we know who to bother if something goes wrong. 🙂

math/trig.c
```diff
@@ -20,8 +20,7 @@
 /* Calculates value of cosine using Maclaurin series. */
```
ziemleszcz: Change to `Calculates value of cosine using Maclaurin series via Horner's method.`

math/trig.c
```diff
@@ -74,8 +71,7 @@ double cos(double x)
 /* Calculates value of sine using Maclaurin series. */
```
ziemleszcz: Change to `Calculates value of sine using Maclaurin series via Horner's method.`

math/trig.c
```diff
@@ -50,22 +49,20 @@ double cos(double x)
...
-			res += xn / factorial;
-		}
-		else {
-			res -= xn / factorial;
-		}
-
-		xn *= xpow;
-		factorial = factorial * ((2 * i) + 3) * ((2 * i) + 4);
-	}
+	res = -0x1.0ce396db7f853p-70;             /* 1/22! */
```
ziemleszcz: This adds one extra term compared to the old version. Please  explain the rationale behind the change as the precision gain appears to be negligible.

math/trig.c
ziemleszcz, line 124: Regression on `tan(M_PI_2)` (result of `cos(M_PI_2)` regression):
 - old version tan(M_PI_2) is large value
- new version tan(M_PI_2) is exact zero

math/trig.c
```diff
@@ -20,8 +20,7 @@
 /* Calculates value of cosine using Maclaurin series. */
 double cos(double x)
```
ziemleszcz, line 21: Regression on `cos(M_PI_2)`:
 - old version cos(M_PI_2) is very small value
- new version cos(M_PI_2) is exact zero

math/trig.c
ziemleszcz, line 45: FYI: simple fix created by AI for `cos(M_PI_2)` and `tan(M_PI_2)` regressions. Change L45-50 to:

    static const double M_PI_2_LO = 0x1.1a62633145c07p-54;
    if (x >= M_PI_2) {
        /* cos(x) = -sin(x - π/2); use extended precision to avoid cos(M_PI_2) = 0 */
        return -sin((x - M_PI_2) - M_PI_2_LO);
    }
    else if (x <= -M_PI_2) {
        return sin((x + M_PI_2) + M_PI_2_LO);
    }

ziemleszcz:
Other than the small regression I mentioned, everything else looks good.

math/trig.c
mateusz-bloch: To be more accurate this is not a regression in cos(), but rather a consequence of the numerically inaccurate implementation of tan()

jakubsmolaga: done

math/trig.c
jakubsmolaga: done

math/trig.c
jakubsmolaga: done

math/trig.c
jakubsmolaga: I just miscounted when manually unrolling the loop

jakubsmolaga: (fixed)

math/trig.c
jakubsmolaga: It is possible that this was caused by the extra term in cosine approximation which made it collapse to 0, but I don't know and don't have anything to run the code on at the moment, will check in the morning.

jakubsmolaga: Verified, after removing the extra term, we're back to the old imprecise answer:
```c
printf("cos(M_PI_2)=%la\n", cos(M_PI_2));
printf("cos(M_PI_2) %c= 0.0\n", cos(M_PI_2) == 0.0 ? '=' : '!');
/* 
 * output:
 * cos(M_PI_2)=0x1.1a62633145c07p-54
 * cos(M_PI_2) != 0.0
 */
```

math/trig.c
jakubsmolaga: Yes, at the asymptotes `tan` should return `INFINITY` or `-INFINITY` depending on the sign bit of zero (as suggested by @adamdebek in a private message). Either way, after removing the extra term in `cos` it seems to be back to the old behavior:
```c
printf("tan(cos(M_PI_2))==%g\n", tan(cos(M_PI_2)));
/*
 * output:
 * tan(cos(M_PI_2))==6.12323e-17
 */
```
@adamdebek suggested adding a change to `tan` in this PR, but after some thought I believe it would probably be better to make that another PR, especially since it would also requires changes to tests and is a much bigger, potentially breaking, change in functionality.

agkaminski (approved):
LGTM

ziemleszcz (approved):

PR #489, opened 2026-06-25
sys/threads: assert on uninitialized mutex/cond handles

xvuko:
This is first step in improving detection of uninitialized handles. Adding checks to non DEBUG builds needs to be considered, but requires more care before it can be safely adopted.

## Motivation and Context
`mutexLock` is frequently used without checking return value. If mutex create is not invoked then all `mutexLock` invocations are silently ignored without any sign that locks are never actually held.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

  * https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/496

### DIFF
--- a/sys/threads.c
+++ b/sys/threads.c
@@ -15,6 +15,11 @@
 
 #include <sys/threads.h>
 #include <errno.h>
+#include <assert.h>
+
+
+/* TODO: expose RESOURCE_ID_MIN from kernel headers */
+#define HANDLE_MIN 1
 
 
 int mutexCreate(handle_t *h)
@@ -28,6 +33,9 @@ int mutexCreate(handle_t *h)
 int mutexLock(handle_t m)
 {
 	int err;
+
+	assert(m >= HANDLE_MIN);
+
 	while ((err = phMutexLock(m)) == -EINTR) ;
 	return err;
 }
@@ -44,6 +52,9 @@ int condCreate(handle_t *h)
 int condWait(handle_t h, handle_t m, time_t timeout)
 {
 	int err, mut_err;
+
+	assert(m >= HANDLE_MIN);
+	assert(h >= HANDLE_MIN);
 
 	err = phCondWait(h, m, timeout);
 
@@ -63,6 +74,9 @@ int mutexLock2(handle_t m1, handle_t m2)
 {
 	int err;
 	int tmp;
+
+	assert(m1 >= HANDLE_MIN);
+	assert(m2 >= HANDLE_MIN);
 
 	if ((err = mutexLock(m1)) < 0)
 		return err;


julianuziemblo:
IIRC imxrt uses uninitialized mutex in `phoenix-rtos/phoenix-rtos-devices/multi/imxrt-multi/gpio.c` (which should be fixed)

sys/threads.c
```diff
@@ -15,6 +15,17 @@
 
 #include <sys/threads.h>
 #include <errno.h>
+#include <stdbool.h>
+#include <assert.h>
+
+
+#ifndef NDEBUG
+static inline bool handleIsValid(handle_t h)
```
oI0ck, line 24: Wouldn't a macro be more fitting here? Also, provide an alternative definition in `#elsif`

oI0ck (commented):
Unsure about this one, maybe asserting things like these should be behind a different macro definition in addition to `NDEBUG`. I can imagine that this might break in unexpected places (which is a good thing in the end, since we shouldn't have uninitialized mutexes, and we often don't check the return value).

Making this opt-in for now would be good for preserving code containing improper uses, as long as it is not fixed. We can then roll this out to be enabled by default and opt-out.

sys/threads.c
xvuko: `#ifndef NDEBUG` is needed here only to avoid unused function warnings. Note that at this point only llvm shows warning here as GCC disables unused check for all `static inline` functions regardless and not only those included from headers.

I prefer to avoid macros when possible. Here function provides better type safety without any drawbacks.

xvuko:
> Unsure about this one, maybe asserting things like these should be behind a different macro definition in addition to `NDEBUG`

`NDEBUG` is an existing mechanism of enabling optional runtime checks. I do not see need to create separate macro here. This is part of phoenix specific API, so it is used only in code we can fix. Note that POSIX locks are not affected as those have explicit handling of this case.

There are some further checks that would be nice to have ie. `assert(err >= 0)`. Those could fail on some (in theory) valid use-cases so it would make sense to make them optional.

After this change is used for some time we could consider removing return value from default mutex API calls and raising trap from kernel. For now NDEBUG is a good way to ensure checks can be disabled on production, but are not missed in CI

sys/threads.c
julianuziemblo, line 24: this name could be somewhat misleading as we don't really check if it is valid for sure, only if it _could be_ valid

julianuziemblo: I'd still provide an alternate definition or state here that this has to be used only inside the `assert` macro

ziemleszcz: Maybe reverse the logic (`handleIsInvalid()`) and add comment that currently false negative is possible.

sys/threads.c
```diff
@@ -28,6 +40,8 @@ int mutexCreate(handle_t *h)
 int mutexLock(handle_t m)
 {
 	int err;
+	assert(handleIsValid(m));
```
ziemleszcz, line 43: This looks strange. Please add blank line before assert.

sys/threads.c
xvuko: Changed to:
```
/* TODO: expose RESOURCE_ID_MIN from kernel headers */
#define HANDLE_MIN 1
...
assert(m >= HANDLE_MIN);
```

sys/threads.c
```diff
@@ -28,6 +33,9 @@ int mutexCreate(handle_t *h)
 int mutexLock(handle_t m)
 {
 	int err;
+
+	assert(m >= HANDLE_MIN);
+
 	while ((err = phMutexLock(m)) == -EINTR) ;
 	return err;
```
julianuziemblo, line 40: maybe we could also assert here that the return value is not -EINVAL, as that would indicate we supplied a non-existing mutex and not just happened to use an existing one?

PR #491, opened 2026-06-30
implement POSIX functions in grp.h header

julianuziemblo:
These functions parse the /etc/group file for information about groups. Added implementations for: getgrnam, getgrgid, getgrnam_r, getgrgid_r, getgrent, endgrent, setgrent.

YT: RTOS-1366

## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/497

## Checklist:

## Special treatment


unistd/grp.c
```diff
@@ -0,0 +1,382 @@
...
+		*buf += len;
+
+		grp->gr_passwd = *buf;
+		len = readgrfield(fp, grp->gr_passwd, *bufsize, false);
+		if (len < 0) {
+			ret = -len;
+			break;
+		}
+		*bufsize -= len;
+		*buf += len;
```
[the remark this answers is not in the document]
julianuziemblo: no - `readgrfield` will give us the length INCLUDING the NULL terminator byte

unistd/grp.c
```diff
@@ -0,0 +1,382 @@
...
+		}
+
+		if (sizeof(char *) > bufsize - len) {
+			return ERANGE;
+		}
+
+		bufsize -= sizeof(char *);
+		*(char **)(buf + bufsize) = buf;
+		buf += len;
+		bufsize -= len;
```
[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist how did you write the 1st comment about null-terminating the string, where you have clearly `i++` in the brackets, so you know that we return the length of the string INCLUDING the null-terminating byte, and then you write these comments stating we don't include them?

unistd/Makefile
```diff
@@ -4,5 +4,5 @@
 # Copyright 2020 Phoenix Systems
 #
 
-OBJS += $(addprefix $(PREFIX_O)unistd/, alarm.o conf.o crypt.o dir.o ioctl-helper.o file.o hostname.o getlogin.o getopt.o getpass.o group.o \
+OBJS += $(addprefix $(PREFIX_O)unistd/, alarm.o conf.o crypt.o dir.o ioctl-helper.o file.o hostname.o getlogin.o getopt.o getpass.o grp.o \
 	hostid.o pause.o pwd.o sys.o user.o)
```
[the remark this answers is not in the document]
julianuziemblo: yeah, I added it back in appropriate file (`stubs.c`)

unistd/grp.c
```diff
@@ -0,0 +1,373 @@
...
+	do {
+		if (grenteof(grp_common.fp)) {
+			return NULL;
+		}
+		currentBuf = grp_common.buf;
+		currentSize = sizeof(grp_common.buf);
+	} while (readgrhdr(grp_common.fp, &grp_common.grp, &currentBuf, &currentSize) != 0 || readgrusers(grp_common.fp, &grp_common.grp, currentBuf, currentSize) != 0);
+
+	return &grp_common.grp;
+}
```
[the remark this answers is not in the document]
julianuziemblo: getgrent should not return ERANGE, but yeah, it should be asserted here that the buffer is enough, will change this

unistd/grp.c
```diff
@@ -0,0 +1,453 @@
...
+		len = readgrfield(fp, grp->gr_passwd, *bufsize);
+		if (len < 0) {
+			ret = -len;
+			break;
+		}
+		*bufsize -= len;
+		*buf += len;
+
+		len = readgrfield(fp, intstr, sizeof(intstr));
+		if (len < 0) {
```
ziemleszcz, line 276: Returns `ERANGE` if group size is 16 chars or longer (`intstr`). This will force `getlinematch()` to grow  common buffer and retry forever until `realloc()` finally fails.

unistd/grp.c
```diff
@@ -0,0 +1,453 @@
...
+		free(grp_common.buf);
+	}
+}
+
+
+static int grp_bufferGrow(void)
+{
+	size_t newSize = max(GRP_DEFAULT_BUFSIZE, grp_common.size + grp_common.size / 2);
+
+	char *newBuf = realloc(grp_common.buf, newSize);
```
ziemleszcz, line 61: Security risk: unbounded `realloc()`. Consider adding a reasonable cap (1-2KB).

unistd/grp.c
```diff
@@ -0,0 +1,453 @@
...
+
+	if (grp_common.fp == NULL) {
+		grp_common.fp = fopen(GRP_PATH, "r");
+		if (grp_common.fp == NULL) {
+			return NULL;
+		}
+	}
+
+	if (grenteof(grp_common.fp)) {
+		return NULL;
```
ziemleszcz: `grenteof()` can return true on error. You should check it with `ferror()` and set errno (see `getlinematch()`).

unistd/grp.c
```diff
@@ -0,0 +1,465 @@
...
+
+static int grp_bufferGrow(void)
+{
+	/* cap the realloc */
+	if (grp_common.size == GRP_MAX_BUSFIZE) {
+		return -ENOMEM;
+	}
+
+	size_t newSize = min(GRP_MAX_BUSFIZE, max(GRP_DEFAULT_BUFSIZE, grp_common.size + grp_common.size / 2));
+
```
ziemleszcz, line 60: There is a potential memory corruption issue if one thread is parsing an entry while another reallocates the buffer at the same time (dangling pointer). POSIX allows this behavior and notes that it may invalidate the entry, but that doesn’t mean we’re allowed to corrupt memory. Consider protecting this structure with a mutex.

unistd/grp.c
ziemleszcz: Why `ferror()` check from L427 is not inside this block?

julianuziemblo: sorry, misunderstood - this makes more sense

unistd/grp.c
```diff
@@ -0,0 +1,485 @@
...
+#include <sys/threads.h>
+#include <limits.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+
+#define GRP_PATH "/etc/group"
+
+#define GRP_DEFAULT_BUFSIZE 128
+#define GRP_MAX_BUSFIZE     2048
```
ziemleszcz: GRP_MAX_BUFSIZE

unistd/grp.c
```diff
@@ -0,0 +1,485 @@
...
+
+
+static int getlinematch(FILE *fp, const char *name, gid_t *gid, struct group *grp, char *buffer, size_t bufsize, struct group **result, bool staticBuf)
+{
+	int ret = 0;
+	char *currentBuf = buffer;
+	size_t currentSize = bufsize;
+
+	if (staticBuf) {
+		if (grp_common.lock == -1 && mutexCreate(&grp_common.lock) < 0) {
```
ziemleszcz: Race on `grp_common.lock == -1` is possible. You can use `pthread_once()` to initialize mutex.

unistd/grp.c
```diff
@@ -0,0 +1,485 @@
...
+	return getgrby_r(NULL, &gid, grp, buffer, bufsize, result, false);
+}
+
+
+struct group *getgrent(void)
+{
+	struct group *result;
+	int ret;
+
+	if (grp_common.fp == NULL) {
```
ziemleszcz: `common.fp` is not protected by the mutex. File leak is possible.

julianuziemblo: but isn't this the expected behavior when doing a multi-threaded getgrent?

ziemleszcz: I don't think we should leak the files.

ziemleszcz (approved):

PR #493, opened 2026-07-02
pthread: add new key to thread's key list

### LINKED ISSUE #1642: pthread_key_delete() returns EINVAL for a valid key
## Category
BUG

## Description
pthread_key_delete() returns EINVAL (22) when called with a key previously
created by pthread_key_create(). The key is valid and has not been deleted prior
to the call.

## Expected behavior
pthread_key_delete() shall return 0 on success for a valid key.

## Actual behavior
Returns 22 (EINVAL) for a key that was successfully created by
pthread_key_create().

## POSIX requirement
POSIX.1-2017 (IEEE Std 1003.1-2017), pthread_key_delete(3p), section DESCRIPTION:
  "The pthread_key_delete() function shall delete a thread-specific data key
  previously returned by pthread_key_create()."

POSIX.1-2017 (IEEE Std 1003.1-2017), pthread_key_delete(3p), section RETURN VALUE:
  "If successful, the pthread_key_delete() function shall return zero;
  otherwise, an error number shall be returned to indicate the error."

The standard defines no mandatory error conditions for pthread_key_delete().
Since the key was validly created, the function shall succeed with return
value 0.

## Minimal reproduction code

```c
#include <stdio.h>
#include <pthread.h>

int main(void)
{
	pthread_key_t key;
	int ret;

	ret = pthread_key_create(&key, NULL);
	printf("pthread_key_create: %d (expected 0)\n", ret);

	ret = pthread_key_delete(key);
	printf("pthread_key_delete: %d (expected 0)\n", ret);

	return 0;
}
```

## Host behavior
```
pthread_key_create: 0 (expected 0)
pthread_key_delete: 0 (expected 0)
```

## Revision
Discovered on 00ac1bc3dfbcfa6e2e6ac7e619910efc5d664d1c

## Target
Reproduced on ia32-generic-qemu, not tested on other targets yet.

## Related tests (damianloew/posix_tests)
- `phoenix-rtos-tests/libc/pthread/pthread_key.c`

These tests are currently skipped on the affected target. After applying a fix,
re-enable them and re-run the suite to verify the fix.

rmikielis:
TASK: RTOS-1375

When new key was created, it wasn't added to thread's context, and `pthread_key_delete()` failed.

## Motivation and Context

Fixes phoenix-rtos/phoenix-rtos-project#1642

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/pthread/pthread.c
+++ b/pthread/pthread.c
@@ -1124,15 +1124,24 @@ int pthread_kill(pthread_t thread, int sig)
 
 int pthread_key_create(pthread_key_t *key, void (*destructor)(void *))
 {
-	int err = 0;
-	*key = malloc(sizeof(__pthread_key_t));
-	if (*key == NULL) {
-		err = ENOMEM;
+	int ret;
+	pthread_key_t localkey;
+
+	localkey = (pthread_key_t)malloc(sizeof(__pthread_key_t));
+	if (localkey == NULL) {
+		return ENOMEM;
 	}
-	else {
-		(*key)->destructor = destructor;
+
+	localkey->destructor = destructor;
+
+	ret = pthread_setspecific(localkey, NULL);
+	if (ret != 0) {
+		free(localkey);
+		return ret;
 	}
-	return err;
+
+	*key = localkey;
+	return 0;
 }
 
 


pthread/pthread.c
```diff
@@ -1124,15 +1124,20 @@ int pthread_kill(pthread_t thread, int sig)
 
 int pthread_key_create(pthread_key_t *key, void (*destructor)(void *))
 {
-	int err = 0;
+	int ret;
 	*key = malloc(sizeof(__pthread_key_t));
```
ziemleszcz: Consider storing `malloc()` result in a temporary variable and updating `*key` only after all checks have passed, just before returning. Some applications may initialize `*key` to NULL and treat non-NULL value after  as an indication of success.

rmikielis: I don't think application is to decide if the key value is correct or not. Key is usually integer up to PTHREAD_KEYS_MAX (which we don't implement either), and not a memory address as in our case. If the application decides to initialize *key = 0 (i dont know what for), `pthread_key_create` can return *key=0 and its a valid key.

ziemleszcz: That was just one example. Some applications don't check the return value at all. They call `pthread_key_create(&key)` during initialization and `pthread_key_delete(&key)` during cleanup. If `pthread_setspecific()` fails, `pthread_key_delete()` could end up double-freeing the key.
The proposed change is simple, costs nothing, and prevents this issue. In general, we shouldn't return invalid pointers to user applications.

pthread/pthread.c
```diff
@@ -1124,15 +1124,23 @@ int pthread_kill(pthread_t thread, int sig)
...
+
+	localkey->destructor = destructor;
+
+	ret = pthread_setspecific(localkey, NULL);
+	if (ret != 0) {
+		free(localkey);
 	}
-	return err;
+
+	*key = localkey;
```
ziemleszcz, line 1143: The idea was to NOT set `*key` on `pthread_setspecific` error.

pthread/pthread.c
```diff
@@ -1124,15 +1124,23 @@ int pthread_kill(pthread_t thread, int sig)
...
+		return ENOMEM;
 	}
-	else {
-		(*key)->destructor = destructor;
+
+	localkey->destructor = destructor;
+
+	ret = pthread_setspecific(localkey, NULL);
+	if (ret != 0) {
+		free(localkey);
```
ziemleszcz, line 1139: `return ret` after `free`.

pthread/pthread.c
```diff
@@ -1124,15 +1124,23 @@ int pthread_kill(pthread_t thread, int sig)
...
 	}
-	return err;
+
+	*key = localkey;
+	return ret;
```
ziemleszcz: return 0

pthread/pthread.c
```diff
@@ -1124,15 +1124,23 @@ int pthread_kill(pthread_t thread, int sig)
 
 int pthread_key_create(pthread_key_t *key, void (*destructor)(void *))
 {
-	int err = 0;
-	*key = malloc(sizeof(__pthread_key_t));
-	if (*key == NULL) {
-		err = ENOMEM;
+	int ret;
+	pthread_key_t localkey = NULL;
```
ziemleszcz: Nitpick: `localkey` is set in next line, why set it to NULL?

ziemleszcz (approved):

PR #510, opened 2026-07-29
!include/errno: transfer errno defines to kernel

julianuziemblo:
It seems that we will need all of those definitions in the kernel at some point, so transfer all of them at once now so that they are available for use later and no more breaking commits than necessary need to be made.

YT: RTOS-1408

## Types of changes

- [x] Chore (refactoring, style fixes, git/CI config, submodule management, no code logic changes)

## How Has This Been Tested?

## Checklist:

## Special treatment

   - https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/820

### DIFF
--- a/include/errno.h
+++ b/include/errno.h
@@ -25,36 +25,6 @@ extern "C" {
 #endif
 
 
-/* Linux/x86-compatible numbers for easier debugging */
-
-#define ENOTEMPTY       39
-#define ELOOP           40
-#define ENOMSG          42
-#define EUNATCH         49
-#define ENODATA         61
-#define ENONET          64
-#define EBADFD          66
-#define EPROTO          71
-#define EBADMSG         74
-#define EOVERFLOW       75
-#define EILSEQ          84
-#define EDESTADDRREQ    89
-#define EPFNOSUPPORT    96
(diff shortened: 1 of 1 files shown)


adamgreloch (approved):
Makes sense for me - we won't risk value clashes that way too. @ziemleszcz what's your take?

include/errno.h
```diff
@@ -25,36 +25,6 @@ extern "C" {
 #endif
 
 
-/* Linux/x86-compatible numbers for easier debugging */
```
ziemleszcz, line 28: I think we should keep this comment in the kernel (assuming it still holds for all error codes in the kernel).

adamgreloch: It doesn't, unfortunately: ENOSYS is swapped with EDEADLK, maybe more. Maybe we should fix this as well.

ziemleszcz: @julianuziemblo Could you fix these error codes to be the same as on Linux and keep the comment.

julianuziemblo: yes, I'm on it already

adamgreloch (approved):

PR #515, opened 2026-08-11
!pthread: implement spin locks

adamgreloch:
TASK: RTOS-1426

## Types of changes

## How Has This Been Tested?

- [X] New test added: lukkrusz/rebased_posix

## Checklist:

## Special treatment


pthread/pthread.c
```diff
@@ -1764,6 +1764,71 @@ int pthread_rwlock_init(pthread_rwlock_t *__restrict__ rwlock, const pthread_rwl
...
+int pthread_spin_lock(pthread_spinlock_t *lock)
+{
+	if (lock == NULL) {
+		return EINVAL;
+	}
+
+	while (atomic_exchange_explicit(&lock->locked, 1, memory_order_acquire) != 0) {
+		while (atomic_load_explicit(&lock->locked, memory_order_relaxed) != 0) {
+			/* TODO: detect if we have SMP. If so, actually spin for a while */
+			sched_yield();
```
ziemleszcz, line 1801: usleep(1) or nsleep() for 1 ns?

include/pthread.h
ziemleszcz: Nitpick: reverse order (the same values as in glibc).

pthread/pthread.c
```diff
@@ -1764,6 +1764,71 @@ int pthread_rwlock_init(pthread_rwlock_t *__restrict__ rwlock, const pthread_rwl
...
+}
+
+
+int pthread_spin_trylock(pthread_spinlock_t *lock)
+{
+	if (lock == NULL) {
+		return EINVAL;
+	}
+
+	if (atomic_exchange_explicit(&lock->locked, 1, memory_order_acquire) == 0) {
```
ziemleszcz, line 1819: Add pre-check similar to the one in `pthread_spin_lock()` to avoid store on busy spinlock?
```
if (atomic_load_explicit(&lock->locked, memory_order_relaxed) != 0) {
    return EBUSY;
}
```

pthread/pthread.c
```diff
@@ -1764,6 +1764,71 @@ int pthread_rwlock_init(pthread_rwlock_t *__restrict__ rwlock, const pthread_rwl
 }
 
 
+int pthread_spin_destroy(pthread_spinlock_t *lock)
```
ziemleszcz, line 1770: Add `_POSIX_SPIN_LOCKS` (value 202405L) to `unistd.h`

pthread/pthread.c
jmaksymowicz: This might be a good idea because otherwise if the thread holding spinlock is of lower priority, the thread waiting on the spinlock will starve it out. With any sort of wait at least there's a chance that the other thread will get scheduled in.

Also our kernel doesn't have real nsleep. All of our timers report with 1 us precision (at most).

pthread/pthread.c
jmaksymowicz: I think this is counter-productive. In the optimistic case (spinlock is not locked) there will be 1 extra load. Because userspace spinlocks are meant to be used in cases when there is very little contention, we should assume the optimistic case.

pthread/pthread.c
adamgreloch: Do we really want to deliberately degrade the performance of a primitive whose only purpose is the performance? If one decides to use spinlocks between threads of varying priority in a priority-preemptive system, he's doomed already. This is an incorrect usage, linux manpage warns about it, I'm not convinced we should care about it. Same-priority threads are scheduled in FIFO, in their case yield/reschedule is sufficient.

pthread/pthread.c
ziemleszcz: I'm worried about the non-optimistic case where trylock() call runs in a loop. This store invalidates the cache line which may contain data protected by the spinlock. It could be even worse on SMP, where N threads do this concurrently.

Remember that this is a user application that will be using this call, so don't expect perfect code.

pthread/pthread.c
ziemleszcz: No, we don't. I proposed this only as a temporary workaround for the livelock. I don't see anything in POSIX standard that prohibits this scenario. I'm open to other solutions to this problem.

pthread/pthread.c
jmaksymowicz: OK, if the user runs it on some platform where atomic stores are a lot more expensive then this may be useful. 
I didn't think about it because on SMP platforms I've worked on (ARMv7 and ARMv8) the data actually needs to be in cache for exclusive accesses to work.

adamgreloch: Added a pre-check then

pthread/pthread.c
adamgreloch: If you're referring to the SMP TODO, it's only a potential for optimization - if we knew we are under SMP, we could spin for a while then re-check again prior to reschedule, hoping that the spinlock owner was executing on some other core.

In non-SMP case I'm not sure there's anything better to do (delay-wise) other than yielding to the next thread as fast as possible.

pthread/pthread.c
```diff
@@ -1767,6 +1767,75 @@ int pthread_rwlock_init(pthread_rwlock_t *__restrict__ rwlock, const pthread_rwl
...
+
+
+int pthread_spin_init(pthread_spinlock_t *lock, int pshared)
+{
+	if (lock == NULL || (pshared != PTHREAD_PROCESS_SHARED && pshared != PTHREAD_PROCESS_PRIVATE)) {
+		return EINVAL;
+	}
+
+	if (pshared != PTHREAD_PROCESS_PRIVATE) {
+		return ENOTSUP;
```
etiaro, line 1783: is that really required?
If we manage to share that memory, why wouldn't that implementation work as SHARED?

adamgreloch: This is an extra functionality that we don't really need for now and hence I haven't dug into this at all. This applies to all the pthreads changes in general. Supporting just the PTHREAD_PROCESS_PRIVATE is POSIX-compliant, so I'd exploit that

pthread/pthread.c
ziemleszcz: I was referring to this livelock scenario:

1. Thread A (priority 7) locks the spinlock.
2. Thread B (priority 1) tries to lock the spinlock.
3. Thread A cannot unlock the spinlock because Thread B is busy-looping on `sched_yield()`.
4. Neither thread can make progress.

`sched_yield() -> usleep(0)` can switch only to a thread of the same priority (as specified by POSIX).

For mutexes, RTOS handles this scenario through priority inheritance.

On non-RTOS systems such as Linux or BSD, this scenario doesn't occur with user-space spinlocks because the scheduler uses time sharing.

Probably the simplest approach is to accept this limitation. Spinlocks should be used only between threads of the same priority and document it. In that case, `sched_yield()` is fine.

One possible optimization would be to busy-spin for a number of iterations before calling `sched_yield()`.

adamgreloch: Ok, then I'm after leaving this as-is and warning about this limitation in the docs.

include/pthread.h
```diff
@@ -39,8 +39,8 @@ extern "C" {
 #define PTHREAD_PRIO_INHERIT 1
 #define PTHREAD_PRIO_PROTECT 2
 
-#define PTHREAD_PROCESS_SHARED  0
-#define PTHREAD_PROCESS_PRIVATE 1
+#define PTHREAD_PROCESS_PRIVATE 0
```
ziemleszcz, line 42: Technically, this breaks the API, so this PR should be marked as a breaking change.

adamgreloch: Done

ziemleszcz (approved):

PR #508, opened 2026-07-21
!pthread: implement mutexattrs, mutex_consistent and mutex_{get,set}prioceiling

adamgreloch:
TASK: RTOS-1399

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (refactoring, style fixes, git/CI config, submodule management, no code logic changes)

## How Has This Been Tested?

- [ ] Already covered by automatic testing.
- [ ] New test added: (add PR link here).
- [ ] Tested by hand on: (list targets here).

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing linter checks and tests passed.
- [ ] My changes generate no new compilation warnings for any of the targets.

## Special treatment

- [X] This PR needs additional PRs to work (list the PRs, preferably in merge-order).
https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/815
- [ ] I will merge this PR by myself when appropriate.

PR #518, opened 2026-08-13
!pthread: implement mutex, cond, rwlock clock{lock,wait} functions 

adamgreloch:
TASK: RTOS-1430

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

(rebased at) https://github.com/phoenix-rtos/libphoenix/pull/508
https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/831

### DIFF
--- a/include/sys/threads.h
+++ b/include/sys/threads.h
@@ -92,16 +92,25 @@ static inline int mutexCreateWithAttr(handle_t *h, const struct lockAttr *attr)
 }
 
 
-extern int phMutexLock(handle_t h);
+extern int phMutexLock(handle_t h, time_t timeout, int clock);
 
 
 extern int mutexLock(handle_t h);
+
+
+extern int mutexLockTimeoutable(handle_t h, time_t timeout, int clock);
 
 
 extern int mutexLock2(handle_t h1, handle_t h2);
 
 
 extern int mutexTry(handle_t h);
+
+
+extern int phMutexWait(handle_t h, time_t timeout, int clock);
+
+
+extern int mutexWait(handle_t h, time_t timeout, int clock);
 
 
 extern int mutexUnlock(handle_t h);
@@ -134,7 +143,10 @@ static inline int condCreateWithAttr(handle_t *h, const struct condAttr *attr)
 extern int condWait(handle_t h, handle_t m, time_t timeout);
 
 
-extern int phCondWait(handle_t h, handle_t m, time_t timeout);
+extern int condClockWait(handle_t h, handle_t m, time_t timeout, int clock);
+
+
+extern int phCondWait(handle_t h, handle_t m, time_t timeout, int clock);
 
 
 extern int condSignal(handle_t h);
--- a/pthread/pthread.c
+++ b/pthread/pthread.c
@@ -236,7 +236,7 @@ static int pthread_create_main(void)
 
 
 int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
-	void *(*start_routine)(void *), void *arg)
+		void *(*start_routine)(void *), void *arg)
 {
 	const pthread_attr_t *attrs = &pthread_attr_default;
 
@@ -655,23 +655,23 @@ int pthread_attr_setstacksize(pthread_attr_t *attr, size_t stacksize)
 
 
 int pthread_attr_setstack(pthread_attr_t *attr, void *stackaddr,
-	size_t stacksize)
+		size_t stacksize)
 {
 	return pthread_attr_setstackaddr(attr, stackaddr) |
-		pthread_attr_setstacksize(attr, stacksize);
+			pthread_attr_setstacksize(attr, stacksize);
 }
 
 
 int pthread_attr_getstack(const pthread_attr_t *attr, void **stackaddr,
-	size_t *stacksize)
+		size_t *stacksize)
 {
 	return pthread_attr_getstackaddr(attr, stackaddr) |
-		pthread_attr_getstacksize(attr, stacksize);
+			pthread_attr_getstacksize(attr, stacksize);
 }
 
 
 int pthread_attr_setschedparam(pthread_attr_t *attr,
-	const struct sched_param *param)
+		const struct sched_param *param)
 {
 	if (attr == NULL || param == NULL) {
 		return EINVAL;
@@ -778,11 +778,11 @@ int pthread_setschedprio(pthread_t thread, int prio);
 
 
 int pthread_getschedparam(pthread_t thread, int *policy,
-	struct sched_param *__restrict param);
+		struct sched_param *__restrict param);
 
 
 int pthread_setschedparam(pthread_t thread, int policy,
-	const struct sched_param *param);
+		const struct sched_param *param);
 
 
 static int pthread_initialize_acquire_release(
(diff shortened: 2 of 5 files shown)


sys/threads.c
```diff
@@ -41,24 +51,42 @@ int condCreate(handle_t *h)
...
+	return condClockWait(h, m, timeout, -1);
+}
+
+
+int mutexWait(handle_t m, time_t timeout, int clock)
+{
+	int err;
+
+	do {
+		err = phMutexWait(m, timeout, clock);
```
ziemleszcz: `phMutexWait()` doesn't exist.

pthread/pthread.c
```diff
@@ -1981,17 +2046,33 @@ int pthread_rwlock_timedrdlock(pthread_rwlock_t *restrict rwlock, const struct t
...
-	const time_t abs_timeout_us = __timespecToUs(abs_timeout);
+	time_t abstime_us = __timespecToUs(abs_timeout);
 
-	if (abs_timeout_us == 0) {
+	if (abstime_us <= 0) {
 		return ETIMEDOUT;
 	}
 
-	return pthread_rwlock_rdlock_ex(rwlock, 1, abs_timeout_us);
+	return pthread_rwlock_rdlock_ex(rwlock, 1, abstime_us, clock);
```
ziemleszcz, line 2044: time_t->int truncation (pre-exisiting issue).

pthread/pthread.c
```diff
@@ -2001,13 +2082,29 @@ int pthread_rwlock_timedwrlock(pthread_rwlock_t *restrict rwlock, const struct t
...
-	const time_t abs_timeout_us = __timespecToUs(abs_timeout);
+	time_t abstime_us = __timespecToUs(abs_timeout);
 
-	if (abs_timeout_us == 0) {
+	if (abstime_us <= 0) {
 		return ETIMEDOUT;
 	}
 
-	return pthread_rwlock_wrlock_ex(rwlock, 1, abs_timeout_us);
+	return pthread_rwlock_wrlock_ex(rwlock, 1, abstime_us, clock);
```
ziemleszcz, line 2080: time_t->int truncation (pre-exisiting issue).

pthread/pthread.c
```diff
@@ -1491,6 +1540,22 @@ int pthread_cond_timedwait(pthread_cond_t *__restrict cond,
...
+	if (clock < 0) {
+		return -clock;
+	}
+	return pthread_cond_clockwait_phx(cond, mutex, clock, abstime);
+}
+
+
+int pthread_cond_timedwait(pthread_cond_t *__restrict cond, pthread_mutex_t *__restrict mutex, const struct timespec *__restrict abstime)
+{
+	return pthread_cond_clockwait_phx(cond, mutex, -1, abstime);
```
ziemleszcz, line 1544: From POSIX: 
```
The pthread_cond_timedwait() function shall be equivalent to pthread_cond_clockwait(), except that it lacks 
the clock_id argument. The clock to measure abstime against shall instead come from the condition variable's 
clock attribute which can be set by [pthread_condattr_setclock()] prior to the condition variable's creation. If no 
clock attribute has been set, the default shall be CLOCK_REALTIME.
```
Also note that our default is CLOCK_MONOTONIC.

sys/threads.c
```diff
@@ -28,7 +28,17 @@ int mutexCreate(handle_t *h)
...
+	while ((err = phMutexLock(m, 0, -1)) == -EINTR)
+		;
+	return err;
+}
+
+
+int mutexLockTimeoutable(handle_t m, time_t timeout, int clock)
+{
+	int err;
+	while ((err = phMutexLock(m, timeout, clock)) == -EINTR)
```
ziemleszcz: Restarts timeout on EINTR (not correct for relative clock). Add FIXME - see `interrupt_queueWait()`.

sys/threads.c
adamgreloch: Leftover, apologies

pthread/pthread.c
```diff
@@ -1491,6 +1540,22 @@ int pthread_cond_timedwait(pthread_cond_t *__restrict cond,
...
+	if (clock < 0) {
+		return -clock;
+	}
+	return pthread_cond_clockwait_phx(cond, mutex, clock, abstime);
+}
+
+
+int pthread_cond_timedwait(pthread_cond_t *__restrict cond, pthread_mutex_t *__restrict mutex, const struct timespec *__restrict abstime)
+{
+	return pthread_cond_clockwait_phx(cond, mutex, PH_CLOCK_REALTIME, abstime);
```
ziemleszcz: Why `PH_CLOCK_REALTIME`? The clock should come from `cond` (see `.pthread_condattr_setclock()`). Probably `pthread_cond_t` should have `clock` field set in `pthread_cond_init_cb()`.

pthread/pthread.c
ziemleszcz: CLOCK_REALTIME? According to POSIX: "The default value of the clock attribute shall refer to the system clock."

pthread/pthread.c
adamgreloch: yes, sorry, too many manpages opened at once - thought that pthread_mutex_timedlock and pthread_cond_timedwait source the clock in the same way.

pthread/pthread.c
ziemleszcz: Use CLOCK_REALTIME. All examples of `pthread_cond_timedwait()` including our tests on default settings set time using:
```
clock_gettime(CLOCK_REALTIME, &abstime);
abstime.tv_sec += secs;
```
With default CLOCK_MONOTONIC and NTP changing utcoffs this can wait very long time for a short timeout.
Please test it on QEMU setting wall clock time before calling `pthread_cond_timedwait()`.
Currently CI will not detect it because utcoffs is 0.

adamgreloch: Corrected. Now if I set a 2 second wait for CLOCK_REALTIME and then jump the wall clock 10s forward, the wait takes ~0 seconds

include/sys/threads.h
```diff
@@ -101,12 +101,15 @@ static inline int mutexCreateWithAttr(handle_t *h, const struct lockAttr *attr)
...
 
 
-int phMutexLock(handle_t h);
+int phMutexLock(handle_t h, time_t timeout, int clock);
 
 
 int mutexLock(handle_t h);
 
 
+int mutexLockTimeoutable(handle_t h, time_t timeout, int clock);
```
ziemleszcz: Shouldn't it be somethink like:
```
int mutexLockWait(handle_t h, time_t timeout);
int mutexLockClockWait(handle_t h, time_t timeout, int clock);
```
to match cond functions? `mutexLockWait` would use PH_CLOCK_RELATIVE similar to` condWait()`.

sys/threads.c
```diff
@@ -43,24 +58,30 @@ int condCreate(handle_t *h)
...
+		err = phCondWait(h, m, timeout, clock);
 	}
 
 	return err;
 }
 
 
+int condWait(handle_t h, handle_t m, time_t timeout)
+{
+	return condClockWait(h, m, timeout, -1);
```
ziemleszcz, line 87: Probably we should force `PH_CLOCK_RELATIVE` instead of relying on the kernel to use it.

adamgreloch: Hmm, the intention was to use the clock set via cond attrs

include/sys/threads.h
```diff
@@ -101,12 +101,18 @@ static inline int mutexCreateWithAttr(handle_t *h, const struct lockAttr *attr)
 }
 
 
-int phMutexLock(handle_t h);
+int phMutexLock(handle_t h, time_t timeout, int clock);
```
ziemleszcz, line 104: Please document this API (clock value -1).

Not sure if it would be better to pass `PH_CLOCK_MONOTONIC` explicitly in `mutexLock()` and remove -1 support from the kernel.

include/sys/threads.h
```diff
@@ -150,7 +156,10 @@ static inline int condCreateWithAttr(handle_t *h, const struct condAttr *attr)
 int condWait(handle_t h, handle_t m, time_t timeout);
 
 
-int phCondWait(handle_t h, handle_t m, time_t timeout);
+int condClockWait(handle_t h, handle_t m, time_t timeout, int clock);
+
+
+int phCondWait(handle_t h, handle_t m, time_t timeout, int clock);
```
ziemleszcz, line 164: Please document this API (clock value -1).

ziemleszcz:
Nitpick (pre-exisiting): __timespecToUs() should round `tv_nsec` up (`(time->tv_nsec + 999) / 1000`).

According to POSIX:
`The timeout shall expire when the absolute time specified by abstime passes, as measured by the clock on which timeouts are based (that is, when the value of that clock equals or exceeds abstime)`

include/sys/threads.h
adamgreloch: > Not sure if it would be better [...]

Then the condattrs would do nothing

ziemleszcz: What it has to do with condattrs? -1 is translated directly to PH_CLOCK_MONOTONIC in the kernel. I'm not talking about `phCondWait()`.

adamgreloch: Sorry, you're right. I was almost certain that lockattr stores the clock as well.

adamgreloch: Dropped -1 support.

ziemleszcz (approved):

PR #519, opened 2026-08-24
pthread: remove _ATOMIC and use atomic builtins

adamgreloch:
Including atomics in sys/types.h (introduced by c89c9f7) was a very bad idea as there exists some C++ code that includes unistd.h under extern "C". Using builtins is much more defensive and handles cases like this.

TASK: RTOS-1426

Depends-On: phoenix-rtos-kernel:master

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/include/sys/types.h
+++ b/include/sys/types.h
@@ -23,11 +23,7 @@
 #include <phoenix/types.h>
 
 #ifdef __cplusplus
-#include <atomic>
-#define _ATOMIC(type) std::atomic<type>
-#else
-#include <stdatomic.h>
-#define _ATOMIC(type) _Atomic(type)
+extern "C" {
 #endif
 
 typedef int clock_t;
@@ -58,11 +54,11 @@ typedef uintptr_t pthread_t;
 
 typedef struct {
 	handle_t mutexh;
-	_ATOMIC(int) initialized;
+	int initialized;
 } pthread_mutex_t;
 
 typedef struct {
-	_ATOMIC(int) locked;
+	int locked;
 } pthread_spinlock_t;
 
 typedef struct {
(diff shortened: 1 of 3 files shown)


include/sys/types.h
julianuziemblo, line 1: quick question: why doesn't this header have to be surrounded with `extern "C"` for C++?

pthread/pthread.c
```diff
@@ -786,19 +785,19 @@ int pthread_setschedparam(pthread_t thread, int policy,
...
-		int state = atomic_load_explicit(initialized_flag, memory_order_acquire);
+		int state = __atomic_load_n(initialized_flag, __ATOMIC_ACQUIRE);
 		if (state == RESOURCE_INITIALIZED) {
 			return EOK;
 		}
 
 		if (state == RESOURCE_UNINITIALIZED) {
 			int expected = RESOURCE_UNINITIALIZED;
-			if (atomic_compare_exchange_strong_explicit(initialized_flag, &expected, RESOURCE_INITIALIZING, memory_order_acquire, memory_order_acquire)) {
+			if (__atomic_compare_exchange_n(initialized_flag, &expected, RESOURCE_INITIALIZING, 0, __ATOMIC_ACQUIRE, __ATOMIC_ACQUIRE)) {
```
julianuziemblo: `__atomic_compare_exchange_n` is documented as taking in a boolean as the 4th argument (`weak`), though I don't think it really matters in libphoenix for now

include/sys/types.h
adamgreloch: it doesn't for now, but you're right, it's a good safety measure

julianuziemblo (approved):

ayoopierre (approved):

include/sys/types.h
ziemleszcz: FYI: `extern "C"` is not required in this header becase there are no function declarations here, only typedefs.

PR #521, opened 2026-08-31
!adapt to kernel priority changes

adamgreloch:
Priority should be signed as range got shifted from [0,7] to [-32,31]. Also, PH_GET_PRIO should be used instead of -1 for getting the current priority.

TASK: RTOS-1434

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/833

### DIFF
--- a/include/sys/threads.h
+++ b/include/sys/threads.h
@@ -61,14 +61,14 @@ extern int spawnSyspage(const char *imap, const char *dmap, const char *name, ch
 extern int threadJoin(int tid, time_t timeout);
 
 
-extern int beginthreadex(void (*start)(void *), unsigned int priority, void *stack, unsigned int stacksz, void *arg, handle_t *id);
+extern int beginthreadex(void (*start)(void *), int priority, void *stack, unsigned int stacksz, void *arg, handle_t *id);
 
 
 __attribute__((noreturn))
 extern void endthread(void);
 
 
-static inline int beginthread(void (*start)(void *), unsigned int priority, void *stack, unsigned int stacksz, void *arg)
+static inline int beginthread(void (*start)(void *), int priority, void *stack, unsigned int stacksz, void *arg)
 {
 	return beginthreadex(start, priority, stack, stacksz, arg, NULL);
 }
@@ -77,7 +77,17 @@ static inline int beginthread(void (*start)(void *), unsigned int priority, void
 extern int threadsinfo(int n, threadinfo_t *info);
 
 
-extern int priority(int priority);
+/* Sets the thread priority to val or retrieves the current thread priority when val == PH_GET_PRIO. Returns the current thread priority in *res if res != NULL. */
+int sys_priority(int val, int *res);
+
+
+int setPriority(int val);
+
+
+int getPriority(void);
+
+
+int priority(int val) __attribute__((deprecated("use setPriority() or getPriority() instead")));
 
 
 extern int phMutexCreate(handle_t *h, const struct lockAttr *attr);
@@ -107,7 +117,8 @@ extern int mutexTry(handle_t h);
 extern int mutexConsistent(handle_t h);
 
 
-extern int mutexPrioCeiling(handle_t h, int prioceiling);
+/* Sets the prioceiling to val or retrieves the current prioceiling when val == PH_GET_PRIO. Returns the current prioceiling in *res if res != NULL. */
+extern int mutexPrioCeiling(handle_t h, int prioceiling, int *res);
 
 
 extern int mutexUnlock(handle_t h);
--- a/include/sys/threads.h
+++ b/include/sys/threads.h
@@ -43,29 +43,28 @@ struct __errno_t {
 };
 
 
-extern void _errno_new(struct __errno_t *e);
+void _errno_new(struct __errno_t *e);
 
 
-extern void _errno_remove(struct __errno_t *e);
+void _errno_remove(struct __errno_t *e);
 
 
-extern int gettid(void);
+int gettid(void);
 
 
-extern int exec(const char *path, char *const argv[], char *const env[]);
+int exec(const char *path, char *const argv[], char *const env[]);
 
 
-extern int spawnSyspage(const char *imap, const char *dmap, const char *name, char *const argv[]);
+int spawnSyspage(const char *imap, const char *dmap, const char *name, char *const argv[]);
 
 
-extern int threadJoin(int tid, time_t timeout);
+int threadJoin(int tid, time_t timeout);
 
 
-extern int beginthreadex(void (*start)(void *), int priority, void *stack, unsigned int stacksz, void *arg, handle_t *id);
+int beginthreadex(void (*start)(void *), int priority, void *stack, unsigned int stacksz, void *arg, handle_t *id);
 
 
-__attribute__((noreturn))
-extern void endthread(void);
+__attribute__((noreturn)) void endthread(void);
 
 
 static inline int beginthread(void (*start)(void *), int priority, void *stack, unsigned int stacksz, void *arg)
@@ -74,7 +73,7 @@ static inline int beginthread(void (*start)(void *), int priority, void *stack, 
 }
 
 
-extern int threadsinfo(int n, threadinfo_t *info);
+int threadsinfo(int n, threadinfo_t *info);
 
 
 /* Sets the thread priority to val or retrieves the current thread priority when val == PH_GET_PRIO. Returns the current thread priority in *res if res != NULL. */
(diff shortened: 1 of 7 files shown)


pthread/pthread.c
ziemleszcz: PH_GET_PRIO

ziemleszcz: Rejects negative `prioceiling`.

include/sys/threads.h
```diff
@@ -77,7 +78,25 @@ static inline int beginthread(void (*start)(void *), unsigned int priority, void
...
+static inline int getPriority(void)
+{
+	int ret, prio;
+
+	ret = sys_priority(PH_GET_PRIO, &prio);
+	assert(ret == 0);
+	(void)ret;
+
+	return prio;
+}
```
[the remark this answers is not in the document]
adamgreloch: @gemini-code-assist hmm, the assertion is there to assert that the sys_priority(PH_GET_PRIO, &prio) never fails. It should never fail unless something crazy happens on the kernel side.

sys/resource.c
```diff
@@ -13,24 +13,40 @@
...
+ * that SCHED_RR threads are unaffected by setpriority/getpriority, just do nothing.
+ */
+static int sched_rr_noop(int which, id_t who)
+{
+	if (which != PRIO_PROCESS && which != PRIO_PGRP && which != PRIO_USER) {
+		return SET_ERRNO(-EINVAL);
+	}
+
+	/* Check that pid exists through kill */
+	if (who != 0 && kill(who, 0) < 0) {
```
ziemleszcz: `who` is PID only for `PRIO_PROCESS`. Maybe just return `ESRCH` for PRIO_USER and PRIO_PGRP?
Add explicit cast `kill((pid_t)who, 0)`.

include/sys/threads.h
```diff
@@ -77,7 +80,21 @@ static inline int beginthread(void (*start)(void *), unsigned int priority, void
 extern int threadsinfo(int n, threadinfo_t *info);
 
 
-extern int priority(int priority);
+extern int sys_priority(int priority, int *prev);
+
+
+static inline int priority(int priority)
```
ziemleszcz: Do not use `static inline` for syscalls (see https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1722).
We might want to add new name for it `setPriority` for symmetry with `getPriority` and mark `priority` as obsolete.

include/sys/threads.h
```diff
@@ -43,6 +43,9 @@ struct __errno_t {
 };
 
 
+#define PH_PRIO_DEFAULT 4
```
ziemleszcz: Move it to `phoenix/threads.h` next to PH_GET_PRIO and use it in kernel instead of hard-coded 4.

sys/resource.c
```diff
@@ -13,24 +13,40 @@
  * %LICENSE%
  */
 
+#include <errno.h>
+#include <sched.h>
+#include <signal.h>
 #include <sys/minmax.h>
```
ziemleszcz: Unused.

include/sys/threads.h
adamgreloch: Added `setPriority()` and marked `priority()` as obsolete through a comment - thought about doing it through `__attribute__((deprecated))`, but this would be unprecedented and tricky, especially since we have -Werror. Let's do it as another PR.

sys/resource.c
ziemleszcz: PID for PRIO_PROCESS.

include/sys/threads.h
```diff
@@ -77,7 +77,17 @@ static inline int beginthread(void (*start)(void *), unsigned int priority, void
 extern int threadsinfo(int n, threadinfo_t *info);
 
 
-extern int priority(int priority);
+extern int sys_priority(int priority, int *prev);
```
ziemleszcz: `proc_threadPriority()` always sets `prev` to the current priority (for SET and GET) not the previous one. I'm not sure if this was intended.

adamgreloch: Yes, it was. `prev` should be `val` instead. The point is not to change the existing semantics of the `sys_priority` in the PR kernel and just move the result value to the argument.

ziemleszcz: Consider adding a comment explaining this function's API, as it is not straightforward. For example:
`Sets the thread priority to val or retrieves the current thread priority when val == PH_GET_PRIORITY. Returns the current thread priority in *res if res != NULL.`

adamgreloch: Done

include/sys/threads.h
```diff
@@ -43,47 +43,57 @@ struct __errno_t {
...
 	return beginthreadex(start, priority, stack, stacksz, arg, NULL);
 }
 
 
-extern int threadsinfo(int n, threadinfo_t *info);
+int threadsinfo(int n, threadinfo_t *info);
 
 
-extern int priority(int priority);
+/* Sets the thread priority to val or retrieves the current thread priority when val == PH_GET_PRIORITY. Returns the current thread priority in *res if res != NULL. */
```
ziemleszcz: PH_GET_PRIORITY -> PH_GET_PRIO

include/sys/threads.h
```diff
@@ -43,47 +43,57 @@ struct __errno_t {
...
+int threadsinfo(int n, threadinfo_t *info);
 
 
-extern int priority(int priority);
+/* Sets the thread priority to val or retrieves the current thread priority when val == PH_GET_PRIO. Returns the current thread priority in *res if res != NULL. */
+int sys_priority(int val, int *res);
 
 
-extern int phMutexCreate(handle_t *h, const struct lockAttr *attr);
+int setPriority(int priority);
```
ziemleszcz: Style: `int priority` -> `int val`

include/sys/threads.h
```diff
@@ -43,47 +43,57 @@ struct __errno_t {
...
 
 
-extern int mutexCreate(handle_t *h);
+int getPriority(void);
+
+
+/* WARN: obsolete - use setPriority() instead */
+int priority(int priority);
```
ziemleszcz: Style: `int priority` -> `int val`

include/sys/threads.h
```diff
@@ -92,43 +102,43 @@ static inline int mutexCreateWithAttr(handle_t *h, const struct lockAttr *attr)
...
-extern int mutexTry(handle_t h);
+int mutexTry(handle_t h);
 
 
-extern int mutexConsistent(handle_t h);
+int mutexConsistent(handle_t h);
 
 
-extern int mutexPrioCeiling(handle_t h, int prioceiling);
+int mutexPrioCeiling(handle_t h, int prioceiling, int *prev);
```
ziemleszcz: `prev` is misleading. Maybe change to `res` and and a comment about PH_GET_PRIO.

pthread/pthread.c
```diff
@@ -953,7 +946,7 @@ static int pthread_mutex_lazy_init(pthread_mutex_t *__restrict mutex, const pthr
 
 int pthread_mutex_setprioceiling(pthread_mutex_t *__restrict mutex, int prioceiling, int *__restrict old_ceiling)
 {
-	if (old_ceiling == NULL || prioceiling < 0) {
+	if (old_ceiling == NULL || prioceiling == PH_GET_PRIO) {
```
ziemleszcz: Why not validate `prioceiling` range like in `pthread_mutexattr_setprioceiling()`?

sys/resource.c
```diff
@@ -13,24 +13,42 @@
...
+
+	return 0;
+}
+
 
 int setpriority(int which, id_t who, int prio)
 {
-	int ret = priority(min(max(prio + 4, 0), 7));
-	return SET_ERRNO(ret < 0 ? ret : 0);
+	return sched_rr_noop(which, who);
```
ziemleszcz, line 48: Add:
```
(void) prio;
```
We treat warnings as errors so adding `-Wunused` would break building.

sys/resource.c
```diff
@@ -13,24 +13,42 @@
...
-#include <sys/threads.h>
 #include <errno.h>
+#include <sched.h>
+#include <signal.h>
+#include <sys/resource.h>
+
+
+/*
+ * Since the kernel doesn't support policies other than SCHED_RR and POSIX says
+ * that SCHED_RR threads are unaffected by setpriority/getpriority, just do nothing.
```
ziemleszcz, line 24: Not sure about `getpriority()`. The POSIX say only:

`
Any processes or threads using SCHED_FIFO or SCHED_RR shall be unaffected by a call to setpriority(). This is not considered an error. A process which subsequently reverts to SCHED_OTHER need not have its priority affected by such a setpriority() call.
`

Nothing about `getpriority()` so I think it should still return thread nice value instead of 0.

unistd/sys.c
```diff
@@ -367,3 +363,13 @@ pid_t fork(void)
 	}
 	return pid;
 }
+
+
+int nice(int incr)
+{
+	/*
+	 * Since the kernel doesn't support policies other than SCHED_RR and POSIX says
```
ziemleszcz, line 371: In my opinion it should either return -1 (failed to increase nice value) or the current nice value (see my comment about getpriority()).

sys/resource.c
adamgreloch: But we don't have a notion of nice value. We don't schedule fairly, so this is useless. We can assume 0 to be the "nice value" and since setpriority is uneffective, then nice is constantly 0, so getpriority can return 0.

ziemleszcz (approved):

PR #527, opened 2026-09-11
(phoenix/fbcon, termios, sys/ioctl): minor fixes

julianuziemblo:
* FBCONGETMODE retrieves the FB mode, so added information about this to the request signature (_IOR instead of _IO)
* termios: in tcsetattr, the ioctl command number was treated as unsigned int, which is incorrect as ioctl cmd is unsigned long and lead to bad conversions on 64-bit platforms.
* sys/ioctl/ioctl_setResponse: get data size from msg
  Fix a minor bug where if a command marked as IOC_NESTED (whose ioctl->size might be >= IOCPARM_LEN(request)) is used to set the response buffer here, only the base structure would be copied instead of the whole, serialized structure (with sub-structures pointed-to by pointers in the main structure packed after it in the same buffer).

YT: RTOS-1355

## Motivation and Context

due to branch name (and bad branch name in p-r-posixsrv...) we depend on this for CI to pass
Depends-On: phoenix-rtos-posixsrv:julianuziemblo/RTOS-1355

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


adamgreloch (dismissed):

sys/ioctl.c
```diff
@@ -82,7 +82,8 @@ const void *ioctl_unpackEx(msg_t *msg, unsigned long *request, id_t *id, void **
 
 void ioctl_setResponse(msg_t *msg, unsigned long request, int err, const void *data)
 {
-	size_t size = IOCPARM_LEN(request);
+	ioctl_in_t *ioctl = (ioctl_in_t *)msg->i.raw;
+	size_t size = max(ioctl->size, IOCPARM_LEN(request));
```
ziemleszcz: Why use `ioctl->size`? For `posix_ioctl()` it's equal to `IOCPARM_LEN(request)`. For custom ioctl the sender sets this field so you don't if it safe to copy this amount of bytes from `data`.

ziemleszcz: I still see no reason to change `IOCPARM_LEN(request)` to `max(ioctl->size, IOCPARM_LEN(request))`. ioctl->size is set to `IOCPARM_LEN(request)` anyway in `ioctl()`.

ziemleszcz (approved):

